PR #262, opened 2025-12-30
Flesh out cargo metadata for publishing

tjpalmer:
- Add temper-core version number, required for publishing
- Add other fields we already have handy, which reduce/remove warnings
- Locally tested `cargo publish --dry-run` on both core and std after these changes, without errors
- I'd added some of these manually before publishing 0.6.0 crates earlier today
- More could be added in the future, especially readme, which python also wants

### DIFF
--- a/be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
+++ b/be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
@@ -5,6 +5,7 @@ package lang.temper.be.js
 import lang.temper.be.inputFileMapFromJson
 import lang.temper.common.stripDoubleHashCommentLinesToPutCommentsInlineBelow
 import lang.temper.log.filePath
+import lang.temper.name.DashedIdentifier
 import kotlin.test.Test
 
 class JsBackendTest {
@@ -105,7 +106,7 @@ class JsBackendTest {
             |            ".": "./index.js"
             |          },
             |          "dependencies": {
-            |              "@temperlang/core": "0.6.0"
+            |              "@temperlang/core": "${DashedIdentifier.temperCoreLibraryVersion}"
             |          }
             |        }
             |      },
--- a/cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
+++ b/cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
@@ -28,7 +28,6 @@ import lang.temper.log.FilePath
 import lang.temper.log.FilePathSegment
 import lang.temper.log.MessageTemplate
 import lang.temper.name.BackendId
-import lang.temper.name.DashedIdentifier
 import lang.temper.name.QName
 import lang.temper.supportedBackends.defaultSupportedBackendList
 import lang.temper.tooling.buildrun.BuildHarness
@@ -396,8 +395,6 @@ class BuildTest {
             assertContains(text, "homepage = \"https://temperlang.dev/\"")
             assertContains(text, "repository = \"https://github.com/temperlang/temper\"")
             assertContains(text, "authors = [\"Temper Contributors\"]")
-            val temperCoreLibraryVersion = DashedIdentifier.temperCoreLibraryVersion
-            assertContains(text, "path = \"../temper-core\", version = \"=$temperCoreLibraryVersion\"")
         }
     }
 
--- a/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
+++ b/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
@@ -5,6 +5,7 @@ import lang.temper.be.assertGeneratedCode
 import lang.temper.be.inputFileMapFromJson
 import lang.temper.log.FilePath
 import lang.temper.log.filePath
+import lang.temper.name.DashedIdentifier
 import kotlin.test.Test
 
 @SuppressWarnings("MaxLineLength")
@@ -54,7 +55,7 @@ class RustBackendTest {
             |          rust-version = "1.71.1"
             |
             |          [dependencies]
-            |          temper-core = { path = "../temper-core", version = "=0.6.0" }
+            |          temper-core = { path = "../temper-core", version = "=${DashedIdentifier.temperCoreLibraryVersion}" }
             |
             |          ```
             |      },


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
```diff
@@ -159,20 +164,28 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
...
+                libraryConfiguration.description()?.let { this["description"] = it }
+                libraryConfiguration.homepage()?.let { this["homepage"] = it }
+                libraryConfiguration.repository()?.let { this["repository"] = it }
+            }.map { (key, value) ->
+                // Top-level string escape all above values.
+                key to jsonEscaper.escape(value)
+            }.toMap() + buildMap {
+                // *Not* top-level string values here.
+                libraryConfiguration.authors()?.let { this["authors"] = "[${jsonEscaper.escape(it)}]" }
             }
```
tjpalmer, line 176: The above remains rather inefficient, but at least this part isn't typically in a tight loop.

name/src/commonMain/kotlin/lang/temper/name/ModuleLocation.kt
```diff
@@ -321,6 +321,13 @@ data class DashedIdentifier(
          */
         val temperCoreLibraryIdentifier = DashedIdentifier("temper-core")
 
+        /**
+         * The current version number for the core library.
+         * TODO Move this and [temperCoreLibraryIdentifier] elsewhere?
+         * TODO Script to update all temper-core backends and also std with this?
+         */
+        const val temperCoreLibraryVersion = "0.6.0"
```
tjpalmer, line 329: We had no handy access to this, and I didn't feel like parsing project metadata files again for such. Note the TODO comments above. This can definitely be improved in the future.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/Cargo.toml
```diff
@@ -4,6 +4,11 @@ version = "0.6.0"
 license = "Apache-2.0 OR MIT"
 edition = "2021"
 rust-version = "1.71.1"
+readme = false
```
tjpalmer, line 7: [Here `false` means no readme.](https://rustwiki.org/en/cargo/reference/manifest.html#the-readme-field) The [crates.io page](https://crates.io/crates/temper-core/0.6.0) would look much better with a readme, though. We have one for py temper-core, though small. Maybe could centralize this.

And for temper-built libs, maybe can standardize forwarding README.md files or manufacturing small ones from other metadata when no readme is present. But I didn't feel like going that deep here.

And I also don't generate either readme (not even to false in this pr) or documentation for temper-built libs yet at all. They aren't required for publishing.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/Cargo.toml
```diff
@@ -4,6 +4,11 @@ version = "0.6.0"
...
+documentation = "https://temperlang.dev/tutorial/"
```
tjpalmer, line 8: I suspect a lot of crates go to docs.rs pages for this field, but I haven't reviewed that.

cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
```diff
@@ -386,9 +387,18 @@ class BuildTest {
...
+        topDir.withTextOf("temper.out/rust/std/Cargo.toml") { text ->
+            assertContains(text, "version = ")
+            assertContains(text, "license = \"Apache-2.0 OR MIT\"")
+            assertContains(text, "description = ")
+            assertContains(text, "homepage = \"https://temperlang.dev/\"")
+            assertContains(text, "repository = \"https://github.com/temperlang/temper\"")
+            assertContains(text, "authors = [\"Temper Contributors\"]")
+            val temperCoreLibraryVersion = DashedIdentifier.temperCoreLibraryVersion
+            assertContains(text, "path = \"../temper-core\", version = \"=$temperCoreLibraryVersion\"")
+        }
```
tjpalmer, line 398: Test libs in RustBackendTest don't have many of the fields above, so I added some checking here.

be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
```diff
@@ -105,7 +106,7 @@ class JsBackendTest {
             |            ".": "./index.js"
             |          },
             |          "dependencies": {
-            |              "@temperlang/core": "0.6.0"
+            |              "@temperlang/core": "${DashedIdentifier.temperCoreLibraryVersion}"
```
tjpalmer, line 109: At least can avoid updating here already in the future.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
```diff
@@ -159,20 +164,28 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
...
                 libraryConfiguration.license()?.let { this["license"] = it }
-            }.entries.joinToString("\n") { (key, value) ->
-                "$key = ${jsonEscaper.escape(value)}"
+                libraryConfiguration.description()?.let { this["description"] = it }
+                libraryConfiguration.homepage()?.let { this["homepage"] = it }
+                libraryConfiguration.repository()?.let { this["repository"] = it }
+            }.map { (key, value) ->
+                // Top-level string escape all above values.
+                key to jsonEscaper.escape(value)
+            }.toMap() + buildMap {
```
mikesamuel, line 173: So you build a map, then you escape the values and then you maybe add on authors as a list.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
```diff
@@ -159,20 +164,28 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
...
+                // *Not* top-level string values here.
+                libraryConfiguration.authors()?.let { this["authors"] = "[${jsonEscaper.escape(it)}]" }
```
mikesamuel, line 175: This seems to take a list of many and turn it into a list of one, unless I'm missing something.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/Cargo.toml
mikesamuel: There's enough in this PR.

cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
mikesamuel: Nice test.

name/src/commonMain/kotlin/lang/temper/name/ModuleLocation.kt
mikesamuel: Also TODO: Could this be a SemVer?  Versions tend to get compared, so storing values in a form that has semantically useful comparison leads to fewer downstream bugs.

tjpalmer: > Also TODO: Could this be a SemVer? Versions tend to get compared, so storing values in a form that has semantically useful comparison leads to fewer downstream bugs.

Good idea. #266

PR #268, opened 2026-01-01
Support and document new multiline strings

### LINKED ISSUE #263: Fix syntax highlighting for new multiline strings
It doesn't know how to handle it at all today.

Image

tjpalmer:
- Fix #263 (at least enough for now)
- See also temperlang/temper-syntax#7
- Examples from vscode and tld

image

image

### DIFF
--- a/build-user-docs/skeletal-docs/temper-docs/docs/tutorial/05-builtin.md
+++ b/build-user-docs/skeletal-docs/temper-docs/docs/tutorial/05-builtin.md
@@ -235,11 +235,6 @@ string. If line content starts with something other than `"`, the string ends.
 This syntax allows control over indentation both inside and outside the string
 content.
 
-You can't do single-line multi-quoted strings. For example, `"""abc"""` is an
-error. And there are reasons for that. The current and future feature set of
-Temper string templates goes much deeper than this (see [issue#18]), but let's
-move on for now.
-
 Temper has character syntax using a string tagged with `char`.
 Character values are simple integer code-point values.
 
--- a/docs/for-users/temper-docs/docs/tutorial/05-builtin.md
+++ b/docs/for-users/temper-docs/docs/tutorial/05-builtin.md
@@ -235,11 +235,6 @@ string. If line content starts with something other than `"`, the string ends.
 This syntax allows control over indentation both inside and outside the string
 content.
 
-You can't do single-line multi-quoted strings. For example, `"""abc"""` is an
-error. And there are reasons for that. The current and future feature set of
-Temper string templates goes much deeper than this (see [issue#18](https://github.com/temperlang/temper/issues/18)), but let's
-move on for now.
-
 Temper has character syntax using a string tagged with `char`.
 Character values are simple integer code-point values.
 


build-user-docs/skeletal-docs/temper-docs/docs/tutorial/05-builtin.md
```diff
@@ -211,35 +211,29 @@ backend, but the exact integer differs depending on the "native string encoding.
...
-interactive#2: "\"\"\""
+```temper
+$ ("""
+    "- An outline
+    // Ignored comment.
+    "  - With indentation
+    "- Final point
+  )
+interactive#1: "- An outline\n  - With indentation\n- Final point"
 ```
```
tjpalmer, line 230: Parens for some closing token. Trailing `.toString()` also works, by the way:

```ts
$ """
    .toString()
interactive#0: ""
```

build-user-docs/skeletal-docs/temper-docs/docs/tutorial/05-builtin.md
```diff
@@ -211,35 +211,29 @@ backend, but the exact integer differs depending on the "native string encoding.
...
 
-But you can't do single-line multi-quoted strings. For example, `"""abc"""` is
-an error. And there are reasons for that. The current and future feature set of
-Temper string templates goes much deeper than this (see [issue#18]), but let's
-move on for now.
```
tjpalmer, line 242: I took out the above paragraph because it's no longer quite as relevant and would need updated, anyway.

kcodegen/src/commonMain/kotlin/lang/temper/kcodegen/vscode/syntax/TemperLanguage.kt
```diff
@@ -183,6 +183,12 @@ private fun round(nested: Ref) = Nest(
 )
 
 private fun string() = choice(
+    // TODO Multiline rgx string?
+    stringMultiline(),
+    stringMultiline().copy(
+        begin = """\b($MATCH_WORD)(""${'"'})""",
+        beginCaptures = taggedStringBeginCaptures,
+    ),
```
tjpalmer, line 191: I think the tagged string stuff is just for competing with rgx maybe. Been a while, but to avoid trouble with other leading tags, I seemed to need both tagged and untagged variants above.

langserver/src/main/js/temper-language-support/syntaxes/temper.tmLanguage.json
```diff
@@ -333,6 +333,75 @@
         },
         "string": {
             "patterns": [
+                {
```
mikesamuel, line 336: Oy.  Having 3 lexers is a bit of a pita: the Kotlin one we use for the compiler, pygments, and this.

mikesamuel (approved):

langserver/src/main/js/temper-language-support/syntaxes/temper.tmLanguage.json
tjpalmer: > Oy. Having 3 lexers is a bit of a pita: the Kotlin one we use for the compiler, pygments, and this.

I agree. Two of them use regexes, vscode/textmate pygments. I'd still like to group at least those two under temper-syntax. Maybe can't share full definitions, but I bet we could factor at least some parts with careful planning.

Less sure how easily that would transfer to real temper lexing, but maybe doable. It would be sweet to have a single temper library that shares as much as possible for core temper lexing and parsing needs.

tjpalmer: If temper std/regex doesn't help out of the box (maybe could?) for some of these things, I wonder if the data model still is good enough for describing lexing in a flexible enough way for all the needs.

PR #275, opened 2026-01-06
Fix contributing targets

### LINKED ISSUE #273: Contributing doc missing java8 and lua; rust version wrong
We should update [the contributing file](https://github.com/temperlang/temper/blob/ea46cb354fa7527e452a36d0ed5d55bd00e66338/CONTRIBUTING.md) with missing/updated info:

- Both java8 and lua are missing in the list.
- Looks like [we also depend on a toolchains.xml for maven](https://github.com/temperlang/temper/blob/ea46cb354fa7527e452a36d0ed5d55bd00e66338/.github/workflows/build-and-run-tests.yml#L28) to find the different versions. (Would be nice to work around this automatically sometime maybe, but we can just update the doc for now.)
- The rust version is also wrong. It should say [rust 1.71.1+](https://github.com/temperlang/temper/blob/ea46cb354fa7527e452a36d0ed5d55bd00e66338/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSpecifics.kt#L71).

Meanwhile, [our main testing workflow](https://github.com/temperlang/temper/blob/ea46cb354fa7527e452a36d0ed5d55bd00e66338/.github/workflows/build-and-run-tests.yml) gives some info on how installation can look.

tjpalmer:
- Fix #273

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,11 @@ This allows Temper to reach as many use cases as possible.
 - **Python 3.11** — For Python backend tests
 - **Rust 1.71.1** — Including cargo, for Rust backend tests
 
+These dependencies are also provided for our [primary testing workflow][workflow]
+and for our helper [Docker image][docker deps]. For the workflow, GitHub base
+images provide some dependencies, so we rely on those rather than manually
+installing specific versions.
+
 Temper-built code typically should also work on newer versions of these
 languages/runtimes.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -6,21 +6,40 @@ Thank you for your interest in contributing to Temper! This guide will help you
...
+That said, if you do want to build Temper or the docs site or run most core tests:
 
-- **JDK 17** — Required by Gradle, the build system
-- **Gradle** — [Installation guide][gradle installation]
+- **JDK 21** — Required by Gradle, the build system
 - **Git** with [commit signing] set up
+- **Python 3.11** — To install git pre-commit hooks or run helper scripts
+
+We build Temper using Gradle, but that should be installed automatically if
+missing.
```
tjpalmer, line 16: This has worked for me on multiple separate computers recently.

CONTRIBUTING.md
```diff
@@ -6,21 +6,40 @@ Thank you for your interest in contributing to Temper! This guide will help you
...
-- **Node.js v18** — For JavaScript backend tests; `nvm install lts-hydrogen`
-- **Python 3.11** — For Python backend tests
+Temper translates to many languages, and some tests exercise these languages.
+Where possible, we try to support older versions of target languages.
+This allows Temper to reach as many use cases as possible.
+
 - **Maven** (>3.2.5) — For Java tests; run `scripts/generate-maven-toolchains-xml` to configure
+- **Java 8** - For Java 8 backend tests
+- **Java 17** - For (more modern) Java backend tests
+- **toolchains.xml** - To tell Maven where Java 8 and 17 are, [docs here][toolchains]
```
tjpalmer, line 29: If we switch to gradle, maybe we could avoid all the above? Gradle just might automate everything.

tjpalmer: Kicking off gradle daemons might be somewhat annoying, though.

Issue #264, opened 2025-12-31
Closure capture inconsistent


tjpalmer:
For [this closure capture example](https://github.com/contextfreecode/capture/blob/f1e42cad3a4920b6527ae4cda60853b83c8a6007/capture.temper):

```ts
let gatherHandlers(hub: Hub): ListBuilder<Int> {
  let items = new ListBuilder<Int>();
  for (var i = 0; i < 3; i++) {
    let i = i;
    hub.onAction {
      items.add(i);
    }
  }
  items
}
```

Python doesn't work like the others:

```
$ temper test -b csharp -b java -b js -b lua -b py -b rust
Test failed (py): main - expected items.get(i) == (0) not (2), expected items.get(i) == (1) not (2)
Tests passed: 4 of 6 (1 not run)
Test failed
```

And that failed run is for Rust, which might be nice to report the backend for. I'll write a separate issue for that.

Also, it would be nice if this weren't needed at all:

```ts
let i = i;
```

tjpalmer:
Unfortunately, this changes the behavior of most backends:

```diff
capture
$ git diff
diff --git a/capture.temper b/capture.temper
index 1daa35c..3be3079 100644
--- a/capture.temper
+++ b/capture.temper
@@ -27,7 +27,7 @@ let repeat(times: Int, act: RepeatActor): Void {
 let gatherHandlers(hub: Hub): ListBuilder<Int> {
   let items = new ListBuilder<Int>();
   for (var i = 0; i < 3; ++i) {
-    let i = i;
+    // let i = i;
     hub.onAction {
       items.add(i);
       // console.log("items: ${items.join(" ") { i => i.toString() }}");
```

For example:

```ts
$ temper test -b lua
Test failed (lua): main - expected items.get(i) == (0) not (3), expected items.get(i) == (1) not (3), expected items.get(i) == (2) not (3)
Tests passed: 0 of 1
Test failed
```

But the same capture trick in python doesn't change to all 3s. Modifying after the temper change, this:

```py
def gather_handlers_17(hub_32: "Hub_5") -> "MutableSequence6[int8]":
    items_34: "MutableSequence6[int8]" = list_162()
    i_35: "int8" = 0
    while i_35 < 3:

        def fn_148(i_35=i_35) -> "None":
            items_34.append(i_35)
            print(items_34)

        hub_32.on_action(fn_148)
        i_35 = int_add_161(i_35, 1)
    return items_34
```

gives this result:

```
[...]

PR #276, opened 2026-01-08
be-rust: Borrow funs only where expected

### LINKED ISSUE #265: Failed closure function type in rust
For [this closure capture example](https://github.com/contextfreecode/capture/blob/f1e42cad3a4920b6527ae4cda60853b83c8a6007/capture.temper):

```ts
  public onAction(handler: Handler): Void {
    handlers.add(handler);
  }
```

Rust code gen is bad. Here's the rustc error:

```rs
error[E0308]: mismatched types
   --> src/mod.rs:16:50
    |
16  |     temper_core::listed::add( & self.0.handlers, & ( * handler__41.clone()), None);
    |     ------------------------                     ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Arc<dyn Fn() + Send + Sync>`, found `&dyn Fn() + Send + Sync`
    |     |
    |     arguments to this function are incorrect
    |
    = note: expected struct `Arc<(dyn Fn() + Send + Sync + 'static)>`
            found reference `&dyn Fn() + Send + Sync`
note: function defined here
   --> /home/tom/contextfree/capture/temper.out/rust/temper-core/src/listed.rs:381:12
    |
381 |     pub fn add<T>(builder: &ListBuilder<T>, value: T, index: Option<i32>) -> () {
    |            ^^^
```

Without digging deep, I expect the Arc is correct, and we shouldn't be passing just a ref in.

tjpalmer:
- Fix #265
- Prep test for #264

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
@@ -282,7 +282,7 @@ internal open class FunctionCall(
     val functionName: String,
     builtinOperatorId: BuiltinOperatorId? = null,
     cloneEvenIfFirst: Boolean = false,
-    /** Is some param a special-tailored fn borrow type? */
+    /** Non-null means the indicated param has a special-tailored fn borrow type. */
     val fnIndex: Int? = null,
     hasGeneric: Boolean = false,
     wrapClosures: Boolean = false,


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -236,17 +236,6 @@ internal object GetPromiseResultSyncSupportCode : RustInlineSupportCode(GetPromi
...
-internal object GetResult : RustInlineSupportCode(ConvertedCoroutineAwakeUponFn.name) {
-    override fun inlineToTree(
-        pos: Position,
-        arguments: List<TypedArg<Rust.Tree>>,
-        returnType: Type2,
-        translator: RustTranslator,
-    ): Rust.Tree {
-        error("Specially handled in translator")
-    }
-}
```
tjpalmer, line 248: Wasn't used, per the error there.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -352,31 +345,45 @@ internal open class FunctionCall(
                 }
                 add(self)
                 // Other args.
-                for (arg in arguments.subListToEnd(1)) {
-                    addArg(arg, translator)
+                val otherArgs = arguments.subListToEnd(1)
+                // TODO If we could see raw param types, we could infer this.
```
tjpalmer, line 349: This todo is key. I spent time looking through available data, and I couldn't find it in near call stack. I could find type-arg-bound types, but I need to know if the original signature for the connected function has a function type the arg below.

So I'm manually providing this arg index now for functions that are specialized.

Or if we add borrow types to Temper directly, we could annotate these arg types as borrowed instead of it being a Rust-specific thing. Basically, I'm just trying to optimize a bit here for core functions.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -696,30 +703,34 @@ private val leStrStr = Infix("LeStrStr", BuiltinOperatorId.LeStrStr, RustOperato
 
 private val listedTypes = listOf("Listed", "List", "ListBuilder")
 
-private val listForEach = FunctionCall("List::forEach", "temper_core::listed::list_for_each", hasGeneric = true)
+private val listForEach =
+    FunctionCall("List::forEach", "temper_core::listed::list_for_each", hasGeneric = true, fnIndex = -1)
```
tjpalmer, line 707: I used negative indexing in most cases here for convenience because usually the fn param is the last one.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -821,7 +837,7 @@ internal object PureVirtualBuiltin : RustInlineSupportCode(pureVirtualBuiltinNam
 private val regexCompileFormatted = FunctionCall("Regex::compileFormatted", "compile_formatted")
 private val regexCompiledFind = FunctionCall("Regex::compiledFind", "compiled_find")
 private val regexCompiledFound = FunctionCall("Regex::compiledFound", "compiled_found")
-private val regexCompiledReplace = FunctionCall("Regex::compiledReplace", "compiled_replace")
+private val regexCompiledReplace = FunctionCall("Regex::compiledReplace", "compiled_replace", fnIndex = 2)
```
tjpalmer, line 840: But it's not the last one here. This also isn't a user-facing function, or else I'd probably adjust it.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -530,6 +530,47 @@ class RustBackendTest {
...
+            |    }).clone()
+            |}
+            |struct HubStruct {
+            |    handlers: temper_core::ListBuilder<std::sync::Arc<dyn Fn () + std::marker::Send + std::marker::Sync>>
+            |}
+            |#[derive(Clone)]
+            |pub (crate) struct Hub(std::sync::Arc<HubStruct>);
+            |impl Hub {
+            |    pub fn on_action(& self, handler__0: std::sync::Arc<dyn Fn () + std::marker::Send + std::marker::Sync>) {
+            |        temper_core::listed::add( & self.0.handlers, handler__0.clone(), None);
```
tjpalmer, line 558: A borrow of `handler__0` was happening here, but it shouldn't be borrowed. This isn't a specialized fn borrow case. It's just storage of a value that happens to be a function because it's a list builder of functions.

functional-test-suite/src/commonMain/resources/functions/locals/locals.temper.md
```diff
@@ -169,3 +169,57 @@ Mutually recursive local functions need hoisting or forward declaration.
...
 ```
+
+### Capture scope in a loop
+
+Capture scope can also get tricky across languages. Here we gather up some
+closures with capture from a loop var.
+
+    let hub = new Hub();
+    let items = gatherHandlers(hub);
+    hub.actionPerformed();
```
tjpalmer, line 180: This example is copied from [my contextfreecode example](https://github.com/contextfreecode/capture/blob/cc66d331767b1de8d9d0c6f04b9d51c9d2a2ce48/capture.temper).

functional-test-suite/src/commonMain/resources/functions/locals/locals.temper.md
```diff
@@ -169,3 +169,57 @@ Mutually recursive local functions need hoisting or forward declaration.
...
+
+    let gatherHandlers(hub: Hub): ListBuilder<Int> {
+      let items = new ListBuilder<Int>();
+      for (var i = 0; i < 3; i++) {
+
+Ideally, we don't need `let i = i;` but our `for` desugaring fails to capture
+the changing value without this.
+
+        let i = i;
```
tjpalmer, line 189: I'd like to address the above in `for` desugaring.

functional-test-suite/src/commonMain/resources/functions/locals/locals.temper.md
```diff
@@ -169,3 +169,57 @@ Mutually recursive local functions need hoisting or forward declaration.
...
+        hub.onAction {
+          items.add(i);
+          // console.log("items: ${items.join(" ") { i => i.toString() }}");
```
tjpalmer, line 192: Python fails this, so don't actually log yet.

functional-test-suite/src/commonMain/resources/functions/locals/locals.temper.md
```diff
@@ -169,3 +169,57 @@ Mutually recursive local functions need hoisting or forward declaration.
...
+
+This could represent any kind of event system.
+
+    @fun interface Handler(): Void;
+
+    class Hub {
+      private handlers: ListBuilder<Handler> = new ListBuilder();
+
+      public onAction(handler: Handler): Void {
+        handlers.add(handler);
```
tjpalmer, line 206: This line alone failed to compile on rust before this pr.

functional-test-suite/src/commonMain/resources/functions/locals/locals.temper.md
```diff
@@ -169,3 +169,57 @@ Mutually recursive local functions need hoisting or forward declaration.
...
+Python without extra attention has all `2` values below. And without
+`int i = i;`, we get `3` for all values everywhere.
+
+Note that this is ignored until we fix Python.
+
+```ignore log
+items: 0
+items: 0 1
+items: 0 1 2
+```
```
tjpalmer, line 225: Note the ignore. Can be unignored and logging above also enabled when fixing #264.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -293,6 +282,8 @@ internal open class FunctionCall(
     val functionName: String,
     builtinOperatorId: BuiltinOperatorId? = null,
     cloneEvenIfFirst: Boolean = false,
+    /** Is some param a special-tailored fn borrow type? */
```
mikesamuel: This comment would make sense for a boolean, but is confusing me.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
mikesamuel: I wonder if the cast-needed checks during TmpL translation would help here.

PR #284, opened 2026-01-12
be-rust: Track mutable captures in tests

tjpalmer:
- Fix `var` capture in tests
- [Originally found here](https://github.com/contextfreecode/capture/blob/cc66d331767b1de8d9d0c6f04b9d51c9d2a2ce48/capture.temper#L3) and tested working from this pr

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -528,7 +528,7 @@ class RustTranslator(
                 captures = mapOf(),
                 decl = test,
                 methodizeds = setOf(),
-                mutableCaptures = setOf(),
+                mutableCaptures = test.mutableCaptures(),
                 returnType = TypeDescription.make(WellKnownTypes.voidType2),
             ),
         )
--- a/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
+++ b/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
@@ -436,15 +436,8 @@ class RustBackendTest {
     )
 
     @Test
-    fun captureMut() = assertGenerateWanted(
+    fun captureMutInTest() = assertGenerateWanted(
         temper = """
-            |export let addUp(n: Int): Int {
-            |  var sum = 0;
-            |  repeat(n) { i =>
-            |    sum += i;
-            |  }
-            |  sum
-            |}
             |test("main") {
             |  var sum = 0;
             |  repeat(3) { i =>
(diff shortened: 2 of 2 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -528,7 +528,7 @@ class RustTranslator(
                 captures = mapOf(),
                 decl = test,
                 methodizeds = setOf(),
-                mutableCaptures = setOf(),
+                mutableCaptures = test.mutableCaptures(),
```
tjpalmer, line 531: 1-line fix. FunctionContext is only used in one other place in the translator, and `captures` and `methodizeds` come from outside in that case, so I'm unsure they apply here. Might need to craft a more interesting test case in the future if we run across issues.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -435,6 +435,118 @@ class RustBackendTest {
...
+            |    return temper_core::read_locked( & sum__0);
+            |}
+            |#[cfg(test)]
+            |mod tests {
+            |    #[test]
+            |    fn main__0() -> temper_core::Result<()> {
+            |        crate::init(None);
+            |        temper_std::init(None);
+            |        let test___0 = temper_std::testing::Test::new();
+            |        let mut sum__1: i32 = 0;
```
tjpalmer, line 506: This declaration and later uses fail to capture this local `var` properly.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -435,6 +435,90 @@ class RustBackendTest {
...
+            |    }
+            |}
+            |#[cfg(test)]
+            |mod tests {
+            |    #[test]
+            |    fn main__0() -> temper_core::Result<()> {
+            |        crate::init(None);
+            |        temper_std::init(None);
+            |        let test___0 = temper_std::testing::Test::new();
+            |        let mut sum__1: std::sync::Arc<std::sync::RwLock<i32>> = std::sync::Arc::new(std::sync::RwLock::new(0));
```
tjpalmer, line 476: This and later uses are correct after the 1-line fix. I recorded the [wrong code gen in an earlier commit](https://github.com/temperlang/temper/pull/284/changes/3de31dd49e28fcbf24007122bb6d0f369d3e49a8#diff-1d412556c62f2370df35d85d2bce101644aca5a39e1f46343bc076d185e3bdc3R506) on this branch for reference.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -531,7 +615,7 @@ class RustBackendTest {
     )
 
     @Test
-    fun closureWrong() = assertGenerateWanted(
+    fun funArgWrong() = assertGenerateWanted(
```
tjpalmer, line 618: Just a slightly better name for this existing test.

PR #305, opened 2026-01-26
Grammar support for complex string expressions

mikesamuel:
Previously, there was lexer support for statements embedded in strings expressions, and there was parser and compiler support for interpolated *expressions*, but there was no middle-to-backend support for the statements.

    /* a left delimiter */                 """
    /* ignored margin continues string */  "
                                           "Character data
                                           "${ interpolatedExpression }
    /* Embedded loop start */              "{: for (let x of xs\) { :}
                                           "  Hello, ${ x }
    /* Embedded loop end */                "{: } :}

An earlier commit reworked *TokenSourceAdapter* to produce a sensible concrete-syntax-tree from string expressions with statement fragments.

This commit adds grammar support and reworks the early stage macros so that they can make it through early stages of the compiler.

A follow-up PR will finish extensing that support.

Notable changes include:

- Grammar changes for the synthetic `{ """ ... """ }` blocks in the CST.
- Changes to documentation for string expression grammars
- Documentation now refers to *StringExpr* (string expressions) instead of string literals.
- The grammar diagrams now emit some noisy productions that are for internal book-keeping.
- Parsing a simple string literal now produces a value tree instead of a call to `cat`. Lift was adjusted so that the quotes are present in the position metadata which is necessary for checking in TyperTest that empty strings have type String.
- Reworked parsing so that tagged expressions now go through a StringExprMacro which bridges the difference between complex string expressions and simple ones without requiring the worst-case application for the common case.
- Reworked regex literal macro and the importy automagic it uses to work when run at a later stage, which happens now because StringExprMacro waits until it knows the calling convention.
- Moved Exporter to where it can be referenced by ImportMe which does the import auto-magic.
- Removed cat from the builtin environment.  People can use interpolation syntax.

Opportunistically improved help text for some functions. This aided debugging as I could distinguish the cat function (translated) from the cat macro (adds implicit .toString() and converts to cat function).


astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -186,10 +188,11 @@ private object ProductionNames : DomainSpecificLanguage() {
     val Stmt = Ref("Stmt")
     val StmtBlock = Ref("StmtBlock")
     val StringGroup = Ref("StringGroup")
+    val StringGroupSynthetic = Ref("StringGroupSynthetic")
     val StringGroupTagged = Ref("StringGroupTagged")
     val StringHole = Ref("StringHole")
     val StringHoleRaw = Ref("StringHoleRaw")
-    val StringLiteral = Ref("StringLiteral")
+    val StringExpr = Ref("StringExpr")
```
mikesamuel, line 195: Literal is confusing since "literal" sounds like it means syntax that specifies a unique value.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -317,6 +320,7 @@ val grammar = ProductionNames.run {
     Stmt `：＝` LabeledStmt /
         MatchBranch /
         StmtBlock /
+        StringGroupSynthetic /
```
mikesamuel, line 323: Inside an inverted statement, we need to recognize statement-like things that are synthesized by the TokenSourceAdapter.  Expressions like

    +++ "foo";
    \interpolated x;

This branch shouldn't show up in the grammar documentation though.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -553,7 +557,10 @@ val grammar = ProductionNames.run {
      * - a string group as in a tagged string template like `callee"foo ${ bar }"`.
      */
     CallArgs `：＝` (
-        (syntheticToken y StringGroupTagged) /
+        (
+            syntheticToken y impliedValue(vStringExprMacro) y shiftLeft y impliedValue(TBoolean.valueTrue) y
```
mikesamuel, line 561: These implied args are explained in the StringExprMacro comment.
The shiftLeft moves the macro value back over the callee.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -1263,28 +1270,86 @@ val grammar = ProductionNames.run {
...
-                `)` y ")"
+        syntheticToken y (
+            (
+                "(" y Operator.QuotedGroup y `(` y
+                    TokenType.LeftDelimiter y delimiterMatcher y
+                    any(StringPart) y
+                    TokenType.RightDelimiter y delimiterMatcher y
+                    `)` y ")"
+                ) / (
+                "{" y Operator.QuotedGroup y funStringSymbol y funTree(
```
mikesamuel, line 1312: Here's the block for the ones that have statement fragments inverted.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/GrammarDiagrams.kt
```diff
@@ -27,7 +27,9 @@ object GrammarDiagrams {
                 GrammarDoc.Terminal("true"),
             ),
         ),
-        "StringLiteral" to GrammarDoc.Choice(
+        "StringLiteral" to GrammarDoc.NonTerminal("StringGroup"),
+        "StringGroupSynthetic" to GrammarDoc.Choice.doNotShow,
```
mikesamuel, line 31: A lot of these things we can just splice out of the user docs.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Lift.kt
```diff
@@ -135,10 +136,12 @@ private class Lift(
...
                     i += 1
                     val leaf = astParts[i] as LeafAstPart
-                    make = { t, _ ->
+                    make = { t, right ->
                         val tt = t.treeType
                         if (tt is LeafTreeType) {
-                            makeLeaf(tt, leaf)
+                            // Include the position of quotation marks in string value leaf positions
+                            val leafPos = listOf(part.pos, right).spanningPosition(right)
+                            makeLeaf(tt, leaf, leafPos)
```
mikesamuel, line 144: Make sure that when creating a value for a constant string, that we include the quotes in the position metadata.  Since the langserver and TyperTest depend on position metadata, it's nice to have non-empty metadata for empty string values.

be-test-helpers/src/commonMain/kotlin/lang/temper/be/GeneratedCodeHelper.kt
```diff
@@ -12,6 +12,7 @@ import lang.temper.common.json.JsonObject
 import lang.temper.common.json.JsonString
 import lang.temper.common.json.JsonValue
 import lang.temper.common.orThrow
+import lang.temper.env.Exporter
```
mikesamuel, line 15: Moved Exporter to where it's available to ImportMe which is used by the rgx and `@json` macros to make sure they have std/ pieces they need.

build-user-docs/src/main/kotlin/lang/temper/docbuild/AnnotationExtractor.kt
mikesamuel, line 1: Figure out exactly which snippet texts we need to bundle up for the REPL help function.

build-user-docs/src/main/kotlin/lang/temper/docbuild/GrammarProductionExtractor.kt
```diff
@@ -338,7 +343,9 @@ private class GrammarDiagramConverter(
     @Suppress("SpreadOperator") // Not performance critical
     fun convert(c: GrammarDoc.Component?): Js.Expression {
         return when (c) {
-            null -> undefinedValue()
+            null,
+            GrammarDoc.Choice.doNotShow,
```
mikesamuel, line 347: Do not show things we should not show.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -628,54 +627,6 @@ private object CharTagFn : NamedBuiltinFun, PureCallableValue {
     )
 }
 
-private object InterpolateMacro : BuiltinMacro(interpolateSymbol.text, null) {
```
mikesamuel, line 631: Not needed anymore

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -1053,7 +1008,9 @@ object BuiltinFuns {
                 a * b
             },
         ),
-    )
+    ).also {
+        helpSnippet(it, "Numeric multiplication", "builtin/%2A")
```
mikesamuel, line 1012: Need to add these to a bunch of other builtins.

builtin/src/commonMain/kotlin/lang/temper/builtin/ImportMe.kt
```diff
@@ -92,7 +97,7 @@ fun Planting.makeImportMeCall(
 }
 
 data object ImportMePostPass : PostPass {
-    override fun rewrite(root: BlockTree) {
+    override fun rewrite(root: BlockTree, immediateCallHelper: ImmediateCallHelper) {
```
mikesamuel, line 100: We immediately apply generated imports so that at the end of the stage, the import records are available.  This means tag macros that run during define stage (when we're confident tag names are resolvable), have all their imports lined up for the type stage to type things.

builtin/src/commonMain/kotlin/lang/temper/builtin/StringExprMacro.kt
```diff
@@ -0,0 +1,489 @@
...
+        val tagExprTree = if (isTagged) { args.valueTree(0) } else { null }
+        val isFunString = args.valueTree(2).symbolContained == funStringSymbol
+
+        val argRange = 2..args.lastIndex
+
+        // If we don't have a tag, and it's not a function string, then we can just concatenate
+        // the parts.
+        if (tagExprTree == null) {
+            if (isFunString) {
+                TODO("${macroEnv.stage} ${macroEnv.pos}: untagged isFunString")
```
mikesamuel, line 147: Getting the new tag application convention working in this PR.  The TODOs will be done in a follow-on PR but since there's no compatibility, it should be a much easier PR.

builtin/src/commonMain/kotlin/lang/temper/builtin/StringExprMacro.kt
```diff
@@ -0,0 +1,489 @@
...
+        }
+
+        when (tagCategory) {
+            TagCategory.FnCallStyle -> if (isFunString) {
+                TODO("${macroEnv.stage} ${macroEnv.pos}: $tagExprTree $tagCategory isTagged, isFunString")
+            } else {
+                macroEnv.replaceMacroCallWith {
+                    val call = macroEnv.call!!
+                    val literalParts = mutableListOf<Tree>()
+                    val interpParts = mutableListOf<Tree>()
```
mikesamuel, line 184: This is what the interpolate macro used to do.

builtin/src/commonMain/kotlin/lang/temper/builtin/StringExprMacro.kt
```diff
@@ -0,0 +1,489 @@
...
+
+private fun Planting.buildToStringCall(subject: Tree) =
+    Call {
+        Call {
+            V(Value(DotHelper(memberAccessor = ExternalBind, symbol = toStringSymbol)))
+            Replant(subject)
+        }
+    }
+
+private fun collapseToString(
```
mikesamuel, line 438: For some reason this doesn't show up as a diff to StrCatMacro but these bits were separated out of there unchanged.

cli/src/main/kotlin/lang/temper/cli/repl/Repl.kt
```diff
@@ -124,6 +126,17 @@ internal class Repl(
                         FilePositions.fromSource(loc, sourceFromLoc(loc) ?: return@positions null)
                     }
                 } as T?
+                LibraryNameLocationKey -> if (ReplChunkIndex.from(loc) != null) {
```
mikesamuel, line 129: ImportMe needs metadata like this to avoid extraneous imports, but this is just generally an improvement.

cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
```diff
@@ -876,7 +876,7 @@ class ReplTest {
             """
                 |Describe interactive#2 @ frontend.syntaxMacroStage.before
                 |  do {
-                |    @implicit let x = 1;${
+                |    @implicit @imported(\(`-repl//i0000/`.x)) let x = 1;${
```
mikesamuel, line 879: doImplicitImports now produces metadata that is consistent with explicit imports.

docs/for-users/temper-docs/docs/snippet/syntax/Expr/snippet.svg
mikesamuel, line 1: TODO: This shouldn't have the Quasi branches.

docs/for-users/temper-docs/docs/snippet/syntax/Stmt/snippet.svg
mikesamuel, line 1: TODO: StringGroupSynthetic shouldn't show up here.

frontend/src/commonMain/kotlin/lang/temper/frontend/importstage/ImportStage.kt
```diff
@@ -252,6 +253,7 @@ private fun maybeImportImplicits(
                 Rn(export.name)
                 V(implicitSymbol)
                 V(void)
+                emplaceImportedMeta(leftPos, export.name)
```
mikesamuel, line 256: As shown in the ReplTest change, we now generate consistent metadata.

frontend/src/commonMain/kotlin/lang/temper/frontend/InterpretiveDanceStage.kt
```diff
@@ -111,8 +114,15 @@ internal fun interpretiveDanceStage(
 
     checkInterpreterReachedAll(root, stage, logSink)
 
+    val callHelper = lazy {
+        object : ImmediateCallHelper() {
+            override val env: Environment = env
+            override fun <T> withBoundMacroEnvironment(call: CallTree, doIt: (MacroEnvironment) -> T): T =
+                interpreter.withMacroEnvironment(call, this.env, InterpMode.Partial, doIt)
```
mikesamuel, line 121: Give ImportMe a way to immediately invoke its ImportMacro calls even though it doesn't have the import macro.

fundamentals/src/commonMain/kotlin/lang/temper/value/MacroValue.kt
```diff
@@ -2,14 +2,15 @@ package lang.temper.value
...
+import kotlin.reflect.full.findAnnotations
 
 /**
  * A value that may be called with a set of AST node arguments.
  *
  * Values that are called with actual values as arguments are a subset of these; functions are
  * a subset of macros that don't need to access trees to do their jobs.
  */
-interface MacroValue : StayReferrer {
+interface MacroValue : StayReferrer, OccasionallyHelpful {
```
mikesamuel, line 13: By default, now the REPL can ask a function to describe itself.

fundamentals/src/commonMain/resources/lang/temper/helpful/helpful-snippets.json
mikesamuel, line 1: A pretty minimal dump of help info for the REPL which is updated by `gradle build-user-docs:updateGeneratedDocs`.

interp/src/commonMain/kotlin/lang/temper/interp/UserFunction.kt
```diff
@@ -52,7 +52,7 @@ sealed class UserFunction(
     val superTypes: SuperTypeTree,
     val formalNamesByIndex: List<TemperName>,
     val isSelfContained: Boolean,
-) : CallableValue, TokenSerializable, OccasionallyHelpful {
+) : CallableValue, TokenSerializable {
```
mikesamuel, line 55: MacroValue now is, but the overrides below customize the user function behavior to return its doc string.

astbuild/src/commonTest/kotlin/lang/temper/astbuild/BuildTreeTest.kt
```diff
@@ -4851,6 +4847,66 @@ class BuildTreeTest {
...
+            |  "{:  }  :}
+            |  "</ul>
+        """.trimMargin(),
+        wantJson = """
+            |[ "Call", [
+            |    [ "Value", "stringExpr: Function" ],
+            |    [ "Value", "null: Null" ],
+            |    [ "Value", "false: Boolean" ],
+            |    [ "Value", "\\funString: Symbol" ],
+            |    [ "Fun", [
```
tjpalmer, line 4866: I think the new tree structures will take me some getting used to, but that's fine.

builtin/src/commonMain/kotlin/lang/temper/builtin/RegexLiteralMacro.kt
```diff
@@ -135,20 +136,20 @@ internal object RegexLiteralMacro : BuiltinMacro(regexLiteralBuiltinName.builtin
             tryString(templateStrings, index + 1) { it }?.let { return@invoke it }
         }
         // Try parsing it.
-        val text = builder.toString()
+        val text = "$builder"
```
tjpalmer, line 139: I typically prefer interpolated form in temper these days also.

docs/for-users/temper-docs/docs/reference/builtins.md
```diff
@@ -700,7 +700,7 @@ As explained above, you cannot mix [*Int32*](types.md#type-Int32) and
 
 <!-- /snippet: temper-code/build-user-docs/build/snippet/builtin/%2B/snippet.md/1 -->
 
-`+` does not work on [*String*](types.md#type-String)s.  Use [`cat`](#builtin-cat) instead.
+`+` does not work on [*String*](types.md#type-String)s.  Use [String interpolation](types.md#syntax-string-interpolation) instead.
```
tjpalmer, line 703: This is better for user facing, I think.

builtin/src/commonMain/kotlin/lang/temper/builtin/RegexLiteralMacro.kt
mikesamuel: An IDE suggestion.  I figure a bit of opportunistic cleanup makes it easier to adopt newer Kotlins

PR #310, opened 2026-01-30
Fix string tag functions

tjpalmer:
- Fix a frontend crash
- There might also be weird backend things that happen on watch on crashes like this, but I didn't track that down carefully


builtin/src/commonMain/kotlin/lang/temper/builtin/StringExprMacro.kt
```diff
@@ -286,7 +286,7 @@ object StringExprMacro : BuiltinStatelessMacroValue, NamedBuiltinFun {
 
                     // Collect string parts and interpolations in separate lists.
                     Call(callPos) {
-                        Replant(tagExprTree)
+                        Replant(freeTree(tagExprTree))
```
mikesamuel, line 289: 🤦

PR #312, opened 2026-02-02
Keep type params off implied apply methods

tjpalmer:
- Fix this:

```ts
    @fun interface ContextPropagator<CONTEXT>(
      before: CONTEXT,
      literalPart: String?,
    ): AdjustedStringAndContextAfter<CONTEXT>;
```

causing this:

```ts
7: e ContextPropagator<CONTEXT>(
                       ⇧
[-work/src/core/context-autoescaping.temper.md:7+37]@A: Illegal type parameter CONTEXT. Overridable methods don't allow generics
```

by keeping type params only on the type, not on the method.

### DIFF
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/disambiguate/TypeDisambiguateMacro.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/disambiguate/TypeDisambiguateMacro.kt
@@ -178,6 +178,7 @@ internal fun typeDisambiguateMacro(
             val argsAfterWord = buildList {
                 for (i in args.indices) {
                     when (args.key(i)) {
+                        typeArgSymbol -> continue // Because this should stay on type only
                         wordSymbol -> continue // So we can use it as a type name
                         outTypeSymbol -> hasOutType = true
                     }
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
@@ -815,6 +815,44 @@ class DefineStageTest {
     )
 
     @Test
+    fun functionalInterfaceGeneric() = assertModuleAtStage(
+        stage = Stage.Define,
+        input = """
+            |export @fun interface MyFunction<T, U>(x: T, y: U): Boolean;
+        """.trimMargin(),
+        want = """
+            |{
+            |  define: {
+            |    body:
+            |      ```
+            |      @typeDecl(MyFunction<T__0, U__0>) @stay @functionalInterface let `test//`.MyFunction;
+            |      `test//`.MyFunction = type (MyFunction);
+            |      do {};
+            |      @typeFormal(\T) @typeDefined(T__0) @fromType(MyFunction<T__0, U__0>) let T__0;
+            |      T__0 = type (T__0);
+            |      @typeFormal(\U) @typeDefined(U__0) @fromType(MyFunction<T__0, U__0>) let U__0;
+            |      U__0 = type (U__0);
+            |      MyFunction<T__0, U__0> extends AnyValue;
+            |      @fn @stay @fromType(MyFunction<T__0, U__0>) let apply__0;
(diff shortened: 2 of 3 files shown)


frontend/src/commonMain/kotlin/lang/temper/frontend/disambiguate/TypeDisambiguateMacro.kt
```diff
@@ -178,6 +178,7 @@ internal fun typeDisambiguateMacro(
             val argsAfterWord = buildList {
                 for (i in args.indices) {
                     when (args.key(i)) {
+                        typeArgSymbol -> continue // Because this should stay on type only
```
tjpalmer, line 181: One-line fix. The rest is testing.

frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
```diff
@@ -814,6 +814,44 @@ class DefineStageTest {
         """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
     )
 
+    @Test
+    fun functionalInterfaceGeneric() = assertModuleAtStage(
+        stage = Stage.Define,
+        input = """
+            |export @fun interface MyFunction<T, U>(x: T, y: U): Boolean;
```
tjpalmer, line 821: Very similar to the case above it, but this one is generic. And I put on more than one to be sure I understood the representation better.

frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
```diff
@@ -814,6 +814,44 @@ class DefineStageTest {
...
+            |      @typeDecl(MyFunction<T__0, U__0>) @stay @functionalInterface let `test//`.MyFunction;
+            |      `test//`.MyFunction = type (MyFunction);
+            |      do {};
+            |      @typeFormal(\T) @typeDefined(T__0) @fromType(MyFunction<T__0, U__0>) let T__0;
+            |      T__0 = type (T__0);
+            |      @typeFormal(\U) @typeDefined(U__0) @fromType(MyFunction<T__0, U__0>) let U__0;
+            |      U__0 = type (U__0);
+            |      MyFunction<T__0, U__0> extends AnyValue;
+            |      @fn @stay @fromType(MyFunction<T__0, U__0>) let apply__0;
+            |      apply__0 = fn apply(x__0 /* aka x */: T__0, y__0 /* aka y */: U__0) /* return__0 */: Boolean {
```
tjpalmer, line 837: This had type params on it before the change.

functional-test-suite/src/commonMain/resources/functions/locals/locals.temper.md
```diff
@@ -197,18 +197,18 @@ the changing value without this.
 
 This could represent any kind of event system.
 
-    @fun interface Handler(): Void;
+    @fun interface Handler<T>(item: T): Void;
```
tjpalmer, line 200: The other changes mostly are to support this change. And maybe I should have put multiple type params here, too, but the stage test earlier suggests this should be ok.

Also, this funtest already has generics elsewhere, so this change shouldn't increase the burden when working through a new backend for this funtest.

functional-test-suite/src/commonMain/resources/functions/locals/locals.temper.md
```diff
@@ -219,7 +219,7 @@ Python without extra attention has all `2` values below. And without
 Note that this is ignored until we fix Python.
 
 ```ignore log
-items: 0
-items: 0 1
-items: 0 1 2
+items: 1
+items: 1 2
+items: 1 2 3
 ```
```
tjpalmer, line 225: Note that we still don't produce or test this output, because we still haven't fixed be-py. See #264 on that. But I manually tested all the backends with this enabled and saw the correct output in all buy be-py, where it said 3 for all of the numbers here.

frontend/src/commonMain/kotlin/lang/temper/frontend/disambiguate/TypeDisambiguateMacro.kt
mikesamuel: Ugh.  Nice that you managed to localize it.

PR #314, opened 2026-02-03
Support decorated overloads

tjpalmer:
- Resolve and replace calls to `@overload("whatever")` methods
- Currently doesn't support top-level functions
- Currently depends on no actual method with the same name as the overload decoration name

### DIFF
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
@@ -56,11 +56,8 @@ internal fun simplifyDotHelper(
     // Give preference to members over extensions
     var lastNonExtensionResolution: VariantResolution? = null
     var lastResolution: VariantResolution? = null
-    var lastWasRefined = false
     for (variant in variants.reversed()) {
-        val unrefinedMatch = variant.first == variantMatch
-        if (unrefinedMatch || variant.first == variantMatchRefined) {
-            lastWasRefined = !unrefinedMatch
+        if (variant.first == variantMatch || variant.first == variantMatchRefined) {
             lastResolution = variant.second
             if (variant.second is Either.Left) {
                 lastNonExtensionResolution = variant.second
@@ -74,9 +71,9 @@ internal fun simplifyDotHelper(
         is Either.Left,
         -> {
             val updatedType = when {
-                lastNonExtensionResolution != null && lastWasRefined -> {
-                    // Presumably an overload now resolved to an individually named method.
-                    variantMatchRefined
+                lastNonExtensionResolution?.let { it.leftOrNull!!.symbol != dotHelper.symbol } == true -> {
+                    // An overload now resolved to an individually named method.
+                    variantMatchRefined ?: variantMatch
                 }
                 else -> when {
                     dotHelper.extensions.isNotEmpty() -> chosenVariantResolution?.let {
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
@@ -2069,12 +2069,15 @@ class TypeStageTest {
             |
             |   @overload("toInt")
             |   public stringToInt(string: String): Int32 throws Bubble { string.toInt32(radix) }
+            |
+            |   @overload("justMe")
+            |   public int32ToInt(int: Int32): Int32 { int }
             |}
             |
             |export let crazySum(intMaker: IntMaker, int: Int64, string: String): Int throws Bubble {
             |   let intInt = intMaker.toInt(int);
             |   let stringInt = intMaker.toInt(string);
-            |   intInt + stringInt
+            |   intMaker.justMe(intInt + stringInt)
             |}
         """.trimMargin(),
         want = """
@@ -2108,21 +2111,27 @@ class TypeStageTest {
             |          };
             |        }
             |      };
+            |      @visibility(\public) @overload("justMe") @fn @stay @fromType(IntMaker) let int32ToInt__0 ⦂(fn (IntMaker, Int32): Int32);
+            |      int32ToInt__0 = (@stay fn int32ToInt(@impliedThis(IntMaker) this__2: IntMaker, int__1 /* aka int */: Int32) /* return__2 */: Int32 {
+            |          fn__2: do {
+            |            return__2 = int__1
+            |          }
+            |      });
             |      @fn @visibility(\public) @stay @fromType(IntMaker) let constructor__0 ⦂(fn (IntMaker, Int32): Void);
-            |      constructor__0 = (@stay fn constructor(@impliedThis(IntMaker) this__2: IntMaker, radix__1 /* aka radix */: Int32) /* return__2 */: Void {
-            |          setp(radix__0, this__2, radix__1);
-            |          return__2 = void
+            |      constructor__0 = (@stay fn constructor(@impliedThis(IntMaker) this__3: IntMaker, radix__1 /* aka radix */: Int32) /* return__3 */: Void {
+            |          setp(radix__0, this__3, radix__1);
+            |          return__3 = void
             |      });
             |      @fn @visibility(\public) @stay @fromType(IntMaker) let getradix__0 ⦂(fn (IntMaker): Int32);
-            |      getradix__0 = (@stay fn (@impliedThis(IntMaker) this__3: IntMaker) /* return__3 */: Int32 {
-            |          return__3 = getp(radix__0, this__3)
+            |      getradix__0 = (@stay fn (@impliedThis(IntMaker) this__4: IntMaker) /* return__4 */: Int32 {
+            |          return__4 = getp(radix__0, this__4)
             |      });
-            |      `test//`.crazySum = fn crazySum(intMaker__0 /* aka intMaker */: IntMaker, int__1 /* aka int */: Int64, string__1 /* aka string */: String) /* return__4 */: (Int32 | Bubble) {
+            |      `test//`.crazySum = fn crazySum(intMaker__0 /* aka intMaker */: IntMaker, int__2 /* aka int */: Int64, string__1 /* aka string */: String) /* return__5 */: (Int32 | Bubble) {
             |        void;
-            |        fn__2: do {
+            |        fn__3: do {
             |          var fail#2 ⦂ Boolean, fail#3 ⦂ Boolean;
             |          let intInt__0 ⦂ Int32;
-            |          intInt__0 = hs ⋖ Int32 ⋗(fail#2, do_bind_int64ToInt(intMaker__0)(int__1));
+            |          intInt__0 = hs ⋖ Int32 ⋗(fail#2, do_bind_int64ToInt(intMaker__0)(int__2));
             |          if (fail#2) {
             |            bubble ⋖ Int32 ⋗()
             |          };
@@ -2131,7 +2140,7 @@ class TypeStageTest {
             |          if (fail#3) {
             |            bubble ⋖ Int32 ⋗()
             |          };
-            |          return__4 = intInt__0 + stringInt__0
+            |          return__5 = do_bind_int32ToInt(intMaker__0)(intInt__0 + stringInt__0);
             |        }
             |      }
             |


be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -40,6 +40,51 @@ class PyBackendTest {
...
+            |   @overload("toInt")
+            |   public int64ToInt(int: Int64): Int32 throws Bubble { int.toInt32() }
+            |
+            |   @overload("toInt")
+            |   public stringToInt(string: String): Int32 throws Bubble { string.toInt32(radix) }
+            |}
+            |
+            |export let crazySum(intMaker: IntMaker, int: Int64, string: String): Int throws Bubble {
+            |   let intInt = intMaker.toInt(int);
+            |   let stringInt = intMaker.toInt(string);
```
tjpalmer, line 56: Looks like overloads.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -40,6 +40,51 @@ class PyBackendTest {
...
+            |    def __init__(this_2, radix_15: 'int3') -> None:
+            |        this_2.radix_7 = radix_15
+            |    @property
+            |    def radix(this_24) -> 'int3':
+            |        return this_24.radix_7
+            |def crazy_sum(int_maker_16: 'IntMaker', int_17: 'int64_21', string_18: 'str4') -> 'int3':
+            |    int_int_20: 'int3'
+            |    int_int_20 = int_maker_16.int64_to_int(int_17)
+            |    string_int_21: 'int3'
+            |    string_int_21 = int_maker_16.string_to_int(string_18)
```
tjpalmer, line 82: But isn't really.

And we'll see the same Temper code in TypeStageTest.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
```diff
@@ -36,13 +37,30 @@ internal fun simplifyDotHelper(
 ) {
     val calleeEdge = call.edge(0)
     val callee = calleeEdge.target
-    val variantMatch = call.typeInferences?.variant ?: return
+    val typeInferences = call.typeInferences
+    val variantMatch = typeInferences?.variant ?: return
+    val variantFunctionType = variantMatch as? FunctionType
+    val variantMatchRefined = (variantFunctionType?.returnType as? AndType)?.let { andType ->
+        when {
+            typeInferences.type in andType.members -> MkType.fnDetails(
```
tjpalmer, line 45: Seems like `OrType` would make more sense than `AndType`, but this is what we have. In any case, we don't have some cover function or anything for overloads. We need to resolve to a specific function with a single type, so prepare that one here for later comparison.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
```diff
@@ -36,13 +37,30 @@ internal fun simplifyDotHelper(
...
 
     // Give preference to members over extensions
     var lastNonExtensionResolution: VariantResolution? = null
     var lastResolution: VariantResolution? = null
+    var lastWasRefined = false
     for (variant in variants.reversed()) {
-        if (variant.first == variantMatch) {
+        val unrefinedMatch = variant.first == variantMatch
+        if (unrefinedMatch || variant.first == variantMatchRefined) {
+            lastWasRefined = !unrefinedMatch
```
tjpalmer: If we had to use the refined type, then we presumably chose a specific individual overload.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
```diff
@@ -54,23 +72,34 @@ internal fun simplifyDotHelper(
...
-            val methodVariants = if (chosenVariantResolution != null) {
-                MkType.or(
-                    variants.mapNotNull {
-                        if (it.second is Either.Left) {
-                            it.first
-                        } else {
-                            null
-                        }
-                    },
-                )
```
tjpalmer, line 67: I'm unsure where the best tests are of this logic, but I tried to retain it below.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
```diff
@@ -54,23 +72,34 @@ internal fun simplifyDotHelper(
...
-            } else {
-                // TODO: is this required?
-                null
```
tjpalmer, line 70: I left out the question branch, though, while leaving it still null in this case.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
```diff
@@ -54,23 +72,34 @@ internal fun simplifyDotHelper(
...
+        -> {
+            val updatedType = when {
+                lastNonExtensionResolution != null && lastWasRefined -> {
+                    // Presumably an overload now resolved to an individually named method.
+                    variantMatchRefined
```
tjpalmer: We at least have a meaningful static type update to make.

But I realized in review that I probably need to deal with the case of a single `@overload` method.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
```diff
@@ -54,23 +72,34 @@ internal fun simplifyDotHelper(
...
+                        )
+                    }
+                    else -> null
+                }
             }
-            calleeEdge.replace {
-                V(callee.pos, Value(DotHelper(dotHelper.memberAccessor, dotHelper.symbol, emptyList())), methodVariants)
+            updatedType?.let {
+                calleeEdge.replace {
+                    val symbol = chosenVariantResolution?.item?.symbol ?: dotHelper.symbol
```
tjpalmer, line 96: Prioritize the name of the match, not the previously used name.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -2732,13 +2735,13 @@ internal class Typer(
...
-            for (thisVariant in allThisVariants) {
-                val typeShape = thisVariant.definition as TypeShape
-                if (DEBUG) {
-                    console.log("$thisVariant has members ${typeShape.members}")
-                }
+            fun processMembers(
+                thisVariant: NominalType,
+                typeShape: TypeShape,
+                symbolMatches: (member: MemberShape) -> Boolean = { it.symbol == memberSymbol },
+            ) {
```
tjpalmer, line 2742: Extracts for checking actual name matches first then overload name matches second.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -2758,6 +2761,20 @@ internal class Typer(
...
+                if (DEBUG) {
+                    console.log("$thisVariant has members ${typeShape.members}")
+                }
+                processMembers(thisVariant, typeShape)
+                if (!anyMatchedSymbol) {
+                    // No actual member had the right symbol, so check for overloads.
+                    // Checking for overloads only on actual member symbol fails should make the common case faster.
+                    // TODO Permit overloads with an actual matching member also?
+                    // TODO If not, validate against that somewhere else.
+                    processMembers(thisVariant, typeShape) { matchesOverload(memberSymbol, it) }
```
tjpalmer, line 2775: I've also currently written docs as `@overload` requires no actual method with the overload name. I think it will do for now, but we might want to loosen that later.

And I haven't actually written error reporting for this with the current pr, either.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -2931,6 +2948,14 @@ internal class Typer(
         }
     }
 
+    private fun matchesOverload(symbol: Symbol, member: MemberShape): Boolean = run {
+        (member.stay?.incoming?.source as? DeclTree)?.parts?.let { parts ->
+            parts.metadataSymbolMap[overloadSymbol]?.target?.valueContained?.let { value ->
+                TString.unpackOrNull(value)
+            }
+        } == symbol.text
+    }
```
tjpalmer, line 2957: This looks awfully expensive, which is part of why I only check it if we don't get a simple match first.

frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2056,6 +2056,174 @@ class TypeStageTest {
...
+            |}
+        """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
+    )
+
+    /**
+     * We currently have some direct overloading support, so this helps fore reviewing that behahior.
+     * TODO Drop support for direct overloading.
+     */
+    @Test
+    fun overloadedMethodsWrong() = assertModuleAtStage(
```
tjpalmer, line 2158: I can drop this out of the current pr if we want. I used it for understanding things, but we expect this to be illegal in the future. Maybe should just keep it, anyway, and see the error reporting on it later.

fundamentals/src/commonMain/kotlin/lang/temper/value/NamesAndSymbols.kt
```diff
@@ -531,6 +531,9 @@ val vInlineUnrealizedGoalSymbol = Value(inlineUnrealizedGoalSymbol)
 val extensionSymbol = Symbol("extension")
 val vExtensionSymbol = Value(extensionSymbol)
 
+/** Provides an overload name to a function or class member. */
+val overloadSymbol = Symbol("overload")
```
tjpalmer, line 535: The `vExtensionSymbol` above isn't ever used, and I'm following its style here, so I didn't bother with it.

interp/src/commonMain/kotlin/lang/temper/interp/BuiltinEnvironment.kt
```diff
@@ -386,8 +387,51 @@ private object Builtins {
...
 
-            // TODO: enable the inert code blocks below when extension functions have
-            // been fully integrated into the interpreter and existing backends.
+            /**
+             * <!-- snippet: builtin/@overload -->
+             * # `@overload` decorator
+             * Gives an overload name to a function or class member. Temper respects
+             * overloads. Some backend languages also might, so backends are permitted
+             * to generate multiple entities with the same overload. See each backend
+             * for details.
```
tjpalmer, line 396: Currently, no backends make overloads from these, but we probably should in C# and Java.

interp/src/commonMain/kotlin/lang/temper/interp/BuiltinEnvironment.kt
```diff
@@ -386,8 +387,51 @@ private object Builtins {
...
+             *
+             * let builder = new ConvertingStringBuilder();
+             * builder.append("Hi ");
+             * builder.append(5);
+             * builder.appendString("!"); // individualized calls also work
+             * builder.content.toString() == "Hi 5!"
+             * ```
+             *
+             * For now, only explicit methods are allowed for overloads.
+             * Constructors, propertiers, and top-level functions are unsupported.
```
tjpalmer, line 421: Here's my documentation of the limitation that it only applies to methods right now. I haven't actually tested anything except normal instance methods yet, either. But I know I haven't implemented anything for top-level functions.

interp/src/commonMain/kotlin/lang/temper/interp/BuiltinEnvironment.kt
```diff
@@ -386,8 +387,51 @@ private object Builtins {
...
+             *
+             * No actual function or method with the overload name itself must be
+             * present.
```
tjpalmer, line 424: Here's my claim that you shouldn't have a method with the same name as a decorated overload name.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
```diff
@@ -54,23 +69,34 @@ internal fun simplifyDotHelper(
...
-                )
-            } else {
-                // TODO: is this required?
-                null
+        -> {
+            val updatedType = when {
+                lastNonExtensionResolution?.let { it.leftOrNull!!.symbol != dotHelper.symbol } == true -> {
+                    // An overload now resolved to an individually named method.
+                    variantMatchRefined ?: variantMatch
+                }
```
tjpalmer, line 77: This logic also handles single-method `@overload` groups.

frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2056,6 +2056,183 @@ class TypeStageTest {
...
+            |   public stringToInt(string: String): Int32 throws Bubble { string.toInt32(radix) }
+            |
+            |   @overload("justMe")
+            |   public int32ToInt(int: Int32): Int32 { int }
+            |}
+            |
+            |export let crazySum(intMaker: IntMaker, int: Int64, string: String): Int throws Bubble {
+            |   let intInt = intMaker.toInt(int);
+            |   let stringInt = intMaker.toInt(string);
+            |   intMaker.justMe(intInt + stringInt)
```
tjpalmer, line 2080: All three of these method calls get replaced later. I don't have the `justMe` case in the python backend test, though.

frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2056,6 +2056,183 @@ class TypeStageTest {
...
+            |          intInt__0 = hs ⋖ Int32 ⋗(fail#2, do_bind_int64ToInt(intMaker__0)(int__2));
+            |          if (fail#2) {
+            |            bubble ⋖ Int32 ⋗()
+            |          };
+            |          let stringInt__0 ⦂ Int32;
+            |          stringInt__0 = hs ⋖ Int32 ⋗(fail#3, do_bind_stringToInt(intMaker__0)(string__1));
+            |          if (fail#3) {
+            |            bubble ⋖ Int32 ⋗()
+            |          };
+            |          return__5 = do_bind_int32ToInt(intMaker__0)(intInt__0 + stringInt__0);
```
tjpalmer, line 2143: No mention of simple `toInt` or `justMe` anymore.

docs/for-users/temper-docs/docs/reference/builtins.md
```diff
@@ -3267,6 +3267,54 @@ console.log(c.y.toString()); //!outputs "2"
 
 <!-- /snippet: builtin/@noProperty -->
 
+<!-- snippet: builtin/@overload -->
+
+<a name="builtin&#45;&#64;overload" class="snippet-anchor-name"></a>
+
+### `@overload` decorator
```
mikesamuel, line 3274: Nice docs

interp/src/commonMain/kotlin/lang/temper/interp/BuiltinEnvironment.kt
mikesamuel: I think that's fair.  If, for example, the JS backend wanted to synthesize a method that uses `typeof` and `instanceof` to try and pick the overload to delegate to, then it'd be awkward to have to convert the simply named method to a private.  And it should be our goal that method calls translated from Temper are unambiguous, and in that case we'd be relying on the backend-synthesized method to always delegate to the sleight-of-handed private method.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
mikesamuel: I think we should always prefer the simple match.  There is a mechanism for tiered overload resolution: *CalleePriority* I think which would preserve that property.

It seems like this could be cached, so if a module uses a method a lot, as is probably the case for tagged string expressions, you only incur the cost once.

mikesamuel (approved):

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
tjpalmer: > I think we should always prefer the simple match. There is a mechanism for tiered overload resolution: _CalleePriority_ I think which would preserve that property.

We do prefer the simple match. I don't even check this if there's a simple match. Some future PR should add validation against an decoration overload name conflicting with a basic name, though.

> It seems like this could be cached, so if a module uses a method a lot, as is probably the case for tagged string expressions, you only incur the cost once.

That's a good idea. Once it's been seen once, it might get seen again. So we could cache something in Typer maybe. I didn't want to add a new property to MethodShape, which was the easiest idea I had. Another idea that I didn't work was to make a new kind of member for overload groups. I guess for plain functions, that would require add a new leaf to the body or something.

But caching in Typer might be simpler than the "invent new entity" and less heavyweight than "brand new property that's rarely used".

PR #319, opened 2026-02-05
std/testing: Properly XML encode when producing JUnit XML

mikesamuel:
Junit XML parsing was failing because HTML metacharacters from test names and error messages were no properly encoded.

```
Caused by: javax.xml.stream.XMLStreamException: ParseError at [row,col]:[3,21]
Message: The value of attribute "name" associated with an element type "testcase" must not contain the '<' character.

...

Caused by: javax.xml.stream.XMLStreamException: ParseError at [row,col]:[4,76]
Message: The entity "Bcy" was referenced, but not declared.
```

This commit adds a minimal XML encoder that works for attribute values and text nodes.

### DIFF
--- a/frontend/src/commonMain/resources/std/testing/testing.temper.md
+++ b/frontend/src/commonMain/resources/std/testing/testing.temper.md
@@ -152,17 +152,16 @@ implementation of testing within the interpreter.
       let totals = "tests='${total}' failures='${fails}'";
       // Just lie about time for now since it's required.
       writeLine("  <testsuite name='suite' ${totals} time='0.0'>");
-      let escape(s: String): String { s.split("'").join("&apos;") { x => x } }
       for (var i = 0; i < testResults.length; i += 1) {
         let testResult = testResults[i];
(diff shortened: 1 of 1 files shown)


frontend/src/commonMain/resources/std/testing/testing.temper.md
```diff
@@ -189,3 +188,33 @@ TODO Is this a better idea than inlining each case? We'd need to generate
...
+    let escapeXml(s: String): String {
+      let sb = new StringBuilder();
+      let end = s.end;
+      var emitted = String.begin;
+      for (var i = String.begin; i < end; i = s.next(i)) {
+        let esc = when (s[i]) {
+          char'&' -> "&amp;";
+          char'<' -> "&lt;";
+          char'>' -> "&gt;";
+          char"'" -> "&#39;";
```
mikesamuel, line 206: `&apos;` is now defined in https://www.w3.org/TR/xml/#sec-predefined-ent but has not always been so older XML parsers may not recognize it.

PR #322, opened 2026-02-06
Lighten up cleaning of temporaries

tjpalmer:
- Prevent out of memory error on degenerate temporary var cleaning
- Speed up a simple case of making a map from a list of many (say 1000) pairs
- Include an unrelated small be-cs cleanup while at it
- Example of the tested case, shrunk to 3 vars instead of 1000. Before inlining:

```ts
var t#12, t#13, t#14, t#15, fail#8, fail#9, fail#10, fail#11;
let `test//`.numbers: Map<String, Int32>;
t#12 = new Pair<String, Int32>("0", 0);
t#13 = new Pair<String, Int32>("1", 1);
t#14 = new Pair<String, Int32>("2", 2);
`test//`.numbers = new Map<String, Int32>(list(t#12, t#13, t#14));
```

- After inlining:

```ts
var t#12, t#13, t#14, t#15, fail#8, fail#9, fail#10, fail#11;
let `test//`.numbers: Map<String, Int32>;
void;
void;
void;
`test//`.numbers = new Map<String, Int32>(list(new Pair<String, Int32>("0", 0), new Pair<String, Int32>("1", 1), new Pair<String, Int32>("2", 2)));
```

- Before these changes, 3 passes were needed for the 3 vars in this case. After these changes, all 3 are inlined on the same pass. In both cases, the last one is inlined first because there are no maybe-effectful statements between the reference and the assignment. And after that, the same is true for the 2nd to last, and so on.
- Note that the fail vars also then are eliminated for being unused in the following pass.
- The cleanup first happens in TypeStage, and the same temp-making (out-lining?) happens again afterward, and then they get inlined again in GenerateCodeStage. I haven't addressed this duplication of effort in this PR.
- I also don't guarantee that I address the variables in proper order. I just address them in the reverse order presented, which works one a single pass for all the list members in the unit test in this PR (which test was written by Mike). But it seems like the vars are present in execution order in cases I've looked at so far, so I suspect it's not entirely accidental. (It does include paths and such in the processing, for example. I just haven't looked closely.)

### DIFF
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/CleanupTemporariesTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/CleanupTemporariesTest.kt
@@ -2091,8 +2091,10 @@ class CleanupTemporariesTest {
                     moduleCustomizeHook = { module, isNew ->
                         if (isNew) {
                             module.addEnvironmentBindings(
-                                extraEnvironmentBindings +
-                                    (StagingFlags.moduleResultNeeded to TBoolean.value(moduleResultNeeded)),
+                                extraEnvironmentBindings + mapOf(
+                                    StagingFlags.keepCleanupTemporariesData to TBoolean.valueTrue,
+                                    StagingFlags.moduleResultNeeded to TBoolean.value(moduleResultNeeded),
+                                ),
                             )
                         }
                         if (module.stageCompleted == Stage.SyntaxMacro) {


be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
```diff
@@ -119,6 +119,7 @@ public static long DivSafe(this long a, long b)
         }
 
         private static System.Tuple<object?> emptySingleton = new System.Tuple<object?>(null);
+
         public static System.Tuple<object?> Empty()
```
tjpalmer, line 123: Blank line from csharpier formatting.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
```diff
@@ -140,6 +141,8 @@ public static void InitSimpleLogging()
         /// </summary>
         public static void Garbage(string message) => throw new Exception(message);
 
+        public static void Garbage() => throw new Exception();
+
```
tjpalmer, line 145: I ran across this issue recently, and I'm sneaking a fix in here to avoid another pr.

frontend/src/commonMain/kotlin/lang/temper/frontend/rw/ReadsAndWrites.kt
```diff
@@ -69,8 +69,11 @@ internal data class ReadsAndWrites(
     /**
      * The set of names that are used only in the current function or module body;
      * their liveness does not extend to nested or containing functions.
+     *
+     * While provided as a list for efficient reverse iteration as needed, each
+     * member is expected to be unique.
      */
-    val localNames: Set<ResolvedName>,
+    val localNames: List<ResolvedName>,
```
tjpalmer, line 76: I looked at all uses, and nobody needs it to be a Set. Just iterating on it.

frontend/src/commonMain/kotlin/lang/temper/frontend/CleanupTemporaries.kt
```diff
@@ -809,7 +810,12 @@ internal class CleanupTemporaries private constructor(
         //   For example, the read of `g` is a no-op and `f(x)` cannot change its value,
         //   so we can reorder `f(x)` after it.
 
-        for (name in readsAndWrites.localNames) {
+        // Finally review in reverse order to make it easier to inline a sequence
+        // of assignments all at once. That can help in common degenerate cases like
+        // 1000s of items going into the same list.
+        val inlineds = mutableSetOf<MaximalPath.Element>()
+
+        for (name in readsAndWrites.localNames.asReversed()) {
```
tjpalmer, line 819: Here's where I want reversed, and I'd have to explicitly make java.util.LinkedHashSet instances to use a set and get low cost reversal. Anyway, the idea is that we want to handle things like this in reverse, and our collection manages to get them in order:

```ts
var t#4, t#5, t#6, t#7, fail#0, fail#1, fail#2, fail#3;
let `test//`.numbers: Map<String, Int32>;
t#4 = new Pair<String, Int32>("0", 0);
t#5 = new Pair<String, Int32>("1", 1);
t#6 = new Pair<String, Int32>("2", 2);
`test//`.numbers = new Map<String, Int32>(list(t#4, t#5, t#6));
```

If we do the last one first, then we know that the assignment is a no-op via the `inlineds` set above, and we can inline them all (or 1000 of them) in one pass when organized like this, instead of making N passes (for O(N^2) overall).

frontend/src/commonMain/kotlin/lang/temper/frontend/CleanupTemporaries.kt
```diff
@@ -885,10 +891,10 @@ internal class CleanupTemporaries private constructor(
...
                         // - nym`<>`(callee, TypeActuals) where callee can be reordered over
                         val callee = t.childOrNull(0)?.functionContained
-                        when {
-                            callee == BuiltinFuns.angleFn && t.size >= 2 -> mayReorderOver(t.child(1))
-                            callee is GetStaticOp -> true
-                            callee is DotHelper && callee.memberAccessor is BindMemberAccessor -> {
+                        when (callee) {
+                            BuiltinFuns.angleFn if t.size >= 2 -> mayReorderOver(t.child(1))
+                            is GetStaticOp -> true
+                            is DotHelper if callee.memberAccessor is BindMemberAccessor -> {
```
tjpalmer: Recommended by ij for our updated kotlin version.

frontend/src/commonMain/kotlin/lang/temper/frontend/CleanupTemporaries.kt
```diff
@@ -1235,10 +1243,14 @@ internal class CleanupTemporaries private constructor(
...
 
+                val keepCleanupTemporariesData =
+                    module.stableEnvironmentValue(StagingFlags.keepCleanupTemporariesData) == TBoolean.valueTrue
                 while (true) {
                     val dataTables = cleaner.clean()
                     val madeProgress = dataTables.edits.isNotEmpty()
-                    allDataTables.add(dataTables)
+                    if (keepCleanupTemporariesData) {
+                        allDataTables.add(dataTables)
+                    }
```
tjpalmer, line 1222: Also avoid storing possibly O(N^2) memory by default. This was the out of memory error. Even though I turned the specific case out of O(N^2), seems also good to be cautious about other potential degenerate cases.

frontend/src/commonMain/kotlin/lang/temper/frontend/Module.kt
```diff
@@ -182,6 +182,10 @@ class Module(
         return immutableEnvironment(parent, stableEnvironmentBindings, isLongLived = true)
     }
 
+    fun stableEnvironmentValue(name: TemperName): Value<*>? {
+        return stableEnvironmentBindings[name]
+    }
```
tjpalmer, line 187: Needed for easy access to staging flags.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -3663,6 +3663,23 @@ class GenerateCodeStageTest {
...
+                append("  new Pair<String, Int>(\"$i\", $i),")
+            }
+            append("]);")
+        },
+        want = """
+            |{
+            |  run: "void: Void",
+            |}
+        """.trimMargin(),
+    )
```
tjpalmer, line 3682: Typically finishes in under 30 seconds on my machine, so hopefully fine here. Usually takes about 10 to 15 seconds longer than a short list of pairs (say 10 items). Update: Although I've also now seen it finish in 10 seconds total. I don't know what's normal to expect.

frontend/src/commonMain/kotlin/lang/temper/frontend/CleanupTemporaries.kt
```diff
@@ -876,39 +883,6 @@ internal class CleanupTemporaries private constructor(
                 // If the preceding, non-containing trees do not have a
                 // side effect then we can replace READ with the
                 // right-hand-side of the upstream write.
-                fun mayReorderOver(t: Tree): Boolean = when (t) {
```
tjpalmer, line 879: Moved this when I thought I'd need to reuse it. Didn't need to, but left it moved because it feels clearer when external.

frontend/src/commonMain/kotlin/lang/temper/frontend/CleanupTemporaries.kt
```diff
@@ -921,7 +895,7 @@ internal class CleanupTemporaries private constructor(
                     val parent = edge.source!! // Safe since we don't go above root
                     for (precedingSiblingIndex in 0 until indexInParent) {
                         val precedingSibling = parent.child(precedingSiblingIndex)
-                        if (!mayReorderOver(precedingSibling)) {
+                        if (precedingSibling in inlinedReadTrees || !mayReorderOver(precedingSibling, readsAndWrites)) {
```
tjpalmer, line 898: We don't always know when previous inlines might have side effects, so don't change order on them.

frontend/src/commonMain/kotlin/lang/temper/frontend/CleanupTemporaries.kt
```diff
@@ -1540,3 +1521,37 @@ private fun mustRemainAtStatementLevel(t: Tree): Boolean {
 
 fun isGetStaticCall(t: Tree) = t is CallTree &&
     t.childOrNull(0)?.functionContained is GetStaticOp
+
+private fun mayReorderOver(t: Tree, readsAndWrites: ReadsAndWrites): Boolean = when (t) {
```
tjpalmer, line 1525: Had to introduce `readsAndWrites` to move this down. Something even more generalized would be nice in the future, especially if we get better at knowing when there might be side effects.

frontend/src/commonTest/kotlin/lang/temper/frontend/CleanupTemporariesTest.kt
```diff
@@ -2091,8 +2091,10 @@ class CleanupTemporariesTest {
                     moduleCustomizeHook = { module, isNew ->
                         if (isNew) {
                             module.addEnvironmentBindings(
-                                extraEnvironmentBindings +
-                                    (StagingFlags.moduleResultNeeded to TBoolean.value(moduleResultNeeded)),
+                                extraEnvironmentBindings + mapOf(
+                                    StagingFlags.keepCleanupTemporariesData to TBoolean.valueTrue,
+                                    StagingFlags.moduleResultNeeded to TBoolean.value(moduleResultNeeded),
+                                ),
```
tjpalmer, line 2097: Added keepCleanupTemporariesData because needed for the tests here.

PR #323, opened 2026-02-07
be-js: Stop adding return type to garbage expressions

tjpalmer:
- Avoid TypeScript `never` return type in js garbage expressions
- Ran across this when I had a bug temporarily in some code I was working on
- Also, TypeScript can infer the type without it:

image

### DIFF
--- a/be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
+++ b/be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
@@ -1307,13 +1307,14 @@ class JsBackendTest {
                 |}
             """.trimMargin(),
         ),
+        // This had an explicit TypeScript-style `never` return type on it, but we should use just JS.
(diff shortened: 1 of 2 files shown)


be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
```diff
@@ -1295,6 +1294,44 @@ class JsBackendTest {
...
+            """.trimMargin(),
+        ),
+        // This had an explicit TypeScript-style `never` return type on it, but we should use just JS.
+        want = """
+            |{
+            |  js: {
+            |    "my-test-library": {
+            |      "foo.js": {
+            |        content: ```
+            |          ((() => {
```
tjpalmer, line 1317: See the first commit of this branch for the `(((): never => {` that was here. We don't want needless parsing errors even if some branches could fail at runtime.

PR #325, opened 2026-02-09
be-rust: Avoid use trait from current module

tjpalmer:
- Fix bogus trait `use` from same module
- This bug is a compiler error in rust, so it's good to fix

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -1524,7 +1524,10 @@ class RustTranslator(
                     is TmpL.Expression -> {
                         callable.method?.enclosingType?.let traitImport@{ owner ->
                             // TODO Do we need to filter out implicits? So far, no examples of hitting that case.
-                            if (owner.abstractness == Abstractness.Abstract && owner.pos.loc != module.pos.loc) {
+                            if (
+                                owner.abstractness == Abstractness.Abstract &&
+                                owner.sourceLocation != module.codeLocation.codeLocation
+                            ) {
                                 // For actual interface methods, we shouldn't have name collisions.
                                 // TODO Can/should we provide traits for extension methods?
                                 // TODO If so, just translate those as qualified calls anyway?
--- a/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
+++ b/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
@@ -128,7 +128,6 @@ class RustBackendTest {
             |              use temper_core::AnyValueTrait;
             |              use temper_core::AsAnyValue;
             |              use temper_core::Pair;
-            |              use TalkerTrait;
             |              pub (crate) fn init() -> temper_core::Result<()> {
             |                  static INIT_ONCE: std::sync::OnceLock<temper_core::Result<()>> = std::sync::OnceLock::new();
             |                  INIT_ONCE.get_or_init(| |{


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1524,7 +1524,10 @@ class RustTranslator(
                     is TmpL.Expression -> {
                         callable.method?.enclosingType?.let traitImport@{ owner ->
                             // TODO Do we need to filter out implicits? So far, no examples of hitting that case.
-                            if (owner.abstractness == Abstractness.Abstract && owner.pos.loc != module.pos.loc) {
+                            if (
+                                owner.abstractness == Abstractness.Abstract &&
+                                owner.sourceLocation != module.codeLocation.codeLocation
+                            ) {
```
tjpalmer, line 1530: Before would see different source files in the same module as different. I think this is now detecting actually different modules correctly.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -25,7 +25,8 @@ class RustBackendTest {
                 |      let d = nums.join("", stringifyValueHere);
                 |      "${a}${b}${c}${d}"
                 |    }
-                |    let { stringify, stringifyValue } = import("./bar");
+                |    export let makeTalkHere(talker: Talker): Void { talker.talk(); }
+                |    let { stringify, stringifyValue, Talker } = import("./bar");
```
tjpalmer, line 29: Here's using a trait from a different module.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -34,9 +35,13 @@ class RustBackendTest {
                 |      export let stringify(i: Int): String { i.toString() }
                 |      export let stringifyValue = stringify;
                 |      console.log("Baz")
+                |      export interface Talker {
+                |        talk(): Void;
+                |      }
                 |      ```,
                 |    boo.temper: ```
                 |      console.log("Boo")
+                |      export let makeTalk(talker: Talker): Void { talker.talk(); }
```
tjpalmer, line 44: And here's using from a different source file in the same module.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -85,6 +90,7 @@ class RustBackendTest {
             |            use temper_core::AnyValueTrait;
             |            use temper_core::AsAnyValue;
             |            use temper_core::Pair;
+            |            use crate::bar::TalkerTrait;
```
tjpalmer, line 93: From another module needs explicit `use` for direct method usage.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -132,9 +141,30 @@ class RustBackendTest {
...
+            |                  fn deref(& self) -> & Self::Target {
+            |                      & ( * self.0)
+            |                  }
+            |              }
             |              pub fn stringify(i__1: i32) -> std::sync::Arc<String> {
             |                  return temper_core::int_to_string(i__1, None);
             |              }
+            |              pub fn make_talk(talker__0: Talker) {
+            |                  talker__0.talk();
+            |              }
```
tjpalmer, line 167: Here's using from the same module but [without a bogus `use` now](https://github.com/temperlang/temper/pull/325/changes/d1d93ba0a3d4a237ce5827b5022de492ad993cd9#diff-1d412556c62f2370df35d85d2bce101644aca5a39e1f46343bc076d185e3bdc3L131).

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
mikesamuel: Possibly fallout from the directory modules everywhere transition.

PR #327, opened 2026-02-11
Fix lru cache entries missing next

tjpalmer:
- The lru cache bug happened to me when I had duplicated definitions in a module
- Also fix rust translator crash on bad imports

### DIFF
--- a/common/src/commonTest/kotlin/lang/temper/common/LruCacheTest.kt
+++ b/common/src/commonTest/kotlin/lang/temper/common/LruCacheTest.kt
@@ -34,7 +34,7 @@ class LruCacheTest {
 private fun churnCache(
     prng: Random,
     maxSize: Int,
-    randUntil: Int
+    randUntil: Int,
 ): Pair<MutableMap<Int, String>, MutableList<Int>> {
     val keys = mutableListOf<Int>()
     val lruCache = lruCacheWithSize<Int, String>(maxSize = maxSize)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -222,7 +222,7 @@ class RustTranslator(
 
     private fun preprocessImports() {
         imports@ for (import in module.imports) {
-            val externalName = import.externalName.name as ExportedName
+            val externalName = import.externalName.name as? ExportedName ?: continue@imports
             val export = (externalName.origin as? ModuleNamingContext)?.owner?.exportMatching(externalName)
             val name = import.localName?.name ?: continue@imports
```
tjpalmer, line 227: The above crashed on a case trying to use a non-exported symbol from another module.

common/src/commonMain/kotlin/lang/temper/common/LruCache.kt
```diff
@@ -25,9 +26,13 @@ private class LruCache<K, V>(private val maxSize: Int) : MutableMap<K, V> {
             } else {
                 last = oldPrev
             }
+            // Plug in e as new last.
             e.next = null
             e.prev = oldLast
             last = e
+            oldLast.next = e
```
tjpalmer, line 33: This line is the fix. Without this, lots of entries had `next` of `null`, which apparently only rarely matters for us.

common/src/commonMain/kotlin/lang/temper/common/LruCache.kt
```diff
@@ -25,9 +26,13 @@ private class LruCache<K, V>(private val maxSize: Int) : MutableMap<K, V> {
...
+            // And if it's lonely, also as the first.
+            // (Although it presumably would already have been last in this case.)
             if (first == null) {
                 first = e
             }
```
tjpalmer, line 38: Also added comments to help understand the flow better.

common/src/commonTest/kotlin/lang/temper/common/LruCacheTest.kt
```diff
@@ -31,4 +23,28 @@ class LruCacheTest {
             assertEquals("$key", lruCache[key])
         }
     }
+
+    @Test
+    fun edginess() = withRandomForTest { prng ->
+        // This case that increases repetition was just plain crashing before.
+        churnCache(prng, maxSize = 3, randUntil = 10)
+    }
```
tjpalmer, line 31: I had trouble constructing a short simple case to cause the crash, but this one does it. I probably could have constructed something explicit if I thought harder about it.

Example crash from before the fix:

```kt
java.lang.NullPointerException
	at lang.temper.common.LruCache.moveToEnd(LruCache.kt:15)
	at lang.temper.common.LruCache.put(LruCache.kt:81)
	at lang.temper.common.LruCacheTestKt.churnCache(LruCacheTest.kt:44)
	at lang.temper.common.LruCacheTestKt.access$churnCache(LruCacheTest.kt:1)
	at lang.temper.common.LruCacheTest.edginess$lambda$4(LruCacheTest.kt:30)
	at lang.temper.common.RandomForTestKt.withRandomForTest(RandomForTest.kt:36)
	at lang.temper.common.RandomForTestKt.withRandomForTest(RandomForTest.kt:24)
	at lang.temper.common.LruCacheTest.edginess(LruCacheTest.kt:28)
```

PR #329, opened 2026-02-12
fix(rust): traverse NestingStatement in varsDeclared for mutable capture detection

### LINKED ISSUE #328: (bug): Rust compilation of mutable vars used in for loops inside `when` on a pattern match produce invalid Rust
Temper's Rust codegen emits `fn` items inside func bodies for when branches. These `fn` items reference the outer mutable var first, but Rust `fn` items cannot capture vars from their enclosing scope. Only closures can do that.

We end up with code like this

```rust
fn when_branch_0(first: &bool, out: &mut StringBuilder, item: &FooItem) {
    // ...references `first` from outer scope...
}
```

One would expect either use closures, or for it to pass the var explicitly without generating `fn` items that ref outer scope.

I have created a repository with Claude's help to minimally reproduce this problem [here](https://github.com/notactuallytreyanastasio/temper_rust_bug_maybe)

notactuallytreyanastasio:
# attempts to fix #328

## Note: I am a noob and I had Claude's help here, but I uncovered the bug in #328 and replicated it [here](https://github.com/notactuallytreyanastasio/temper_rust_bug_maybe)

The original Rust compiler had 13 total errors of two types:

```
  E0434 — "can't capture dynamic environment in a fn item" (7 occurrences)
```

```
  error[E0434]: can't capture dynamic environment in a fn item
      --> src/mod.rs:4357:21
       |
  4357 |                     first__929 = false;
       |                     ^^^^^^^^^^
       |
       = help: use the `|| { ... }` closure form instead
```

The generated Rust that I got had a `fn` method inside a `ClosureGroup` struct that tried to assign to `first__929` using the bare var name instead of `self.first__929`. Rust `fn` items can't capture that enclosing scope.

Also:
```
  E0308 — "mismatched types: expected &Arc<RwLock<bool>>, found &bool" (6 occurrences)
```

```
  error[E0308]: mismatched types
      --> src/mod.rs:4354:52
       |
  4354 |                     if ! temper_core::read_locked( & self.first__929) {
       |                          ------------------------  ^^^^^^^^^^^^^^^^^ expected `&Arc<RwLock<bool>>`, found `&bool`
       |                          |
       |                          arguments to this function are incorrect
       |
       = note: expected reference `&Arc<std::sync::RwLock<bool>>`
                  found reference `&bool`
```

---

`varsDeclared()` skipped `NestingStatement` nodes (`when` blocks, `if`/`else`, `while` loops), but `functionsDeclared()` did not. This asymmetry caused `mutableCaptures()` to find closures inside `when` blocks but not the mutable vars they capture.

The Rust BE wraps mutable captured variables in `Arc<RwLock<T>>` @ dec time + generates `read_locked()/write().unwrap()` for access.

When the var wasn't detected as a mutable capture, it was dec'd as a plain type (like `bool` or whatever), but the closure still shat out `Arc<RwLock<bool>>` access patterns — causing `E0308` type mismatches and `E0434` `fn-item` capture errors.

This attempted fix adds an `includeNesting` parameter to `varsDeclared()` and pass it from `mutableCaptures()`. When traversing into nesting statements, `includeBlocks` is also enabled so block bodies inside match arms are explored.

This fixes 13 Rust compilation errors in projects using `var` + `for` + `when` pattern.

---

### The JS/Python/Lua backends remain unaffected (shared analysis code, not codegen).

CC @mikesamuel @tjpalmer @ShawSumma I'm sure there are mistakes here or like some bad style but I think I might have it kind of right?

Obviously this needs a serious review from y'all and this is conjecture at this point. 

# Update

This also seemingly uncovered another small bug.

The old expected output for the test I was modifying was buggy I think.

It had bare a__0 = m__0.clone() inside a fn item — that's the exact E0434 error the fix addresses. The old test was asserting wrong Rust output I think?

My fix exposed this and the expected output now shows correct `Arc<RwLock<...>>` wrapping that we'd want.

```
  export let hi(n: Int): Void {
    for (var i = 0; i < n; i += 1) {
      var a = "${i}";
      let blah(m: String): Void {
        a = m;    // captures mutable var a
      }
      blah("hi");
    }
  }
```

This was the old test function capturing that mutable a

The old expected Rust had

```
  impl ClosureGroup___0 {
      fn blah__0(& self, m__0: impl temper_core::ToArcString) {
          let m__0 = m__0.to_arc_string();
          a__0 = m__0.clone();  // <-- bare a__0, not self.a__0
      }
  }
```
fn items can't capture from enclosing scope.

Commit 8c1646f addresses this

### DIFF
--- a/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
+++ b/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
@@ -410,15 +410,17 @@ class RustBackendTest {
             |    let mut i__0: i32 = 0;
             |    'loop___0: while Some(i__0) < Some(n__0) {
             |        t___0 = temper_core::int_to_string(i__0, None);
-            |        let mut a__0: std::sync::Arc<String> = std::sync::Arc::new(format!("{}", t___0));
+            |        let mut a__0: std::sync::Arc<std::sync::RwLock<std::sync::Arc<String>>> = std::sync::Arc::new(std::sync::RwLock::new(std::sync::Arc::new(format!("{}", t___0))));
             |        #[derive(Clone)]
             |        struct ClosureGroup___0 {
-            |            a__0: std::sync::Arc<String>
+            |            a__0: std::sync::Arc<std::sync::RwLock<std::sync::Arc<String>>>
             |        }
             |        impl ClosureGroup___0 {
             |            fn blah__0(& self, m__0: impl temper_core::ToArcString) {
             |                let m__0 = m__0.to_arc_string();
-            |                a__0 = m__0.clone();
+            |                {
+            |                    * self.a__0.write().unwrap() = m__0.clone();
+            |                }
             |            }
             |        }
             |        let closure_group = ClosureGroup___0 {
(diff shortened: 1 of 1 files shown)


be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -440,15 +410,17 @@ class RustBackendTest {
...
-            |            a__0: std::sync::Arc<String>
+            |            a__0: std::sync::Arc<std::sync::RwLock<std::sync::Arc<String>>>
             |        }
             |        impl ClosureGroup___0 {
             |            fn blah__0(& self, m__0: impl temper_core::ToArcString) {
             |                let m__0 = m__0.to_arc_string();
-            |                a__0 = m__0.clone();
+            |                {
+            |                    * self.a__0.write().unwrap() = m__0.clone();
+            |                }
```
tjpalmer, line 423: Yeah, turns out this was already a test that we were doing wrong, as indicated by the name `captureBad`. I think I just didn't work through the issue before.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -2000,6 +1972,104 @@ class RustBackendTest {
             """.trimMargin(),
         )
     }
+
+    @Test
+    fun mutableVarInWhenBranch() = assertGenerateWanted(
```
tjpalmer, line 1977: So far, I've left this out of #330 because I think the other turns out already to be a test.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -450,7 +455,7 @@ fun TmpL.Tree.varsDeclared(
     }
     for (kid in children) {
         // Exclude nested blocks in any case.
-        yieldAll(kid.varsDeclared(keep = keep))
+        yieldAll(kid.varsDeclared(includeBlocks = includeNesting, includeNesting = includeNesting, keep = keep))
```
tjpalmer, line 458: I think the `includeNesting` idea was the intent of the existing `includeBlocks`, so I combined them together in #330. I'm unsure why the comment "Exclude nested blocks in any case" above, as things seem to work recursively passing down includeNesting as you do here and also in #330.

mikesamuel:
Thanks for the patch. Tom's right that he's the right one to look at it as he's the primary be-rust author.

PR #330, opened 2026-02-12
Find capture vars recursively through nestings

### LINKED ISSUE #328: (bug): Rust compilation of mutable vars used in for loops inside `when` on a pattern match produce invalid Rust

tjpalmer:
- Fix #328
- Based on #329
- I also tested locally that the repro repo now runs without error:

```
PS C:\Users\work2\Documents\projects\old\temper_rust_bug_maybe\heex> git log -1
commit d7e774535c7244c92219ed33a2039eaf2bcd505b (HEAD -> main, upstream/main)
Author: Robert Grayson <bobbbygrayson+github@gmail.com>
Date:   Wed Feb 11 19:54:29 2026 -0500

    better writeup
PS C:\Users\work2\Documents\projects\old\temper_rust_bug_maybe\heex> temper test -b rust
Tests passed: 3 of 3
PS C:\Users\work2\Documents\projects\old\temper_rust_bug_maybe\heex>
```

### DIFF
--- a/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
+++ b/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
@@ -416,7 +416,7 @@ class RustBackendTest {
     )
 
     @Test
-    fun captureBad() = assertGenerateWanted(
+    fun captureUnderNesting() = assertGenerateWanted(
         temper = $$"""
             |export let hi(n: Int): Void {
             |  for (var i = 0; i < n; i += 1) {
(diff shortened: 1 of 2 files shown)


be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -416,7 +416,7 @@ class RustBackendTest {
     )
 
     @Test
-    fun captureBad() = assertGenerateWanted(
+    fun captureUnderNesting() = assertGenerateWanted(
```
tjpalmer, line 419: Changed the name since it's not bad anymore.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -429,15 +429,15 @@ fun TmpL.Tree.functionsDeclared(): Sequence<TmpL.FunctionLike> = sequence {
...
     keep: (TmpL.VarLike) -> Boolean = { true },
 ): Sequence<ResolvedName> = sequence {
     when (this@varsDeclared) {
-        is TmpL.BlockStatement -> if (!includeBlocks) {
+        is TmpL.NestingStatement -> if (!includeNesting) {
             return@sequence
         }
 
-        is TmpL.FunctionLike, is TmpL.NestingStatement -> {
+        is TmpL.FunctionLike -> {
```
tjpalmer, line 440: Generalize from blocks to nestings.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -449,8 +449,7 @@ fun TmpL.Tree.varsDeclared(
         else -> {}
     }
     for (kid in children) {
-        // Exclude nested blocks in any case.
-        yieldAll(kid.varsDeclared(keep = keep))
+        yieldAll(kid.varsDeclared(includeNesting = includeNesting, keep = keep))
```
tjpalmer, line 452: I'm unsure why I didn't want to recurse including blocks before, but tests still pass with this change, and the reported bug gets fixed, which makes sense.

PR #331, opened 2026-02-13
Overload on generics

tjpalmer:
- Just add extra checks when types are well known enough to respect type args also
- New check gives up for now when types themselves are type formals
  - Mostly an issue of not finding good existing helpers for this case
  - I'd rather punt than spend lots of time now, but we should generalize better later

### DIFF
--- a/fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
+++ b/fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
@@ -893,8 +893,8 @@ class TypeSolver(
                 //
                 // If an actual is nullable and a formal isn't, the callee doesn't fit.
                 //
-                // If an actual has a lower or common bound that doesn't have an
-                // inheritance chain to shapes required by the formal, it isn't a fit.
+                // If an actual has a lower or common bound that isn't compatible with
+                // the formal, it isn't a fit.
                 //
                 // We look at the argument expression's common and lower bounds
                 // (and choices) but not upper bounds because, as the below shows,


frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2064,20 +2064,20 @@ class TypeStageTest {
...
+            |  public int32ToInt(int: Int32): Int32 { int }
             |}
             |
             |export let crazySum(intMaker: IntMaker, int: Int64, string: String): Int throws Bubble {
-            |   let intInt = intMaker.toInt(int);
-            |   let stringInt = intMaker.toInt(string);
-            |   intMaker.justMe(intInt + stringInt)
+            |  let intInt = intMaker.toInt(int);
+            |  let stringInt = intMaker.toInt(string);
+            |  intMaker.justMe(intInt + stringInt)
```
tjpalmer, line 2080: Just fixing old indentation.

frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2150,6 +2150,66 @@ class TypeStageTest {
...
+            |  public stringifyInt32List(ints: Listed<Int32>): String;
+            |
+            |  @overload("stringify")
+            |  public stringifyStringList(string: Listed<String>): String;
+            |}
+            |
+            |// Purposely receive List but use as Listed above.
+            |export let stringifyLists(stringer: Stringer, int: Int, ints: List<Int>, strings: List<String>): String {
+            |  "${stringer.stringify(int)}, ${stringer.stringify(ints)}, ${stringer.stringify(strings)}"
+            |}
```
tjpalmer, line 2172: Overload on list type arg.

frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2150,6 +2150,66 @@ class TypeStageTest {
...
+            |      @visibility(\public) @overload("stringify") @fn @stay @fromType(Stringer) let stringifyStringList__0 ⦂(fn (Stringer, Listed<String>): String);
+            |      stringifyStringList__0 = fn stringifyStringList(@impliedThis(Stringer) this__2: Stringer, string__0 /* aka string */: Listed<String>) /* return__2 */: String {
+            |        fn__2: do {
+            |          pureVirtual ⋖ String ⋗()
+            |        }
+            |      };
+            |      `test//`.stringifyLists = fn stringifyLists(stringer__0 /* aka stringer */: Stringer, int__1 /* aka int */: Int32, ints__1 /* aka ints */: List<Int32>, strings__0 /* aka strings */: List<String>) /* return__3 */: String {
+            |        void;
+            |        fn__3: do {
+            |          return__3 = cat(do_bind_stringifyInt32(stringer__0)(int__1), ", ", do_bind_stringifyInt32List(stringer__0)(ints__1), ", ", do_bind_stringifyStringList(stringer__0)(strings__0))
```
tjpalmer, line 2203: Got the matches.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
```diff
@@ -893,10 +893,8 @@ class TypeSolver(
                 //
                 // If an actual is nullable and a formal isn't, the callee doesn't fit.
                 //
-                // If an actual has a lower or common bound with a type shape
-                // that doesn't have an inheritance chain to shapes required by the formal,
-                // it isn't a fit.
-                // This is equivalent to generic erasure argument analysis.
+                // If an actual has a lower or common bound that doesn't have an
+                // inheritance chain to shapes required by the formal, it isn't a fit.
```
tjpalmer: Removed erasure comment.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
```diff
@@ -930,12 +928,16 @@ class TypeSolver(
                     // Then we'll check that we can find any declared class/interface
                     // types in the argument declaration and make sure we have a path to them.
                     class ShapeAndNullityInfo {
+                        // Track both types and erased shapes for now.
+                        // TODO Improve checking on the types themselves so we don't need separate shapes.
+                        val types = mutableSetOf<TypeOrPartialType>()
```
tjpalmer, line 933: Just add full types along with shapes for now.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
```diff
@@ -944,7 +946,7 @@ class TypeSolver(
                                     // TODO: avoid cycles
                                     // TODO: set out.canBeNull if any super-type is
                                     // nullable.
-                                    unpackTypeShapes(it.definition, out)
+                                    unpackTypeShapes(null, it.definition, out)
```
tjpalmer, line 949: No Type2 here to pass along.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
```diff
@@ -953,7 +955,7 @@ class TypeSolver(
                     val actualInfo = ShapeAndNullityInfo().also { info ->
                         when (val argChoice = argNode.choice) {
                             is TypeOrPartialType -> {
-                                unpackTypeShapes(argChoice.definition, info)
+                                unpackTypeShapes(argChoice, argChoice.definition, info)
```
tjpalmer, line 958: I guess I could automate the definition, but meh.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
```diff
@@ -999,17 +1001,23 @@ class TypeSolver(
...
+                                // This part checks only erased generics.
                                 typeShapesInFormal.all { formalShape ->
                                     typeContext.extendsPath(actualShape, formalShape) != null
                                 }
+                            } && actualInfo.types.all { actualType ->
+                                // This part cares about generic type args.
+                                formalInfo.types.all { formalType ->
+                                    isMaybeCompatibleSubtype(actualType, formalType)
+                                }
                             }
```
tjpalmer, line 1021: Here's the only new real logic. And if I include only this, I get more things unsolvable. I expect it's because the other branch handles cases better there the main time is a type formal.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
tjpalmer, line 1033: I think the `<= 1` thing isn't ideal, because it might be something we haven't actually checked here yet, and it might also be incompatible. But when I changed this to `< 1`, a different assertion elsewhere failed. It had an Unsolvable for a different var with an actual choice. Not sure how that works, so I put this back at `<=`.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
```diff
@@ -1549,6 +1557,24 @@ class TypeSolver(
...
+     * TODO Improve tightness on the matching.
+     */
+    private fun isMaybeCompatibleSubtype(
+        actualType: TypeOrPartialType,
+        formalType: TypeOrPartialType,
+    ): Boolean = when (actualType) {
+        // Type params are awkward because upper bounds aren't Type2. TODO Is some better helper available?
+        is TypeParamRef -> true // TODO limit by bounds
+        is Type2 -> when (formalType) {
+            is TypeParamRef -> true // TODO limit by bounds
```
tjpalmer, line 1571: I couldn't find a straightforward way to check this against bounds. If there's something easy I missed, we should do that. I've it's lots of algorithm design time, I'd rather punt.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
```diff
@@ -1549,6 +1557,24 @@ class TypeSolver(
...
+            is Type2 -> typeContext.isSubType(actualType, formalType)
+            else -> null
+        }
+        else -> null
+    } ?: typeContext.isSubTypeOptimistic(actualType, formalType)
```
tjpalmer, line 1576: Erring on the side of not rejecting for now here, also.

fundamentals/src/commonTest/kotlin/lang/temper/type2/TypeSolverTest.kt
```diff
@@ -684,7 +685,7 @@ class TypeSolverTest {
                     gPass to "List<Int32>",
                     gFail to "[]",
 
-                    hCallee to "Unsolvable",
+                    hCallee to "1",
```
tjpalmer, line 688: I think both should be rejected. Maybe that would happen in the type checker at generate code stage, though. And we still allow contravariant arg on List, by the way, which we need to stop sometime. But before, both were accepted, which is how it was Unsolvable. Because we didn't look at type args at all.

I spent a while trying to get Unsolvable back here, but I was failing. I had something at some point with 3 overload options, but it also made a different var Unsolvable, and I managed to lose track of what I'd done to get that, anyway.

So I punted and just changed this to 1 now.

fundamentals/src/commonMain/kotlin/lang/temper/type2/TypeSolver.kt
mikesamuel: Could we just say that if actual has a lower or common bound that is not compatible with the formal, it isn't a fit?

PR #332, opened 2026-02-14
SAST transparency for contextual autoescapers

mikesamuel:
This adds an experimental micro-pass for secure composition code that makes translations transparent to static analysis.

Instead of picking escapers at runtime, escapers are picked at compile time and inlined so that SAST tools see the chosen pickers and can compare those to their own expectations due to flow sensitive analysis.

This implements a builtin, `doPure { ... }` which allows select complex but pure initializers to be run at compile time and so available to macros that invoke the interpreter to predict runtime behaviour.

### DIFF
--- a/builtin/src/commonMain/kotlin/lang/temper/builtin/DoPureFn.kt
+++ b/builtin/src/commonMain/kotlin/lang/temper/builtin/DoPureFn.kt
@@ -9,6 +9,7 @@ import lang.temper.type.MkType
 import lang.temper.type2.Signature2
 import lang.temper.type2.hackMapOldStyleToNew
 import lang.temper.value.Fail
+import lang.temper.value.FunTree
 import lang.temper.value.MacroEnvironment
 import lang.temper.value.NamedBuiltinFun
 import lang.temper.value.PartialResult
@@ -78,6 +79,20 @@ object DoPureFn : SpecialFunction, NamedBuiltinFun {
             // Erase this call so we don't have to translate it.
             val argTree = args.valueTree(0)
             macroEnv.replaceMacroCallWith {
+                val parts = (argTree as? FunTree)?.parts
+                if (parts != null && parts.formals.isEmpty()) {
+                    val returnDecl = parts.returnDecl
+                    val returnName = returnDecl?.parts?.name?.content
+                    if (returnName != null) {
+                        Block(argTree.pos) {
+                            Replant(freeTree(returnDecl))
+                            Replant(freeTree(parts.body))
+                            Rn(argTree.pos.rightEdge, returnName)
+                        }
+                        return@replaceMacroCallWith
+                    }
+                }
+
                 Call(macroEnv.pos) {
                     Replant(freeTree(argTree))
                 }
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
@@ -3701,11 +3701,9 @@ class GenerateCodeStageTest {
             |    body: ```
             |        let return__0, @stay @imported(\(`test//c/`.C)) C__0;
             |        C__0 = type (C);
-            |        let c__0, fn__0;
-            |        fn__0 = (@stay fn /* return__1 */: C {
-            |            return__1 = new C()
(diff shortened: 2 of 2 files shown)


builtin/src/commonMain/kotlin/lang/temper/builtin/StringExprMacro.kt
```diff
@@ -532,9 +532,9 @@ private fun pointAppendsAtAccumulator(funTree: FunTree, isTagged: Boolean) {
     }
 }
 
-private val appendSafeDotName = Symbol("appendSafe")
-private val appendDotName = Symbol("append")
-private val accumulatedDotName = Symbol("accumulated")
+val appendSafeDotName = Symbol("appendSafe")
+val appendDotName = Symbol("append")
+val accumulatedDotName = Symbol("accumulated")
```
tjpalmer, line 537: And I still want to change to build/built.

frontend/src/commonMain/kotlin/lang/temper/frontend/function/OptimizeContextualAutoescapingBlocks.kt
```diff
@@ -0,0 +1,715 @@
...
+ *       ...
+ *       // accumulator.appendSafe("...");
+ *       collector.appendSafe("...");
+ *       ...
+ *       // accumulator.accumulated
+ *       SafeHtmlBuilder.fromCollector(collector)
+ *     ...
+ *
+ * That rewrite relies on a flow-sensitive propagation of contexts & automatonStacks
+ * across safe parts and inlining of escaper decisions for non-safe parts.
```
tjpalmer, line 117: I was wondering if what you were writing was general or not. I guess not, but that's ok.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
```diff
@@ -1017,7 +1017,7 @@ private val sharedStdModulesMayNotRun = lazy {
         CustomValueFormatter.Nope,
     )
     val advancer = ModuleAdvancer(logSink)
-    buildStdModules(advancer, console, mayRun = false)
+    buildStdModules(advancer, console, mayRun = true)
```
tjpalmer, line 1020: This seems like a big change.

builtin/src/commonMain/kotlin/lang/temper/builtin/DoPureFn.kt
```diff
@@ -0,0 +1,106 @@
...
+import lang.temper.value.freeTree
+import lang.temper.value.unpackOrFail
+
+/**
+ * <!-- snippet: builtin/doPure -->
+ * # doPure
+ *
+ * `doPure { x }` is the same as `x` except that the result will be available to
+ * macro code early.  `x` must be pure code: it can't mutate any environment bindings
+ * or objects it does not create and may not print or cause other side effects.
```
tjpalmer, line 28: Or else the side effects won't be replicated in live code? Worth stating directly? Do we hope to validate later if not yet?

Maybe `doConst` or `makeConst` as an alternative if we never plan to check. I'm also ok with `doPure`, though. Just thinking about options.

mikesamuel: Yeah.  Let's workshop names and guarantees on Wednesday.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
mikesamuel: It allows using regular expressions from within macro code.

I thought about trying to minimize this, maybe by converting the transition table regexen to code.  I want to keep the translated code size for the transition functions small though.

tjpalmer: > It allows using regular expressions from within macro code.
> 
> I thought about trying to minimize this, maybe by converting the transition table regexen to code. I want to keep the translated code size for the transition functions small though.

This change causes #338, for which we apparently don't have any proper testing right now, since even integration tests pass. Maybe we need to test multiple rounds of watch on something that actually uses std? Not sure.

PR #333, opened 2026-02-16
Sanitize junit xml before parsing

tjpalmer:
- Fix backends that don't escape control chars in junit xml output
- Here's the breakdown I see for current backends and the testing frameworks we use there:
  - C\# escapes as `\u0000`
  - Java escapes as `&amp;#0;`
  - JS maybe deletes them entirely and other whitespace below 0x20 to space?
  - As of this PR, interpreter escapes as `[0x0]` for control chars and 0xFFFE and 0xFFFF, though not all the other risky chars in later escaping
  - Others pass raw control chars through unchanged

### DIFF
--- a/frontend/src/commonMain/resources/std/testing/testing.temper.md
+++ b/frontend/src/commonMain/resources/std/testing/testing.temper.md
@@ -199,13 +199,22 @@ XML text node or attribute value.
       let end = s.end;
       var emitted = String.begin;
       for (var i = String.begin; i < end; i = s.next(i)) {
-        let esc = when (s[i]) {
+        let c = s[i];
+        let esc = when (c) {
           char'&' -> "&amp;";
           char'<' -> "&lt;";
           char'>' -> "&gt;";
           char"'" -> "&#39;";
           char'"' -> "&#34;";
-          else -> continue;
+          char'\n', char'\r', char'\t' -> continue;
+          else -> if (c < char' ' || c == 0xFFFE || c == 0xFFFF) {
+            // Illegal chars under XML 1.0, but also possibly trouble, so extra
+            // escape them, in a way people aren't likely to accidentally
+            // unescape.
+            "[0x${c.toString(16)}]"
+          } else {
+            continue;
+          }
         }
         sb.appendBetween(s, emitted, i);
         sb.append(esc);
--- a/result-helpers/src/commonMain/kotlin/lang/temper/result/junit/JUnitResultParser.kt
+++ b/result-helpers/src/commonMain/kotlin/lang/temper/result/junit/JUnitResultParser.kt
@@ -4,6 +4,8 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.serializer
+import lang.temper.common.DECIMAL_RADIX
+import lang.temper.common.HEX_RADIX
 import nl.adaptivity.xmlutil.ExperimentalXmlUtilApi
 import nl.adaptivity.xmlutil.XmlDeclMode
 import nl.adaptivity.xmlutil.serialization.DefaultXmlSerializationPolicy
@@ -21,7 +23,7 @@ fun parseJunitResults(input: String?): JUnitResults {
     // Sanitize risky chars already escaped in xml source.
     val inputSansEscapedRisks = input.replace(Regex("&#(x?)([0-9A-Fa-f]+);")) { match ->
         val isHex = match.groupValues[1] == "x"
-        val code = match.groupValues[2].toInt(if (isHex) 16 else 10)
+        val code = match.groupValues[2].toInt(if (isHex) HEX_RADIX else DECIMAL_RADIX)
         when {
             isRiskyChar(code) -> sanitizeRiskyChar(code)
             else -> match.value
@@ -42,13 +44,17 @@ fun parseJunitResults(input: String?): JUnitResults {
     return JUnitResults(suites = results.suites)
 }
 
+@Suppress("MagicNumber")
 private fun isRiskyChar(code: Int): Boolean {
-    // Apparently some of these are legal in XML 1.1, but might as well be aggressive.
+    // Some of these are illegal in at least some versions of XML, and some are just discouraged.
+    // https://www.w3.org/TR/WD-xml-970807#NT-Char
+    // https://en.wikipedia.org/wiki/Valid_characters_in_XML
+    // This list could get longer still.
     return (code < 0x20 && code != 0x09 && code != 0x0A && code != 0x0D) ||
         (code in 0x7F..0x84) || (code in 0x86..0x9F) || (code == 0xFFFF) || (code == 0xFFFE)
 }
 
-private fun sanitizeRiskyChar(code: Int) = "[0x${code.toString(16).uppercase()}]"
+private fun sanitizeRiskyChar(code: Int) = "[0x${code.toString(HEX_RADIX).uppercase()}]"
 
 fun combineSurefireResults(input: Iterable<String>): String {
     val results = input.map { xmlTolerant.decodeFromString<TestSuite>(it) }


frontend/src/commonMain/resources/std/testing/testing.temper.md
```diff
@@ -199,13 +199,22 @@ XML text node or attribute value.
...
           char'>' -> "&gt;";
           char"'" -> "&#39;";
           char'"' -> "&#34;";
-          else -> continue;
+          char'\n', char'\r', char'\t' -> continue;
+          else -> if (c < char' ' || c == 0xFFFE || c == 0xFFFF) {
+            // Illegal chars under XML 1.0, but also possibly trouble, so extra
+            // escape them, in a way people aren't likely to accidentally
+            // unescape.
+            "[0x${c.toString(16)}]"
```
tjpalmer, line 214: This is the same format used for later sanitize-before-parse, but I didn't bother with quite so many code points here. This does get the core illegal-under-xml chars, though.

result-helpers/src/commonMain/kotlin/lang/temper/result/junit/JUnitResultParser.kt
```diff
@@ -18,10 +20,42 @@ fun parseJunitResults(input: String?): JUnitResults {
     if (input.isNullOrBlank()) {
         return JUnitResults(emptyList())
     }
-    val results = xmlTolerant.decodeFromString<TestSuites>(input)
+    // Sanitize risky chars already escaped in xml source.
+    val inputSansEscapedRisks = input.replace(Regex("&#(x?)([0-9A-Fa-f]+);")) { match ->
```
tjpalmer, line 24: I didn't check if other backends hide even char entities like this, but it seemed worth doing for extra safety here.

result-helpers/src/commonMain/kotlin/lang/temper/result/junit/JUnitResultParser.kt
```diff
@@ -18,10 +20,42 @@ fun parseJunitResults(input: String?): JUnitResults {
...
+        val code = match.groupValues[2].toInt(if (isHex) HEX_RADIX else DECIMAL_RADIX)
+        when {
+            isRiskyChar(code) -> sanitizeRiskyChar(code)
+            else -> match.value
+        }
+    }
+    // Sanitize raw risky chars that might even just be illegal. We don't control full test framework path.
+    // TODO Also sanitize up front in our generated failure messaging? I'm mixed on that.
+    val inputSansRawRisks = buildString(inputSansEscapedRisks.length) {
+        for (char in inputSansEscapedRisks) {
```
tjpalmer, line 35: This escapes actual raw control chars and such.

result-helpers/src/commonMain/kotlin/lang/temper/result/junit/JUnitResultParser.kt
```diff
@@ -18,10 +20,42 @@ fun parseJunitResults(input: String?): JUnitResults {
...
+private fun isRiskyChar(code: Int): Boolean {
+    // Some of these are illegal in at least some versions of XML, and some are just discouraged.
+    // https://www.w3.org/TR/WD-xml-970807#NT-Char
+    // https://en.wikipedia.org/wiki/Valid_characters_in_XML
+    // This list could get longer still.
+    return (code < 0x20 && code != 0x09 && code != 0x0A && code != 0x0D) ||
+        (code in 0x7F..0x84) || (code in 0x86..0x9F) || (code == 0xFFFF) || (code == 0xFFFE)
+}
+
+private fun sanitizeRiskyChar(code: Int) = "[0x${code.toString(HEX_RADIX).uppercase()}]"
```
tjpalmer, line 57: And the sanitizing doesn't try to be immediately compatible with standard parsing out there. That's on purpose.

result-helpers/src/commonTest/kotlin/lang/temper/result/junit/JUnitResultParserTest.kt
```diff
@@ -42,6 +42,22 @@ class JUnitResultParserTest {
         assertEquals(3, testResults.testsRun)
         assertEquals(0, testResults.failures.size)
     }
+
+    @Test
+    fun controlCharsSanitized() {
+        val testResults = parseJunitResults(controlChars)
+        val failure = testResults.suites[0].testCases[0].failure!!
+        assertEquals("[0x1][0x2][0x3]", failure.message)
```
tjpalmer, line 50: Escaping that we need to do for sanitizing control chars and such.

result-helpers/src/commonTest/kotlin/lang/temper/result/junit/JUnitResultParserTest.kt
```diff
@@ -42,6 +42,22 @@ class JUnitResultParserTest {
...
+        assertEquals(
+            """
+                |
+                |Character entity escapes don't count in CDATA anyway, but meh?
+                |[0x1][0x2][0x3]
+                |
+            """.trimMargin(),
```
tjpalmer, line 57: Yeah, so if I understand correctly, we don't really need to escape in CDATA, and in rare cases, this escaping might annoy some people, but this still is just test failure reports.

result-helpers/src/commonTest/kotlin/lang/temper/result/junit/JUnitResultParserTest.kt
```diff
@@ -91,3 +107,15 @@ private const val SAMPLE_SUREFIRE_2 = """<?xml version="1.0" encoding="UTF-8"?>
...
 """
+
+private val controlChars = """<?xml version="1.0" encoding="UTF-8"?>
+<testsuites name="Mocha Tests" time="0.0030" tests="1" failures="1">
+<testsuite name="Root Suite" timestamp="2022-07-19T13:41:13" tests="1" time="0.0020" failures="1">
+<testcase name="a failing test" time="0.0000" classname="a failing test">
+<failure message="${"\u0001"}&#x2;&#3;" type="AssertionError"><![CDATA[
+Character entity escapes don't count in CDATA anyway, but meh?
+${"\u0001"}&#x2;&#3;
+]]></failure>
```
tjpalmer, line 118: I based this on a different one already in this file. Just added different forms of control char representation.

It's hard to add this a failing test in asserts funtest, because we typically check the error message, which changes across backends because of different escaping. See the PR description for examples.

frontend/src/commonMain/resources/std/testing/testing.temper.md
```diff
@@ -199,13 +199,22 @@ XML text node or attribute value.
...
+        let c = s[i];
+        let esc = when (c) {
           char'&' -> "&amp;";
           char'<' -> "&lt;";
           char'>' -> "&gt;";
           char"'" -> "&#39;";
           char'"' -> "&#34;";
-          else -> continue;
+          char'\n', char'\r', char'\t' -> continue;
+          else -> if (c < char' ' || c == 0xFFFE || c == 0xFFFF) {
```
mikesamuel, line 210: Could this just call isRiskyChar?

tjpalmer: > Could this just call isRiskyChar?

I could factor one here to make it look more like the Kotlin. And ideally, we could use the Temper implementation in Kotlin also, but fine-grained dependencies are more awkward. Also just updating them. Like, we'd have to tag some Maven package somewhere, such as for a tagged pre-release of Temper std in this case.

mikesamuel: Silly me.  I forgot two different languages are in play.
Being sick messes with your head.

tjpalmer: > Silly me. I forgot two different languages are in play. Being sick messes with your head.

No worries. Being sick is no fun.

I wonder if we should use our updateGeneratedCode things to build temper to java, copying unpublished temper std into some subproject of our repo for easier access to latest std and maybe other library changes. Could make sharing code easier maybe. (Out of scope for this pr, though.)

PR #334, opened 2026-02-16
frontend: autoesc optimizations are compatible with abstract collectors

mikesamuel:
Recent changes to contextual-autoescaping API conventions allow for contextual autoescapers to choose their own collectors.

This commit changes how contextual autoescaping blocks are optimized so that:

- the tag static methods are used to allocate the collector
- the collector type is interrogated to check if there's an appendSafe method.  If it's a simple StringBuilder, we don't bother distinguishing.

This commit adds some tweaks to support overload lookup in TypeShape.

### DIFF
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/function/OptimizeContextualAutoescapingBlocks.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/function/OptimizeContextualAutoescapingBlocks.kt
@@ -14,7 +14,6 @@ import lang.temper.env.InterpMode
 import lang.temper.frontend.InterpretationContext
 import lang.temper.frontend.structureBlock
 import lang.temper.frontend.syntax.isAssignment
-import lang.temper.interp.New
 import lang.temper.log.Position
 import lang.temper.name.ExportedName
 import lang.temper.name.ParsedName
@@ -30,11 +29,14 @@ import lang.temper.type.ExternalGet
 import lang.temper.type.MethodKind
 import lang.temper.type.MethodShape
 import lang.temper.type.TypeShape
+import lang.temper.type.Visibility
 import lang.temper.type2.DefinedNonNullType
 import lang.temper.type2.MkType2
+import lang.temper.type2.Signature2
 import lang.temper.type2.SuperTypeTree2
 import lang.temper.type2.Type2
 import lang.temper.type2.hackMapOldStyleToNew
+import lang.temper.type2.withType
 import lang.temper.value.Abort
 import lang.temper.value.BlockChildReference
 import lang.temper.value.BlockTree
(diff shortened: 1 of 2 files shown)


fundamentals/src/commonMain/kotlin/lang/temper/type/TypeDefinition.kt
```diff
@@ -314,8 +316,11 @@ sealed interface TypeShape : TypeDefinition {
 
     /**
      * All members with the given symbol, including those from the super type, in inheritance order.
+     *
+     * @param includeOverloads if false, the default, then only exact symbol matches are included, but
+     *   if true, then also included are normal methods that have matching `@overload("...")` metadata.
      */
-    fun membersMatching(symbol: Symbol): Iterable<MemberShape>
+    fun membersMatching(symbol: Symbol, includeOverloads: Boolean = false): Iterable<MemberShape>
```
tjpalmer, line 323: I suspect internal method calls haven't supported overloads, and I was considering looking into working on that. The improved helper here seems likely to lean toward that. Maybe I'll wait until this goes in before looking into that topic.

I've also considered adjusting to creating a overload class member that references all the overload participants. Building that could be a way to have it cached and handy. When building it might also be a good time to validate that no member already has the overload name.

frontend/src/commonMain/kotlin/lang/temper/frontend/function/OptimizeContextualAutoescapingBlocks.kt
```diff
@@ -92,19 +94,18 @@ import lang.temper.value.void
...
  * Our goal is to eliminate the first two and just operate directly on the last one.
  *
  *     ...
  *       // let accumulator = new SafeHtmlBuilder();
- *       let collector = new Collector<SafeHtml>();
+ *       let collector = SafeHtml.newCollector();
  *       ...
  *       // accumulator.append(...);
- *       collector.append(escapeHtmlPcdata(...));
+ *       collector.append(HtmlPcdataEscaper.instance.escape(...));
```
tjpalmer, line 108: Updated docs are nice. Tests could be nice, too.

Overall, I'm having trouble following all the rules for your current high speed dev. It might be nice to discuss what the overall effect is on the language when this is done and/or there's a breather.

fundamentals/src/commonMain/kotlin/lang/temper/type/TypeDefinition.kt
mikesamuel: Yeah.  The override metadata comes in surprisingly handy.

frontend/src/commonMain/kotlin/lang/temper/frontend/function/OptimizeContextualAutoescapingBlocks.kt
mikesamuel: Yeah.  I think I'm going to add tests to the secure-composition library project.

PR #340, opened 2026-02-18
std/regex: some minor fixes

mikesamuel:
regex compilation escapes `&` but JavaScript complains when it is escaped.

```js
$ node
Welcome to Node.js v25.2.1.
Type ".help" for more information.
> new RegExp("\\&")
/\&/
> new RegExp("\\&", "dgu")
Uncaught SyntaxError: Invalid regular expression: /\&/dgu: Invalid escape
> new RegExp("[&]", "dgu")
/[&]/dgu
```

There is a note in std/regex that Python requires escaping `&`.

This commit reworks the core.js function that creates the RegExp to unescape `&` specifically.

This also fixes a problem in regex.temper.md where code checks `<= 127` before looking into the list of safe ASCII codepoints, but the code that assembles the list does `< 127` leading to panics when compiling a regular expression that explicitly mentions U+7f.

### DIFF
--- a/frontend/src/commonMain/resources/std/regex/regex.temper.md
+++ b/frontend/src/commonMain/resources/std/regex/regex.temper.md
@@ -685,7 +685,7 @@ let needsNumericEscape = 1;
 let needsSimpleEscape = 2;
 let buildEscapeNeeds(): List<Int> {
   let escapeNeeds = new ListBuilder<Int>();
-  for (var code = 0; code < 0x7F; code += 1) {
+  for (var code = 0; code <= 0x7F; code += 1) {
     escapeNeeds.add(
       if (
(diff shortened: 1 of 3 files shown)


frontend/src/commonMain/resources/std/regex/regex.temper.md
```diff
@@ -685,7 +685,7 @@ let needsNumericEscape = 1;
 let needsSimpleEscape = 2;
 let buildEscapeNeeds(): List<Int> {
   let escapeNeeds = new ListBuilder<Int>();
-  for (var code = 0; code < 0x7F; code += 1) {
+  for (var code = 0; code <= 0x7F; code += 1) {
```
mikesamuel, line 688: See the use of <= 0x7f in pushCode at L509.

PR #342, opened 2026-02-18
Fix autoescaper optimization.

mikesamuel:
propagateOver expected an extra argument.

regexs couldn't be evaluated by macros because singletons and lookup tables were unavailable.  Added doPure calls to address that.

### DIFF
--- a/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustFunctionalTest.kt
+++ b/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustFunctionalTest.kt
@@ -23,11 +23,6 @@ class RustFunctionalTest : FunctionalTestRunner<RustBackend>(RustBackend.Factory
         super.algosHelloWorld()
     }
 
-    @Test // do not commit
-    override fun regexMatch() {
-        super.regexMatch()
-    }
-
     override fun runGeneratedCode(
         backend: RustBackend,
         modules: List<Module>,


frontend/src/commonMain/resources/std/regex/regex.temper.md
```diff
@@ -163,25 +163,25 @@ classes.
...
+export let Begin: Special = doPure { (): Special => class Begin extends Special {}; new Begin() };
+export let Dot: Special = doPure { (): Special => class Dot extends Special {}; new Dot() };
+export let End: Special = doPure { (): Special => class End extends Special {}; new End() };
 // TODO(tjp, regex): We can't easily support this at present across backends.
-// export let GraphemeCluster = do {
+// export let GraphemeCluster = doPure { (): Special =>
 //   class GraphemeCluster extends Special {}; new GraphemeCluster()
 // };
-export let WordBoundary: Special = do {
+export let WordBoundary: Special = doPure { (): Special =>
```
mikesamuel, line 173: Without these signatures, it infers the type as WordBoundary which actually breaks be-rust when we try to assign the WordBoundary to Special.

tjpalmer: > Without these signatures, it infers the type as WordBoundary which actually breaks be-rust when we try to assign the WordBoundary to Special.

I wonder if the outer type annotations could be removed now.

But it will be nice to get in context-based lambda param and return type inference.

frontend/src/commonMain/resources/std/regex/regex.temper.md
```diff
@@ -679,7 +679,7 @@ class RegexFormatter {
 }
 
 // Cache which chars you just but a blackslash in front of for escaping.
-let escapeNeeds = buildEscapeNeeds();
+let escapeNeeds = doPure { buildEscapeNeeds() };
```
tjpalmer, line 682: Yeah, maybe doPure is just an assertion later once we can infer/optimize better.

mikesamuel: Might `purely` be a better name then?

tjpalmer: > Might `purely` be a better name then?

I suspect `doPure` would still be clearer, but I could be wrong.

PR #344, opened 2026-02-19
cli: `temper repl -wr ...` now builds the work root so modules are av…

mikesamuel:
…ailable for import

Previously, `temper repl` had a work root which was used when running a target language's repl as via `temper repl -b js`.  But that didn't work for the regular `temper repl`.

This commit reworks *ModulesExternalToRepl* which *Repl* uses to allow importing `std` modules.  Now, it will build not just std, but if a work root is explicitly mentioned, it will build it and std.

I had to diagnose a problem which turned out to be premature closing of the *Build* object before *ModulesExternalToRepl* could advance modules from GenerateCode stage to Run.  For that it was convenient to rework interpreter continue conditions from `() -> Boolean` functions to a `fun interface` type.  This seems like a generally good thing, so I left it in place.

I also had trouble remembering import paths so I added to the `help` function something to list the available imports, shown below.

```
$ temper repl -w .
$ help()
help
  The help command takes a single argument that is a topic string or a function or type with a doc string. The topics are:
      help: this help topic
      describe: detailed information about an interactive result
      translate: translates an interpreted object into backend code
      loggingConfigNames: List of logging configuration names
      unresolved: Info about currently unresolved promises
      available-imports: List of modules available for import
      js/about: About the JavaScript backend (-b js)
      js/repl: Running a JavaScript REPL with libraries loaded

interactive#0: void
$ help("available-imports")
available-imports
  Here are the available modules.
  You can copy/paste the import statements.
  ╔══════════════════╦═════════╦══════════════════════════════════════════════════╗
  ║library           ║module   ║import                                            ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║secure-composition║core/    ║let {...} = import("secure-composition/core");    ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║examples/║let {...} = import("secure-composition/examples");║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║html/    ║let {...} = import("secure-composition/html");    ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║path/    ║let {...} = import("secure-composition/path");    ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║sql/     ║let {...} = import("secure-composition/sql");     ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║url/     ║let {...} = import("secure-composition/url");     ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║std               ║json/    ║let {...} = import("std/json");                   ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║net/     ║let {...} = import("std/net");                    ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║regex/   ║let {...} = import("std/regex");                  ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║temporal/║let {...} = import("std/temporal");               ║
  ╠══════════════════╬═════════╬══════════════════════════════════════════════════╣
  ║                  ║testing/ ║let {...} = import("std/testing");                ║
  ╚══════════════════╩═════════╩══════════════════════════════════════════════════╝
interactive#1: void
```

### DIFF
--- a/cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
+++ b/cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
@@ -5,9 +5,12 @@ import lang.temper.common.Console
 import lang.temper.common.assertStringsEqual
 import lang.temper.common.temperEscaper
 import lang.temper.env.InterpMode
+import lang.temper.fs.RealWritableFileSystem
 import lang.temper.fs.getTestDirectories
+import lang.temper.fs.runWithTemporaryDirectory
 import lang.temper.log.Debug
 import lang.temper.log.LogConfigurations
+import lang.temper.log.filePath
 import lang.temper.name.BuiltinName
 import lang.temper.name.TemperName
 import lang.temper.value.ActualValues
@@ -37,10 +40,15 @@ class ReplTest {
 
     @BeforeTest
     fun setupRepl() {
+        initializeRepl()
+    }
+
+    internal fun initializeRepl(workRootInfo: WorkRootInfo? = null) {
         _repl = Repl(
             writeToPendingConsole,
             executorService = executorService,
             directories = testDirectories(),
+            workRootInfo = workRootInfo,
         )
     }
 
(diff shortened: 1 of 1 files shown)


tjpalmer (commented):
So we can actually import the current work root libraries/modules in the repl. We've sure been needing this for a long time.

Any way to make a ReplTest for this without too much pain? Or is there a good enough test here that I missed?

mikesamuel:
Good idea.  I don't know if a ReplTest would make sense there.  ReplTest creates Repl instances not via the *Cli* so it wouldn't be testing the passing through of arguments.

But I can probably get a test that creates a temp directory, writes out a file, and imports it and then does a thing with it.

tjpalmer (approved):

PR #346, opened 2026-02-19
some improvements for better debugging via REPL

mikesamuel:
When testing in the REPL often there are some actions to take before you get to the exploratory work like `import`ing required libraries. This commit adds the `-e` flag so that the below runs an import before the REPL starts accepting user input.  Because it's separate, the `-e` commands do not show up in the user's history.

    temper repl -e 'let { ... } = import("std/temporal");'

I've been dogfooding and running into problems because Temper lacks a static check that `class` types implement all the `interface` methods they're supposed to.  That now happens in TypeDeclCheck.

For example:

    export interface I<T> { f(x: T): Void; }
    export class C extends I<String> {
      // but does not override f()
    }

> Type C must implement f from I.  Maybe add `public f(x: String): Void`

This also opportunistically fixes a type error (hit when typing `[^]` in the REPL without `rgx"..."` around it) in Typer.  It also sets REPL's very low continue condition quota to the same as ModuleAdvancer.

### DIFF
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeDeclChecker.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeDeclChecker.kt
@@ -1,0 +1,140 @@
+package lang.temper.frontend.generate
+
+import lang.temper.common.Log
+import lang.temper.frontend.Module
+import lang.temper.log.LogSink
+import lang.temper.log.MessageTemplate
+import lang.temper.log.unknownPos
+import lang.temper.name.Symbol
+import lang.temper.type.Abstractness
+import lang.temper.type.MethodKind
+import lang.temper.type.MethodShape
+import lang.temper.type.TypeShape
+import lang.temper.type.Visibility
+import lang.temper.type.WellKnownTypes
+import lang.temper.type2.DefinedNonNullType
+import lang.temper.type2.MkType2
+import lang.temper.type2.Signature2
+import lang.temper.type2.SuperTypeTree2
+import lang.temper.type2.mapType
+import lang.temper.value.connectedSymbol
+import lang.temper.value.parameterNameSymbols
+
+internal class TypeDeclChecker(val module: Module, val logSink: LogSink) {
+    private val abstractMethodCache = mutableMapOf<TypeShape, List<MethodDescriptor>>()
+
+    fun checkDeclaredTypeShapes() {
+        for (typeShape in module.declaredTypeShapes) {
+            when (typeShape.abstractness) {
+                Abstractness.Abstract -> {}
+                Abstractness.Concrete -> checkAllMethodsOverridden(typeShape)
+            }
+        }
+    }
+
+    private fun checkAllMethodsOverridden(typeShape: TypeShape) {
+        val superTypeShapes = mutableSetOf<TypeShape>()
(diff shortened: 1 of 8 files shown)


frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeDeclChecker.kt
```diff
@@ -0,0 +1,140 @@
...
+        val isProcessingImplicits = module.isEffectivelyImplicits
+        val isStd = module.isEffectivelyStd
+        val allAbstractMethodDescriptors = mutableListOf<MethodDescriptor>()
+        for (strictSuperTypeShape in superTypeShapes) {
+            val abstractMethods = abstractMethodCache.getOrPut(strictSuperTypeShape) {
+                strictSuperTypeShape.methods.mapNotNull { m ->
+                    if (isProcessingImplicits && connectedSymbol in m.metadata) {
+                        null // Some Implicits methods have no body because they must connect.
+                    } else if (isStd && m.visibility == Visibility.Private && connectedSymbol in m.metadata) {
+                        null // std has some required connections too which are sneakily hidden away
```
mikesamuel, line 56: Regex.compileFound

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeDeclChecker.kt
```diff
@@ -0,0 +1,140 @@
...
+                val type = MkType2(typeShape).actuals(
+                    typeShape.formals.map { MkType2(it).get() },
+                ).get() as DefinedNonNullType
+                val superTypeInContext = SuperTypeTree2.of(type)[example.enclosingType].firstOrNull()
+                var sig = example.descriptor
+                    ?: Signature2(WellKnownTypes.voidType2, false, listOf())
+                if (sig.hasThisFormal) { sig = sig.copy(requiredInputTypes = sig.requiredInputTypes.drop(1)) }
+                if (superTypeInContext != null) {
+                    val bindings = (example.enclosingType.formals zip superTypeInContext.bindings).associate { it }
+                    sig = sig.mapType(bindings)
```
mikesamuel, line 96: This signature mapping lets us know in the test case that, since C extends `I<T>` that `f`'s signature in `C` would be `f(x: String): Void`.

It occurs as I write this that this might be a good place to do something similar to check that overrides that are present are compatible, but for that, we'd need to not skip interface types, and collect all above so we could handle cases like:

    interface I { f(): Int32; }
    interface J { f(x: Int32): Void; }

     class C extends I & J { public f(): Int32 { 0 } } // compatible with I but not J

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -3712,6 +3712,23 @@ class GenerateCodeStageTest {
...
+            |export interface I<T> { f(x: T): Void; }
+            |export class C extends I<String> {
+            |  // but does not override f()
+            |}
+        """.trimMargin(),
+        want = """
+            |{
+            |  run: "void: Void",
+            |  errors: ["Type C must implement f from I.  Maybe add `public f(x: String): Void`!"]
+            |}
```
tjpalmer, line 3729: Happy to see this checked!

PR #347, opened 2026-02-19
Support empty regex code sets

tjpalmer:
- Fix temperlang/temper-regex-parser#26
- Support both `/[^]/` (match anything, including newlines) and `/[]/` (match nothing)

### DIFF
--- a/functional-test-suite/src/commonMain/resources/regex/match/match.temper.md
+++ b/functional-test-suite/src/commonMain/resources/regex/match/match.temper.md
@@ -229,6 +229,14 @@ And this should match nothing.
 
     console.log(/[]+/.find(simplerLines).full.value orelse "-");
 
+But this should match a different kind of nothing, the empty string.
+
+```temper inert
+    // Ignore this case, because it fails on Lua.
+    // If enabled, we need a blank line added to the next log block.
+    console.log(/[]*/.find(simplerLines).full.value);
+```
+
 ```log
 It was the best of times.
 It was the worst of times.


be-csharp/src/commonMain/resources/lang/temper/be/csharp/std/Regex/RegexSupport.cs
```diff
@@ -40,7 +40,7 @@ internal static IRegexNode AdjustCodeSet(
             RegexRefs regexRefs
         )
         {
-            if (!codeSet.Negated)
+            if (!codeSet.Negated || codeSet.Items.Count == 0)
```
tjpalmer, line 43: Customizing `/[^]/` is elsewhere, so don't go crazy on it here.

frontend/src/commonMain/resources/std/regex/regex.temper.md
```diff
@@ -567,14 +567,26 @@ class RegexFormatter {
...
-          pushCodeSetItem(adjusted.items[i]);
+        if (adjusted.items.isEmpty) {
+          // Many regex engines don't like empty code sets.
+          if (adjusted.negated) {
+            // Match anything.
+            out.append(raw"[\s\S]");
+          } else {
+            // Match nothing.
+            out.append("(?:$.)");
+          }
```
tjpalmer, line 578: Here's where we customize empty code sets. Both of these seem to work across all backends right now. Interp doesn't use this code path, though.

functional-test-suite/src/commonMain/resources/regex/match/match.temper.md
```diff
@@ -215,6 +215,26 @@ negated basic true: true
...
+
+And this should match nothing.
+
+    console.log(/[]+/.find(simplerLines).full.value orelse "-");
+
+```log
+It was the best of times.
+It was the worst of times.
+-
+```
```
tjpalmer, line 244: Match everything and match nothing tests.

regex/src/commonMain/kotlin/lang/temper/regex/RegexFormat.kt
```diff
@@ -287,12 +287,24 @@ private fun PatternFormatContext.formatUnwrapped(codeRange: CodeRange) {
...
+                    if (adjusted.negated) {
+                        out.append('^')
+                    }
+                    adjusted.items.forEach { formatCodeSetItem(it) }
+                    out.append(']')
+                }
             }
-            adjusted.items.forEach { formatCodeSetItem(it) }
-            out.append(']')
         }
```
tjpalmer, line 308: This is where the logic for interp is, so duplicated here.

regex/src/commonTest/kotlin/lang/temper/regex/RegexMatchTest.kt
```diff
@@ -78,6 +78,7 @@ class RegexMatchTest {
             "moreBoundaryUgly" to
                 RegexCheck(true, RegexMatch("""(?!$word)ø${wordBoundary}8""", "ø8")),
             "digitsNonAsciiUgly" to RegexCheck(false, RegexMatch("""[0-9]+""", "\u0662")),
+            "matchAll" to RegexCheck(true, RegexMatch("""[\s\S]+""", "a\nb")),
```
tjpalmer, line 81: I didn't realize where the problem was in dotnet for a while, so I was testing this here. Decided to leave it.

regex/src/commonTest/kotlin/lang/temper/regex/RegexMatchTest.kt
```diff
@@ -202,7 +203,7 @@ internal fun checkMatchDotnet(checks: Iterable<RegexCheck>) {
             List.map
                 (fun pair ->
                     try
-                        Nullable<bool>(Regex.IsMatch(pair.matchText, pair.pattern))
+                        Nullable<bool>(Regex.IsMatch(pair.matchText, "^" + pair.pattern + "$"))
```
tjpalmer, line 206: Realized I wasn't testing full regexes, but I think it was implied to do so. All the current tests pass either way, but I could construct cases where it made a difference.

frontend/src/commonMain/resources/std/regex/regex.temper.md
mikesamuel: Does the difference between `$` and `\z` matter here?

I guess `.` doesn't match anything allowed by `.` on perlre.

functional-test-suite/src/commonMain/resources/regex/match/match.temper.md
```diff
@@ -215,6 +215,26 @@ negated basic true: true
 negated basic false: false
 ```
 
+### Empty sets
+
+This should match everything, including newlines.
+
+    let simplerLines = """
+      "It was the best of times.
+      "It was the worst of times.
```
mikesamuel, line 224: It was the best of lines.
It was the worst of lines.

functional-test-suite/src/commonMain/resources/regex/match/match.temper.md
```diff
@@ -215,6 +215,26 @@ negated basic true: true
...
+    ;
+    console.log(/[^]+/.find(simplerLines).full.value);
+
+And this should match nothing.
+
+    console.log(/[]+/.find(simplerLines).full.value orelse "-");
```
mikesamuel, line 230: I guess the semantics of []* are that it still matches the empty string.

frontend/src/commonMain/resources/std/regex/regex.temper.md
tjpalmer: > Does the difference between `$` and `\z` matter here?

I think `$` only matches end of string on our current configurations. Part of the reason to test a multiline string. I think it's safe. Do you have a particular concern?

functional-test-suite/src/commonMain/resources/regex/match/match.temper.md
tjpalmer: > I guess the semantics of []* are that it still matches the empty string.

Just tested it, and seems it works that way everywhere except Lua. 😭

I can add a commit with this case added but commented out for future reference, if you want. It seems edge case enough to me that we can defer it.

tjpalmer: I added a PR with the commented-out case.

functional-test-suite/src/commonMain/resources/regex/match/match.temper.md
```diff
@@ -215,6 +215,34 @@ negated basic true: true
...
+
+But this should match a different kind of nothing, the empty string.
+
+```temper inert
+    // Ignore this case, because it fails on Lua.
+    // If enabled, we need a blank line added to the next log block.
+    console.log(/[]*/.find(simplerLines).full.value);
+```
```
tjpalmer, line 238: Added because of [this comment](https://github.com/temperlang/temper/pull/347#discussion_r2829935549).

PR #348, opened 2026-02-20
StringExprMacro coalesces string chunks together.

mikesamuel:
We split string chunks with escape sequences up.  To minimize the number of appendSafe calls, now StringExprMacro coalesces adjacent trusted string writes together.

### DIFF
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
@@ -3451,7 +3451,6 @@ class DefineStageTest {
             |{
             |  define: {
             |    body: ```
-            |
             |      @stay @imported(\(`test//html/`.html)) let html__0;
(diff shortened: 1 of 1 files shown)


frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
```diff
@@ -3329,18 +3329,15 @@ class DefineStageTest {
             |            do_bind_appendSafe(accumulator#0)("Zero: ");
             |## Unsafe interpolations become regular appends.
             |            do_bind_append(accumulator#0)(0);
-            |            do_bind_appendSafe(accumulator#0)("\n");
-            |            void;
-            |            do_bind_appendSafe(accumulator#0)("One: ");
+            |            do_bind_appendSafe(accumulator#0)("\nOne: ");
```
tjpalmer, line 3332: So looks like each line chunk was separate before.

PR #349, opened 2026-02-20
flatten unnecessary recursivity when farming a tree

mikesamuel:


fundamentals/src/commonMain/kotlin/lang/temper/value/TreeFarm.kt
```diff
@@ -425,37 +425,60 @@ private class SingleTreePlanting(
...
-                walk(i + 1, t.pos.rightEdge, right)
+                out.add(CompletedTree(t))
+                left = t.pos.rightEdge
                 t.pos.leftEdge
             }
+
             else -> {
-                val index = out.size
-                out.add(null) // Placeholder.  Will be replaced with a non-null value.
-                val fromTail = walk(i + 1, left, right)
```
mikesamuel, line 449: Instead of pre-recursing here, we instead just do a forward path, remember the lefts and then do a backwards pass if necessary.

tjpalmer (approved):
Is this just optimizing execution, as no test results are changed?

mikesamuel:
Nothing should change.  Just avoiding a stack overflow when there's an egregious syntax error that leads to a lot of garbage.

PR #350, opened 2026-02-21
Add checks that overrides are compatible with overridden signatures

mikesamuel:
This commit implements override checking, but it currently breaks because we have not been careful about overrides especially whether a method bubbles or not.  I'm making this a draft and will try to address some of those problems in follow-on commits.


tjpalmer:
Being continued in #475.

PR #354, opened 2026-02-23
cli: `--logformat json` logs to JSONL format

mikesamuel:
This adds a `--logformat` flag to the CLI.
`temper help build` now includes the below:

```
    --logformat [Text] -> Specifies how log messages are delivered.

            text: Human-readable text which includes extra detail
                  like ASCII-art source snippets for context.
            json: One JSON record per line à la jsonlines.org.

            { Value should be one of [text, json] }
```

And here's how it practically looks in use.

```
🐚$ temper test --logformat text -b js
Test failed (js): values in JavaScript - Runnable failed with falsy or undefined exception. Please throw an Error instead.
Tests passed: 40 of 41
Test failed
🐚$ temper test --logformat json -b js
{ "level": "error", "template": "TestFailed", "values": [ "js", "values in JavaScript", "Runnable failed with falsy or undefined exception. Please throw an Error instead." ] }
{ "level": "summary", "template": "TestSummaryAllRun", "values": [ 40, 41 ] }
{ "task": "Test", "ok": false }
🐚$
```

This commit plumbs it into the main build tasks: `build`, `test`, `serve`, and `run`.

It could be used in `temper repl -b ...` to deliver JSON output when building the workroot, but this commit does not bother to do that. It seems low value, and inconsistent with how the REPL would deliver results.  The REPL takes direct control over the console frequently.

----

This commit needed to change several things.

*ConsoleBackedContextualLogSink* is where entries are converted to strings.  To keep *ValueFormatter* behaviour the same, this commit reworks some logic around formatting so that the bulk is the same until it's time to slot formatted values into a template.

This commit also cleans up some build machinery especially in TestTally that directly output to the console.  Now, they go via the log sink allow test result summaries to be delivered the same way.

The decision as to which logging format to use is made in *LogAsJson*.

Especially in `--verbose` mode, users might see log messages about staging of *Implicits.temper* and `std/...` files. For a host of pragmatic reasons, those can kick off without any input from *Build.kt*.

I wanted *Cli* to be able to flip all logging easily, so I made the log formatting decision a global one.  *Cli* controls it by a block call.  `withLogAsJson(logAsJson=true) {...}`. In practice, there's only one caller for that, but its semantics is to bump an *AtomicInt* on entry and decrement on exit, so if it is reentrant, nothing bad will happen.

The new log format is called JSON.  Every log record is formatted as JSON, but the records are separated by newlines. So the actual format of stderr is [JSON lines](jsonlines.org).

> This page describes the JSON Lines text format, also called
> newline-delimited JSON. JSON Lines is a convenient format for
> storing structured data that may be processed one record at a
> time. It works well with unix-style text processing tools and shell
> pipelines. It's **a great format for log files**. It's also a flexible
> format for passing messages between cooperating processes.

----

This commit adds a logging level between *Log.Warn* and *Log.Error*: *Log.Summary*.  Even if we want to filter out all the *Info* level requests and the ignorable warnings, we might still want summaries like the count of tests that passed.

### DIFF
--- a/scripts/integration_test/checks.py
+++ b/scripts/integration_test/checks.py
@@ -2,6 +2,7 @@ import collections as c
 import contextlib as ctx
 import glob
 import inspect
+import json
 import os.path
 import pathlib as p
 import shutil as sh
@@ -73,6 +74,8 @@ class Checks:
                 # Including multiple together, since we had a regression with that.
                 "test-backend=java",
                 "test-backend=csharp,js",
+                # Test JSON formatting of log output
+                "test-backend=py-logformat=json",
                 "watch",
                 # Slow and currently broken, so exclude by default.
                 # "mypy_produces_binaries",
@@ -218,10 +221,12 @@ class Checks:
         )
 
     @report_check
-    def check_test(self, backend: str) -> None:
+    def check_test(self, backend: str, logformat: str | None = None) -> None:
         with self.tmp_copy("cli/src/test/resources/testing/multi") as tmp:
             backends = backend.split(",")
             backend_args = [arg for b in backends for arg in ("-b", b)]
+            if logformat is not None:
+                backend_args[:0] = ['--logformat', logformat]
             process = self.run_temper(
                 args=["test", *backend_args, *self.verbosity()],
                 cwd=tmp,
@@ -235,7 +240,40 @@ class Checks:
             print(f"Temper test {backend}\nstderr:\n{stderr}\nstdout:\n{stdout}")
             if not self.verbose:
                 n = 3 * len(backends)
-                self.check("test", lambda: f"Tests passed: {n} of {n}" in stderr)
+                if logformat == 'json':
+                    # Need to check the output log message and also
+                    # that each line of stderr is its own JSON record
+                    records = self._unpack_logformat_json_stderr(stderr)
+                    all_records_ok = all(
+                        isinstance(x, dict) or
+                        not x.startswith('INVALID_JSON: ')
+                        for x in records
+                    )
+                    wanted_record = ({ "level": "summary",
+                                       "template": "TestSummaryAllRun",
+                                       "values": [ n, n ] })
+                    wanted_record_present = wanted_record in records
+                    def condition() -> bool:
+                        return all_records_ok and wanted_record_present
+                    def on_failure():
+                        print('Records parsed as JSON')
+                        for record in records:
+                            print(f'- {record !r}')
+                        if not wanted_record_present:
+                            print(f'Expected {wanted_record !r}')
+                else:
+                    condition = lambda: f"Tests passed: {n} of {n}" in stderr
+                    on_failure = None
+                self.check("test", condition, on_failure = on_failure)
+
+    def _unpack_logformat_json_stderr(self, stderr: str) -> list[str | dict[str, any]]:
+        def unpack(x: str) -> str:
+            """Unpack a JSON record from a line"""
+            try:
+                return t.cast(dict[str, any], json.loads(x))
+            except json.decoder.JSONDecodeError:
+                return 'INVALID JSON: {x}'
+        return [unpack(line) for line in stderr.splitlines()]
 
     @report_check
     def check_watch(self) -> None:


cli/src/main/kotlin/lang/temper/cli/Cli.kt
```diff
@@ -320,17 +340,21 @@ abstract class Main {
                     """.trimMargin(),
                 ) {
                     val wr = workRootOpt(this)
+                    val ignoreFile = ignoreFileOpt(this)
```
mikesamuel, line 343: idk if the order of creation here affects the order of presentation in `temper help` but I figure I'd group wr and ignoreFile together since the latter closely affects the former.

cli/src/main/kotlin/lang/temper/cli/Cli.kt
```diff
@@ -799,16 +836,33 @@ abstract class Main {
         }
 
         val executorService = makeExecutorServiceForBuild()
-        return try {
-            process(console, executorService).also { ok ->
-                if (!ok) {
+        return withLogAsJson(logAsJson = logFormat == LogFormat.Json) {
```
mikesamuel, line 839: This is where Cli changes the behaviour of logging machinery.

This adds a level of nesting, so viewing with `-w` as a diff option can help.  Manor changes to the body include the "$label failed" final logging.

cli/src/main/kotlin/lang/temper/cli/Cli.kt
```diff
@@ -799,16 +836,33 @@ abstract class Main {
...
+                    if (finalSummary != null) {
+                        if (!ok) {
+                            console.error(finalSummary)
+                        } else {
+                            console.log(finalSummary, level = Log.Summary)
+                        }
+                    }
                 }
+            } finally {
+                console.textOutput.flush()
```
mikesamuel, line 863: We want the shutdown to happen within the scope of the set of the logging global bit.

format/src/commonMain/kotlin/lang/temper/format/ConsoleBackedContextualLogSink.kt
mikesamuel, line 1: This is where the bulk of the changes to the logging happen.

format/src/commonMain/kotlin/lang/temper/format/ConsoleBackedContextualLogSink.kt
```diff
@@ -78,38 +83,83 @@ open class ConsoleBackedContextualLogSink(
...
+            val formatSink = TextOutputFormatSink(
+                textOutput = textOutput,
+                style = level.style,
+                customValueFormatter = CustomValueFormatter.chain(
+                    customValueFormatter,
+                    logSinkValueFormatter,
+                ),
+            )
 
+            if (logAsJson) {
```
mikesamuel, line 119: See new file in same directory.

format/src/jvmMain/kotlin/lang/temper/format/LogAsJson.jvm.kt
```diff
@@ -0,0 +1,19 @@
+package lang.temper.format
+import java.util.concurrent.atomic.AtomicInteger
+
+private val globalLogAsJsonBit = AtomicInteger(0)
```
mikesamuel, line 4: This is mutable global state.

I don't love it, but I think it's better to have consistent logging for Implicits and std/ without entangling building those with Build.kt which would have to otherwise pass preference flags around.

format/src/jvmMain/kotlin/lang/temper/format/LogAsJson.jvm.kt
```diff
@@ -0,0 +1,19 @@
...
+
+actual val logAsJson: Boolean get() = globalLogAsJsonBit.get() > 0
+
+actual fun <T> withLogAsJson(logAsJson: Boolean, body: () -> T): T {
```
mikesamuel, line 8: This is meant to allow running builds in a scoped block where logging is JSON if the argument is true.

Using an AtomicInteger instead of an AtomicBoolean allows side-stepping any semantic confusion if multiple blocks end up calling this.  They shouldn't, but it's not expensive to just have a clear semantics for this block-taking function.

langserver/src/main/kotlin/lang/temper/langserver/UpdatePublisher.kt
```diff
@@ -130,7 +130,7 @@ fun LogEntry.toDiagnostic(filePositions: FilePositions): Diagnostic? {
         range = entry.pos.toRange(filePositions = filePositions)
         severity = when (entry.level) {
             Log.Fine -> DiagnosticSeverity.Hint
-            Log.Info -> DiagnosticSeverity.Information
+            Log.Info, Log.Summary -> DiagnosticSeverity.Information
```
mikesamuel, line 133: Log.Summary gets formatted as non-error content, but is higher priority than ignorable warnings.

mikesamuel (commented):
Best to view diffs with whitespace ignored

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -767,3 +769,5 @@ class PyBackendTest {
 }
 
 private val dunderInitFiles = listOf("$DUNDER_INIT.py", "$DUNDER_INIT.py.map")
+
+private const val LONG_TEST_TIMEOUT_SECONDS = 60L
```
tjpalmer, line 773: Good idea.

scripts/integration_test/checks.py
```diff
@@ -73,6 +74,8 @@ def check_all(self, checks: t.Optional[t.List[str]] = None) -> bool:
                 # Including multiple together, since we had a regression with that.
                 "test-backend=java",
                 "test-backend=csharp,js",
+                # Test JSON formatting of log output
+                "test-backend=py-logformat=json",
```
mikesamuel, line 78: Plumbing in an integration test for --logformat=json

scripts/integration_test/checks.py
```diff
@@ -235,7 +240,40 @@ def check_test(self, backend: str) -> None:
             print(f"Temper test {backend}\nstderr:\n{stderr}\nstdout:\n{stdout}")
             if not self.verbose:
                 n = 3 * len(backends)
-                self.check("test", lambda: f"Tests passed: {n} of {n}" in stderr)
+                if logformat == 'json':
+                    # Need to check the output log message and also
+                    # that each line of stderr is its own JSON record
+                    records = self._unpack_logformat_json_stderr(stderr)
```
mikesamuel, line 246: Parsing each line as JSON will flush out problems.

scripts/integration_test/checks.py
```diff
@@ -235,7 +240,40 @@ def check_test(self, backend: str) -> None:
...
+                    all_records_ok = all(
+                        isinstance(x, dict) or
+                        not x.startswith('INVALID_JSON: ')
+                        for x in records
+                    )
+                    wanted_record = ({ "level": "summary",
+                                       "template": "TestSummaryAllRun",
+                                       "values": [ n, n ] })
```
mikesamuel, line 254: Here's the exact test summary record

tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/TestTally.kt
```diff
@@ -47,22 +52,52 @@ data class TestTally(
...
+    ),
+    TestSummaryAllRun(
+        Log.Summary,
+        "Tests passed: %d of %d",
+    ),
+    TestSummaryNotAllRun(
+        Log.Summary,
+        "Tests passed: %d of %d (%d not run)",
+    ),
+}
```
tjpalmer, line 94: I presume the summary output is the same as before, given what these look like.

mikesamuel: Yep.  This commit should not change anything for the text form.

PR #355, opened 2026-02-24
Fix overload finding and some missing error reports

mikesamuel:
The `@overload` checking was done in the same loop as the pre-`@overload` checking which led to some bad interactions when a sub-type overrides some overloads but not others from a super-type.

I've used a pattern where I declare many overloads in an interface and most of them have default method bodies that delegate to one and in a subtype I overload just that one.  That runs into this problem.

This commit also fixes some missing errors:

    let x: NonNullType = null;

In this case the `null` is inferred as having a type of *NonNullType*. Typer needs to be free to do a number of strategies to find the right type for `null` values, so this is not unreasonable.  TypeChecker is responsible for flagging errors, so this is fixed there.

    let one = 1i6;

That probably meant 1i6**4** but that failed silently too.  This commit tweaks *UnpackValue* to return more information on a failed unpacking and tweaks *Lift* to log that info.


frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -2767,12 +2767,17 @@ internal class Typer(
...
                 }
                 processMembers(thisVariant, typeShape)
-                if (!anyMatchedSymbol) {
-                    // No actual member had the right symbol, so check for overloads.
-                    // Checking for overloads only on actual member symbol fails should make the common case faster.
-                    // TODO Permit overloads with an actual matching member also?
-                    // TODO If not, validate against that somewhere else.
-                    processMembers(thisVariant, typeShape) { matchesOverload(memberSymbol, it) }
+            }
+            if (!anyMatchedSymbol) {
```
mikesamuel, line 2771: As written before, if we processed overload members in one typeShape then now anyMatchedSymbol is set and that blocks considering overloads in other typeshapes.

frontend/src/commonTest/kotlin/lang/temper/frontend/SyntaxMacroStageTest.kt
```diff
@@ -2255,4 +2255,25 @@ class SyntaxMacroStageTest {
             |}
         """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
     )
+
+    @Test
+    fun malformedStringLiteralErrors() = assertModuleAtStage(
+        stage = Stage.SyntaxMacro,
+        input = """
+            |export let oneTwoThree = 123i6;
```
mikesamuel, line 2263: i6 is missing the '4'

mikesamuel:
This is ready for review now.

docs/for-users/temper-docs/docs/reference/types.md
```diff
@@ -794,7 +794,7 @@ Temper also does not recognize all C's number suffixes.
 
 ```temper
 1F
-// ❌ Expected subtype of Top, but got Invalid!
+// ❌ Malformed number!
 ```
```
tjpalmer, line 798: Definitely better error message.

frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2150,6 +2150,69 @@ class TypeStageTest {
...
+            |
+            |export class C extends I {
+            |  @overload("foo")
+            |  public fooInt32(x: Int32): Void { fooString("Int32 $x"); }
+            |
+            |  // Does not overload foolean
+            |
+            |  // Implements fooString but does not redeclare metadata
+            |  public fooString(x: String): Void {
+            |    ;
```
tjpalmer, line 2188: The `;` isn't needed here, is it? In some rare cases, I do still say `void`, but I wouldn't expect that to be needed here, either. Maybe the semi is just for emphasis here?

PR #356, opened 2026-02-24
be-js: more readable strings

mikesamuel:
When testing code that generates HTML, it's hard to deal with escape sequences like `\u003c` instead of just `<` inside quotes.  We had been using the JSON escaper which is HTML-embedding-safe by default.

This just writes a custom JS encoder instead of reusing the JSON one and adds some tests for it.

### DIFF
--- a/build-user-docs/src/test/kotlin/lang/temper/docbuild/GrammarProductionExtractorTest.kt
+++ b/build-user-docs/src/test/kotlin/lang/temper/docbuild/GrammarProductionExtractorTest.kt
@@ -30,9 +30,7 @@ class GrammarProductionExtractorTest {
                     """
                     |## Syntax for *JsonBoolean*
                     |
-                    |![JsonBoolean &#58;&#61; &#34;false&#34; &#124; &#34;true&#34;]${
-                        ""
-                    }[snippet/syntax/JsonBoolean.svg]
+                    |![JsonBoolean &#58;&#61; &#34;false&#34; &#124; &#34;true&#34;][snippet/syntax/JsonBoolean.svg]
                     |
                     |Truth values are represented using the keywords `false` and `true`.
                     """.trimMargin(),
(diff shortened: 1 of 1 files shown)


be-js/src/commonMain/kotlin/lang/temper/be/js/StringTokenText.kt
```diff
@@ -1,23 +1,85 @@
...
+            val c = value[i]
+            i += 1
+            if (c.isJsLineTerminatorChar) {
+                lineBreakCount++
+                if (i < n && c == '\r' && value[i] == '\n') {
+                    i += 1 // Count CR LF as one terminator
+                }
+            }
+        }
+        if (lineBreakCount >= BACKQUOTED_LF_COUNT_THRESHOLD) {
```
mikesamuel: I don't quite know what this was for, but it was the prior behaviour.  If you've got enough line breaks and you can just slap backquotes around it, do so.

be-js/src/commonTest/kotlin/lang/temper/be/js/StringTokenTextTest.kt
```diff
@@ -0,0 +1,37 @@
...
+    @Test
+    fun allAscii() {
+        val ascii = buildString {
+            for (i in 0..127) {
+                appendCodePoint(i)
+            }
+        }
+        assertStringsEqual(
+            listOf(
+                """\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e\x0f""",
```
mikesamuel, line 16: 16 per row.

be-js/src/package-lock.json
mikesamuel, line 1: There are some npm audit reports.  Nothing that looks important.  They often flag on dev dependencies.

be-js/src/commonMain/kotlin/lang/temper/be/js/StringTokenText.kt
```diff
@@ -1,29 +1,61 @@
...
 internal fun stringTokenText(value: String): String {
-    var lineBreakCount = 0
-    var hasTemplateStringSpecial = false
-    loop@
+    var lowUsedMask = 0L
     for (c in value) {
-        if (c.isJsLineTerminatorChar) {
-            lineBreakCount += 1 // modulo CRLF
-        } else if (c == '\\' || c == '$' || c == '`') {
-            hasTemplateStringSpecial = true
```
mikesamuel, line 13: On second thought, got rid of the backticked multiline string case.  Apparently no existing tests really care about it, so just simplified it out.

be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
```diff
@@ -674,7 +674,52 @@ class JsBackendTest {
             |        "list.js": {
             |          "_name": "list.js",
             |          "_type": "txt",
-            |          "content": "/** @type {Array\u003cnumber\u003e} */\nconst return_0 = Object.freeze([3, 4]);\nexport default return_0;\n"
+            |          "content": ```
+            |            /** @type {Array<number>} */
+            |            const return_0 = Object.freeze([3, 4]);
+            |            export default return_0;
```
tjpalmer, line 680: Splitting the lines also is nice.

Issue #271, opened 2026-01-03
Rust builders missing type params


tjpalmer:
For this temper:

```ts
    export class Range<Item>(
      public start: Int,
      public end: Int,
    ) {}
```

I get this rust error:

```rs
error[E0412]: cannot find type `Item` in this scope
   --> /home/tom/projects/riot/engine/temper.out/rust/rio-engine/src/mod.rs:960:33
    |
960 |     pub fn build(self) -> Range<Item> {
    |                                 ^^^^ not found in this scope
    |
help: you might be missing a type parameter
    |
959 | impl<Item> RangeBuilder {
    |     ++++++
```

PR #357, opened 2026-02-25
be-rust: Fill out interfaces; improve generics

### LINKED ISSUE #345: Rust sealed enums need type parameters
We fail to put the generics on `AEnum` in the generated code here:

```rs
$ sealed interface A<T> {}; class B<T> extends A<T> {}; class C<T> extends A<T> {}
interactive#0: C__4
$ translate(0, "rust")
Translated rust for interactive#0
  interactive/
    src/
      i0000/
        mod.rs: text/rust
...
          enum AEnum {
              B(B), C(C)
          }
          trait ATrait<T: Clone + std::marker::Send + std::marker::Sync + 'static>: temper_core::AsAnyValue + temper_core::AnyValueTrait + std::marker::Send + std::marker::Sync {
              fn as_enum(& self) -> AEnum;
              fn clone_boxed(& self) -> A<T>;
          }
...
          struct BStruct<T: Clone + std::marker::Send + std::marker::Sync + 'static> {
              phantom_T: std::marker::PhantomData<T>
          }
...
          struct CStruct<T: Clone + std::marker::Send + std::marker::Sync + 'static> {
              phantom_T: std::marker::PhantomData<T>
          }
...
```

It should say something about like this maybe:

```rs
enum AEnum<T: Clone + std::marker::Send + std::marker::Sync + 'static> {
    B(B<T>), C(C<T>)
}
```

tjpalmer:
- Fix many of the bigger issues needed for our current demo work
- Some issues remain, especially dealing with type parameter bindings not being delivered in tmpl (haven't checked frontend yet) for constructor calls
- Fix #345

### DIFF
--- a/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
+++ b/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
@@ -67,7 +67,7 @@ class RustBackendTest {
             |      src: {
             |        lib.rs: {
             |          content: ```
-            |            #![allow(warnings)]
+            |            #![allow(dependency_on_unit_never_type_fallback, warnings)]
             |            pub mod bar;
             |            mod r#mod;
             |            pub use r#mod::*;
@@ -86,7 +86,7 @@ class RustBackendTest {
             |        main.rs: "__DO_NOT_CARE__",
             |        mod.rs: {
             |          content: ```
-            |            #![allow(warnings)]
+            |            #![allow(dependency_on_unit_never_type_fallback, warnings)]
             |            use temper_core::AnyValueTrait;
             |            use temper_core::AsAnyValue;
             |            use temper_core::Pair;
@@ -124,7 +124,7 @@ class RustBackendTest {
             |        bar: {
             |          mod.rs: {
             |            content: ```
-            |              #![allow(warnings)]
+            |              #![allow(dependency_on_unit_never_type_fallback, warnings)]
             |              use temper_core::AnyValueTrait;
             |              use temper_core::AsAnyValue;
             |              use temper_core::Pair;
@@ -1067,7 +1067,7 @@ class RustBackendTest {
             |      src: {
             |        lib.rs: {
             |          content: ```
-            |            #![allow(warnings)]
+            |            #![allow(dependency_on_unit_never_type_fallback, warnings)]
             |            pub mod bar;
             |            pub mod bob;
             |            mod support;
@@ -1086,7 +1086,7 @@ class RustBackendTest {
             |        bar: {
             |          mod.rs: {
             |            content: ```
-            |              #![allow(warnings)]
+            |              #![allow(dependency_on_unit_never_type_fallback, warnings)]
             |              use temper_core::AnyValueTrait;
             |              use temper_core::AsAnyValue;
             |              use temper_core::Pair;
@@ -1122,7 +1122,7 @@ class RustBackendTest {
             |            "beth": {
             |              "mod.rs": {
             |                content: ```
-            |                  #![allow(warnings)]
+            |                  #![allow(dependency_on_unit_never_type_fallback, warnings)]
             |                  use temper_core::AnyValueTrait;
             |                  use temper_core::AsAnyValue;
             |                  use temper_core::Pair;
@@ -2424,7 +2424,7 @@ private fun assertGenerateWanted(modules: List<ModuleInfo>) {
             |                    "mod.rs": {
             |                        "content":
             |```
-            |#![allow(warnings)]
+            |#![allow(dependency_on_unit_never_type_fallback, warnings)]
             |use temper_core::AnyValueTrait;
             |use temper_core::AsAnyValue;
             |use temper_core::Pair;


be-rust/src/commonMain/kotlin/lang/temper/be/rust/rust.out-grammar
```diff
@@ -50,11 +50,16 @@ BreakExpr.value.default = `null`;
 
 Call ::=
       ((needsParens%`Boolean` == true => "(" & callee%Expr & ")") || callee)
-    & "(" & args%Expr*"," & ")"
+    & "("
+    & args%Expr*","
+    // Optional `where` here is purely a hack for a specific macro form we have.
+    & where%Where?
+    & ")"
```
tjpalmer, line 57: This is very hacky just to support the macro I've expanded for delivering type param bounds.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -54,6 +54,24 @@ fun makeSrcFilePath(relDir: List<FilePathSegment>): FilePath {
...
+            translatedFormal.pos,
+            id = translatedFormal.id,
+            bounds = translatedFormal.bounds.slice(0..<ownBoundsCount).map { it.deepCopy() },
+        )
+    }
+    when {
+        whereItems.isEmpty() -> null
+        else -> Rust.Where(pos, whereItems)
+    }
+}
```
tjpalmer, line 82: There'll be some examples of using the hacky `where` in backend test.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -649,8 +669,14 @@ internal fun Rust.Type.isUnit() = this is Rust.Id && this.outName.outputNameText
...
-internal fun TypeDefinition.allInterfaces(): Sequence<Pair<TypeDefinition, Type2>> = sequence {
-    // SuperTypeTree requires a NominalType to start with, and I didn't find that in TmpL.TypeDeclaration.
+internal fun TypeDefinition.allInterfaces(
+    allowStart: Boolean = false,
+): Sequence<Pair<TypeDefinition, Type2>> = sequence {
+    val thisType = this@allInterfaces
+    if (allowStart && thisType is TypeShape && thisType.abstractness == Abstractness.Abstract) {
+        // For interfaces, we need to implement in the trait for its own wrapper, so provide that here.
+        yield(thisType to MkType2(thisType).position(pos).get())
+    }
```
tjpalmer, line 688: We need to implement traits for our translated class types, but I'm now also doing that for interface types. And in that case, we also need to implement the trait for its own trait wrapper. Again, examples later.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -644,20 +648,6 @@ class RustTranslator(
...
-            when (member) {
-                is TmpL.NormalMethod -> member.name.name
-                // We sometimes get names like "getwhatever" rather than "get.whatever", so be explicit.
-                is TmpL.GetterOrSetter -> when (member) {
-                    is TmpL.Getter -> "get"
-                    is TmpL.Setter -> "set"
-                }.let { BuiltinName("$it.${member.dotName}") } // changed to "nym`...`" below
-                else -> error("unexpected")
-            }.displayName
-        }
```
tjpalmer, line 660: Moved elsewhere.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -676,21 +666,46 @@ class RustTranslator(
             // Because type defs are always top level, we don't need a stack of indicators.
             insideMutableType = false
         }
-        // Also all trait impls, remembering for AnyValue macro use.
+        // Implement traits, including AnyValue.
+        val supTypes = implTraits(pos, decl, typeRef, generics)
```
tjpalmer, line 670: The new `implTraits` helper is now used for both classes and interfaces.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -708,65 +723,112 @@ class RustTranslator(
...
+                                        args = callClone.deepCopy().let { clone ->
+                                            when {
+                                                subShape.isInterface() ->
+                                                    clone.box(wanted = subShape, translator = this@RustTranslator)
+
+                                                else -> clone
+                                            }
+                                        }.let { listOf(it) },
+                                    )
+                                },
```
tjpalmer, line 747: I have a generalized area elsewhere for temper-defined methods. This is for inventions purely on the rust side that I'd previously made called `as_enum` above and `clone_boxed` below. Special logic for classes, but interface wrappers just forward to their trait definitions in all cases.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -708,65 +723,112 @@ class RustTranslator(
...
-                            val setterName = BuiltinName("set.${supProperty.symbol.text}").displayName
-                            instanceMethods[setterName]?.let { addAll(buildForwarder(it)) }
+                            maybeAddTraitForwarder(
+                                pos,
+                                instanceMethods,
+                                isInterface = isInterface,
+                                methodKind = MethodKind.Setter,
+                                methodName = BuiltinName("set.${supProperty.symbol.text}").displayName,
+                                superShape = supProperty,
+                            )
```
tjpalmer, line 806: I'm not sold on this factoring, but this lets me centralize some differences for classes vs interfaces for the cases above.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -708,65 +723,112 @@ class RustTranslator(
...
+        methodKind: MethodKind,
+        methodName: String,
+        superShape: VisibleMemberShape,
+        returnType: Type2? = null,
+    ) {
+        when {
+            isInterface -> buildForwarderForTrait(pos, superShape, methodKind)
+            else -> when (val method = instanceMethods[methodName]) {
+                null -> listOf()
+                else -> buildForwarder(method, returnType = returnType)
```
tjpalmer, line 829: The `else` here was the main old logic, and `isInterface` is the main new.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -816,7 +878,7 @@ class RustTranslator(
                         // TODO Only if all sealed subtypes are public?
                         id = AS_ENUM_NAME.toId(pos),
                         params = selfParams.deepCopy(),
-                        returnType = enumId.deepCopy(),
+                        returnType = enumId.makeTypeRef(generics),
```
tjpalmer, line 881: We were missing generics on these.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -890,12 +952,14 @@ class RustTranslator(
...
-        // Implement AnyValue.
+        // Implement traits including AnyValue.
         val typeRef = id.makeTypeRef(generics)
+        implTraits(pos, decl, typeRef, generics)
         Rust.Call(
             pos,
             callee = "temper_core".toKeyId(pos).extendWith("impl_any_value_trait_for_interface!"),
             args = listOf(typeRef.deepCopy()),
+            where = whereForAnyValueImpl(pos, generics),
         ).also { moduleItems.add(Rust.ExprStatement(pos, it).toItem()) }
```
tjpalmer, line 963: Add the hack `where` clause to the macro call.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -37,10 +37,19 @@ import lang.temper.type2.withType
...
+ * be good to support latest expectations.
  */
 internal fun allowWarnings(pos: Position): Rust.AttrInner = Rust.AttrInner(
     pos,
-    Rust.Call(pos, "allow".toId(pos), listOf("warnings".toId(pos))),
+    Rust.Call(
+        pos,
+        "allow".toId(pos),
+        listOf("dependency_on_unit_never_type_fallback", "warnings").map { it.toId(pos) },
+    ),
```
tjpalmer, line 52: See #359 on this.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1008,6 +1073,103 @@ class RustTranslator(
...
+    /**
+     * Build a forwarder from a trait wrapper to a trait method that *isn't*
+     * overridden in the current trait. We need this to handle methods for
+     * which we have only frontend descriptions, not tmpl.
+     */
+    private fun buildForwarderForTrait(
+        pos: Position,
+        shape: VisibleMemberShape,
+        methodKind: MethodKind,
+    ): List<Rust.Item> = run {
```
tjpalmer, line 1085: As mentioned in the comment, this code fails to reuse other codegen here because I'm basing things on frontend type information, which lacks some information that tmpl has, and it isn't in the same format. But at least it's limited in scope.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1008,6 +1073,103 @@ class RustTranslator(
...
+                            )
+                        },
+                    ),
+                )
+            }
+            MethodKind.Getter -> {
+                val methodId = shape.symbol.text.camelToSnake().toId(pos)
+                val returnType = when (val descriptor = shape.descriptor) {
+                    is Signature2 -> descriptor.returnType2
+                    is Type2 -> descriptor
```
tjpalmer, line 1135: Could be from either a property or a method. Same for setters below.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1146,13 +1308,16 @@ class RustTranslator(
...
         // Enum type.
         Rust.Enum(
             pos,
             id = enumId,
+            generics = generics,
             items = decl.typeShape.sealedSubTypes!!.map { sub ->
                 val subId = translateTypeOutName(sub.name).toId(pos)
-                Rust.EnumItemTuple(pos, id = subId, fields = listOf(Rust.TupleField(pos, type = subId.deepCopy())))
+                val subType = subId.deepCopy().makeTypeRef(generics)
+                Rust.EnumItemTuple(pos, id = subId, fields = listOf(Rust.TupleField(pos, type = subType)))
```
tjpalmer, line 1320: Fix #345.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -3276,6 +3442,22 @@ class RustTranslator(
...
+            is TmpL.NormalMethod -> member.name.name
+            // We sometimes get names like "getwhatever" rather than "get.whatever", so be explicit.
+            is TmpL.GetterOrSetter -> when (member) {
+                is TmpL.Getter -> "get"
+                is TmpL.Setter -> "set"
+            }.let { BuiltinName("$it.${member.dotName}") } // changed to "nym`...`" below
+            else -> error("unexpected")
+        }.displayName
+    }
+}
```
tjpalmer, line 3459: Factored from elsewhere.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -3364,3 +3546,5 @@ internal const val TO_LISTED_TO_LISTED_NAME = "temper_core::ToListed::to_listed"
 internal const val TRAIT_NAME_SUFFIX = "Trait"
 internal const val TYPE_ID_NAME = "std::any::TypeId"
 internal const val TYPE_ID_OF_NAME = "std::any::TypeId::of"
+
+internal val commonTypeBounds = listOf("Clone", SEND_NAME, SYNC_NAME, STATIC_LIFETIME)
```
tjpalmer, line 3550: Allows easier support on finding where custom ones end when this list is factored out.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/lib.rs
```diff
@@ -32,12 +32,14 @@ pub fn read_locked<T: Clone>(x: &Arc<RwLock<T>>) -> T {
 #[macro_export]
 macro_rules! impl_any_value_trait { // for concrete types
     // Two versions here. One for type args and one without.
-    ($type:ident<$($param:tt),*>, [$($target:ty),*]) => {
-        impl<$($param: Clone + Send + Sync + 'static),*> temper_core::AnyValueTrait for $type<$($param),*> {
+    ($type:ident$(<$($param:tt),*>)?, [$($target:ty),*] $(where $($bounds:tt)*)?) => {
+        impl$(<$($param: Clone + Send + Sync + 'static),*>)? temper_core::AnyValueTrait for $type $(<$($param),*>)?
+        $(where $($bounds)*)?
+        {
```
tjpalmer, line 38: Combine the with/out type arg versions. More importantly, support extra `where` bounds passed in.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/lib.rs
```diff
@@ -47,64 +49,46 @@ macro_rules! impl_any_value_trait { // for concrete types
...
         }
     };
-    ($type:ident, [$($target:ty),*]) => {
-        impl temper_core::AnyValueTrait for $type {
+}
+
+#[macro_export]
+macro_rules! impl_any_value_trait_for_interface { // for abstract types
+    ($type:ident $(<$($param:tt),*>)? $(where $($bounds:tt)*)?) => {
+        impl$(<$($param: Clone + Send + Sync + 'static),*>)? temper_core::AnyValueTrait for $type $(<$($param),*>)?
```
tjpalmer, line 73: Beyond matching above changes, also now implement AnyValueTrait itself instead of just AsAnyValue.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/lib.rs
```diff
@@ -47,64 +49,46 @@ macro_rules! impl_any_value_trait { // for concrete types
...
+                temper_core::AnyValueTrait::cast(&*self.0, type_id)
             }
             fn is(&self, type_id: std::any::TypeId) -> bool {
-                type_id == std::any::TypeId::of::<$type>()
-                $(|| type_id == std::any::TypeId::of::<$target>())*
+                temper_core::AnyValueTrait::is(&*self.0, type_id)
             }
             fn ptr_id(&self) -> usize {
-                std::sync::Arc::as_ptr(&self.0) as usize
+                temper_core::AnyValueTrait::ptr_id(&*self.0)
```
tjpalmer, line 83: This forwarding is approximately the same as what we generate for other impls for trait wrapper -> trait now in RustTranslator as of this pr. And I considered generating macros for each temper-interface-to-rust-trait case so we just call a macro for each trait instead of repeating impls, but I didn't chase that path.

More serious procedural macros to just prefix people's own interface and class definitions in rust for temper compatibility would be nicer than any of this, but I haven't dug into that yet, either.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/lib.rs
```diff
@@ -47,64 +49,46 @@ macro_rules! impl_any_value_trait { // for concrete types
...
-        }
-    };
-    ($type:ty) => {
-        impl temper_core::AsAnyValue for $type {
+        impl$(<$($param: Clone + Send + Sync + 'static),*>)? temper_core::AsAnyValue for $type $(<$($param),*>)?
+        $(where $($bounds)*)? 
+        {
             fn as_any_value(&self) -> temper_core::AnyValue {
-                self.0.as_any_value()
+                temper_core::AsAnyValue::as_any_value(&*self.0)
```
tjpalmer, line 91: Use the more explicit form generally for trait wrapper -> trait forwarding because it handles ambiguous cases better. That matters especially for `clone_boxed`, which I define in all our traits, but maybe could apply elsewhere, too.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -152,6 +152,14 @@ class RustBackendTest {
...
             |                  }
             |              }
+            |              impl TalkerTrait for Talker {
+            |                  fn clone_boxed(& self) -> Talker {
+            |                      TalkerTrait::clone_boxed( & ( * self.0))
+            |                  }
+            |                  fn talk(& self) -> () {
+            |                      TalkerTrait::talk( & ( * self.0))
+            |                  }
+            |              }
```
tjpalmer, line 162: All trait wrappers get trait impls now, too. We could already call the trait methods through wrappers because of Deref, but we couldn't use the wrappers as type args because they didn't actually implement the traits. Now they do.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
     }
 
+    @Test
+    fun genericConstraints() {
+        assertGenerateWanted(
+            temper = """
+                |interface A {
+                |  public var prop: String;
+                |  public get thing(): String;
+                |  public set thing(that: String): Void;
```
tjpalmer, line 1258: Both kinds of props so I could test and address both.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+                |  public greeting(): String { "Hi!" }
+                |  public whatever(): String;
+                |}
+                |interface B<T extends A> extends A {
+                |  // No greeting here.
+                |  public whatever(): String { "blah" }
```
tjpalmer, line 1264: Methods that either get overridden or not in intermediate interfaces.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+                |}
+                |class C<T extends A>(
+                |  public var prop: String,
+                |  public var thing: String,
+                |) extends B<T> {
+                |  public greeting(): String { "Ha!" }
+                |  public spawn(): C<B<A>> { new C<B<A>>("", "") }
+                |}
```
tjpalmer, line 1272: Variations on constrained type parameter usage.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+                |    fn deref(& self) -> & Self::Target {
+                |        & ( * self.0)
+                |    }
+                |}
+                |trait BTrait<T: ATrait + Clone + std::marker::Send + std::marker::Sync + 'static>: temper_core::AsAnyValue + temper_core::AnyValueTrait + std::marker::Send + std::marker::Sync + ATrait {
+                |    fn clone_boxed(& self) -> B<T>;
+                |    fn whatever(& self) -> std::sync::Arc<String> {
+                |        return std::sync::Arc::new("blah".to_string());
+                |    }
+                |}
```
tjpalmer, line 1334: Here's the actual B trait with its overridden method.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+                |        B(std::sync::Arc::new(selfish))
+                |    }
+                |}
+                |impl<T: ATrait + Clone + std::marker::Send + std::marker::Sync + 'static> BTrait<T> for B<T> {
+                |    fn clone_boxed(& self) -> B<T> {
+                |        BTrait::clone_boxed( & ( * self.0))
+                |    }
+                |    fn whatever(& self) -> std::sync::Arc<String> {
+                |        BTrait::whatever( & ( * self.0))
+                |    }
```
tjpalmer, line 1348: Calling the inner nested dyn trait instance allows dynamic dispatch. I legally can skip defining this method here because trait B already defines it, but if I did that, instances of this wrapper type would get the B version of the method called instead of dispatching to the wrapped instead which can get the base type's overrides called instead.

All this wrapping sparked my concerns again that my wrapping of Trait and Struct types is wasteful and bother. But for class struct wrappers, it means we can hide things such as whether we need mutex wrapping for mutable types. And I don't remember all the rabbit hole that got me to the wrappers, but the mutex thing is at least one big one.

And it also lets us pass around more convenient wrapper types.

And technically the wrappers could define methods with variance for when a wrapper type is being used directly rather than as a dyn trait.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+                |        self.greeting()
+                |    }
+                |    fn prop(& self) -> std::sync::Arc<String> {
+                |        self.prop()
+                |    }
+                |    fn set_prop(& self, newProp__0: std::sync::Arc<String>) {
+                |        self.set_prop(newProp__0)
+                |    }
+                |}
+                |temper_core::impl_any_value_trait!(C<T>, [B<T>, A] where T: ATrait);
```
tjpalmer, line 1444: I think the above is good translation, but I haven't tried running it. That's why some amount of this testing is also added to the funtest in this pr.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+        )
+    }
+
+    @Test
+    fun genericEnum() {
+        assertGenerateWanted(
+            temper = """
+                |export interface I {}
+                |export sealed interface Hi<T extends I> {}
+                |export class Lo<T extends I> extends Hi<T> {}
```
tjpalmer, line 1455: Both generic sealed types and bounds on the type params.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+                |        ATrait::whatever( & ( * self.0))
+                |    }
+                |    fn prop(& self) -> std::sync::Arc<String> {
+                |        ATrait::prop( & ( * self.0))
+                |    }
+                |    fn set_prop(& self, value: std::sync::Arc<String>) {
+                |        ATrait::set_prop( & ( * self.0), value)
+                |    }
+                |}
+                |temper_core::impl_any_value_trait_for_interface!(B<T> where T: ATrait);
```
tjpalmer, line 1373: Here's one example of passing in trait bound `T: ATrait` where before we only had the common ones that I add on to everything.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+                |temper_core::impl_any_value_trait_for_interface!(I);
+                |impl std::ops::Deref for I {
+                |    type Target = dyn ITrait;
+                |    fn deref(& self) -> & Self::Target {
+                |        & ( * self.0)
+                |    }
+                |}
+                |pub enum HiEnum<T: ITrait + Clone + std::marker::Send + std::marker::Sync + 'static> {
+                |    Lo(Lo<T>)
+                |}
```
tjpalmer, line 1488: Per #345, we were failing to make generic enum types. Here's part of getting it right.

And nobody forced enum support in rust for temper sealed types, but it seems like something people would want, which is why I had implemented it before.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1232,6 +1248,346 @@ class RustBackendTest {
...
+                |temper_core::impl_any_value_trait!(Lo<T>, [Hi<T>] where T: ITrait);
+            """.trimMargin(),
+        )
+    }
+
+    @Test
+    fun needlesslyGenericBuilder() {
+        assertGenerateWanted(
+            temper = """
+                |export class Ha<T>(public i: Int, public j: Int) {}
```
tjpalmer, line 1545: This duplicates #271, which happens when the type param isn't actually used. But I think I forgot to actually fix it in this pr. Maybe best put off, anyway, since this pr is already large.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1858,9 +2223,15 @@ class RustBackendTest {
...
+                |  public greet(text: String): String {
+                |    needy(this, text)
+                |  }
+                |}
+                |export let needy(a: A, text: String): String {
+                |  a.adjust(text)
                 |}
             """.trimMargin(),
-            // TODO Fix broken things below.
+            // TODO Fix broken things below. Is that just the list things that might just be variance matters?
```
tjpalmer, line 2234: I didn't comment here well enough before, but I saw this comment in passing and looked a little at the code.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1858,9 +2223,15 @@ class RustBackendTest {
                 |}
                 |export class B extends A {
                 |  public adjust(text: String): String { "${text} there" }
+                |  public greet(text: String): String {
+                |    needy(this, text)
+                |  }
+                |}
+                |export let needy(a: A, text: String): String {
+                |  a.adjust(text)
                 |}
```
tjpalmer, line 2232: I think I added this to see a case of needing to wrap `this: A` for sending it to a different function.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1900,6 +2279,10 @@ class RustBackendTest {
                 |        let text__1 = text__1.to_arc_string();
                 |        return std::sync::Arc::new(format!("{} there", text__1));
                 |    }
+                |    pub fn greet(& self, text__2: impl temper_core::ToArcString) -> std::sync::Arc<String> {
+                |        let text__2 = text__2.to_arc_string();
+                |        return needy(A::new(self.clone()), text__2.clone());
```
tjpalmer, line 2284: Here's it getting wrapped in `A::new(self.clone())` although calling `clone_boxed()` is probably better for avoiding deep wrapping. Just wrote up #360 on this.

functional-test-suite/src/commonMain/resources/interfaces/property-members/property-members.temper.md
```diff
@@ -40,3 +44,40 @@ Casting a C up to an I still allows reading x.
...
+The test above already includes type parameter bounds, but we don't test bounds
+on type parameters on types elsewhere, so include that here.
+
+    class D<T extends I>(public thing: T) {}
+
+Use both the top and intermediate interfaces here to prove we can. And store
+them in intermediate variables to ensure the types are being handled correctly.
+
+    let di = new D<I>(a as I orelse panic()); // TODO Remove cast once types delivered.
+    let dj = new D<J>(a as J orelse panic()); // TODO Remove cast once types delivered.
```
tjpalmer, line 59: This casting nonsense is because we don't actual provide the actual type information to the constructor calls here, so be-rust fails to wrap them in the trait wrapper properly. I decided to make that out of scope for this pr, since it's already big.

functional-test-suite/src/commonMain/resources/interfaces/property-members/property-members.temper.md
```diff
@@ -40,3 +44,40 @@ Casting a C up to an I still allows reading x.
...
+    }
+
+    class SSub<T extends I>(public thing: T) extends SI<T> {}
+
+    let si: SI<C> = new SSub(a);
+    console.log("Sealed thing is ${si.thing.x}.");
+
+```log
+Sealed thing is foo.
+```
```
tjpalmer, line 83: Generic sealed type testing.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -37,10 +37,19 @@ import lang.temper.type2.withType
 /**
  * We generate plenty of warnings. Some we could try to clean up, but it's awkward.
  * Ignoring warnings means we have to pay manual attention to generated public names.
+ *
+ * Also allow `dependency_on_unit_never_type_fallback` in our generated code for now.
+ * This was a backward incompatible change in newer versions of Rust, even without a
+ * change in edition. We should investigate this more in the future, though. It would
```
mikesamuel, line 43: I'm glad you're keeping on top of these happenings.

PR #358, opened 2026-02-25
lexer: simplify the way the lexer categorizes `<` and `>` tokens

mikesamuel:
This commit does a round of lexer simplification that should make it easier for TextMate and similar grammars to accurately mimic what the Temper compiler does.

This is timely, because I need to implement the simpler multi-line string parsing strategy.

The old approach was to lookahead pairing brackets, and to treat angle brackets as paired when they could and there are not certain, carefully chosen "interrupting" tokens between them at the same level of nesting.

That approach had some nice properties but also led to some surprising results.

    f(a < b, c > d)

The new approach relies on a space-sensitive approach:

- When a `<` token is not preceded by a space or comment token then it is a bracket token.
- When outputing a `<` bracket token the lexer increments a counter.
- When a `>` token is seen, if that count is non-zero, we classify the token as a bracket and decrement the count.

For example:

    C<A, B>      // Angle brackets, no space between C and <
    c < a        // Comparison
    a < b, c > d // Two comparisons, as above


builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -1331,6 +1330,73 @@ object BuiltinFuns {
      *
      * See the [snippet/general-comparison/algo] for details of how they are compiled and
      * especially the [snippet/general-comparison/caveats].
+     *
+     * ⎀ syntax/less-than-space-sensitivity
+     *
+     * <!-- snippet: syntax/less-than-space-sensitivity -->
+     * # Syntactic corner case: `<` ambiguity
+     *
+     * Tldr: always put spaces around infix operators like `<`.
```
mikesamuel, line 1339: Added documentation of the ambiguity rules and how to avoid them.

fundamentals/src/commonMain/resources/lang/temper/helpful/helpful-snippets.json
mikesamuel, line 1: Opportunistically added information to the builtin comparison operators so that `help( < )` in the REPL will show the reference documentation article.

release-notes/snippets/lt-ambiguity-resolution-change.md
mikesamuel, line 1: Added a release note since this is not a completely backwards compatible change.

PR #361, opened 2026-02-26
secure-composition: callouts allow explaining problems with templates

mikesamuel:
Now, OptimizeContextualAutoescapingBlocks creates a callout function.

Parts of the secure composition library take the callout which gives it a channel to communicate error messages back based on its analysis of the fixed parts but without wasting time constructing messages at runtime.

Now the optimizer collects error messages and reports them via the log sink, giving developers feedback about problems in their templates early.

For example, given HTML like:

    <a href=...

It can tell the tag is missing a `>`.

### DIFF
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/function/OptimizeContextualAutoescapingBlocks.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/function/OptimizeContextualAutoescapingBlocks.kt
@@ -11,6 +11,7 @@ import lang.temper.common.buildListMultimap
 import lang.temper.common.console
 import lang.temper.common.firstOrNullAs
 import lang.temper.common.putMultiList
+import lang.temper.common.putMultiSet
 import lang.temper.common.subListToEnd
 import lang.temper.env.InterpMode
 import lang.temper.frontend.InterpretationContext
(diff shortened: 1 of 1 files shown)


builtin/src/commonMain/kotlin/lang/temper/builtin/TestMacro.kt
```diff
@@ -114,6 +114,8 @@ internal object TestMacro : BuiltinMacro("test", null) {
         funTree.insert(beforeBlock) {
             V(vWordSymbol)
             V(Value(funName.toSymbol()))
+            V(vTestSymbol)
+            V(Value(name, TString))
```
mikesamuel: It's nice to be able to look up from a macro, see if we're in a testing context, and downgrade error messages to info.  That way we're not complaining about errors in HTML in tests for HTML corner cases.

Some kind of `@Suppress` mechanism would be more targeted longer term.

frontend/src/commonMain/kotlin/lang/temper/frontend/syntax/SyntaxMacroStage.kt
```diff
@@ -124,8 +124,14 @@ private fun declareModuleConsole(root: BlockTree) {
         Decl {
             Ln { it.unusedTemporaryName("console").also { name -> consoleName = name } }
             V(vInitSymbol)
-            Call {
-                Rn(getConsoleBuiltinName)
+            Call(BuiltinFuns.vDoPure) {
+                Fn {
+                    Block {
+                        Call {
+                            Rn(getConsoleBuiltinName)
```
mikesamuel: Adding console logging to the library code keeps making it fail during expansion time which was making me sad.

tjpalmer (approved):

PR #363, opened 2026-02-27
Allow console to be used by macro code.

mikesamuel:
Uses of `console` as a free name in a module body get rewritten to a local reference that caches `builtin.getConsole()`.

This allows translating to systems that have a logger per module.

    console.log("Hello, World!");

    // desugars to

    let console#0 = getConsole();  // SHARED & HOISTED
    console#0.log("Hello, World!");

For macros to speculatively execute some user code, it's nice if that user code can add logging for printf debugging.

This commit changes that `getConsole()` call to happen inside a `doPure { ... }` block.

    let console#0 = doPure { (): Console => // SHARED & HOISTED
      getConsole()
    };
    console#0.log("Hello, World!");

This commit also fixes up test goldens that have to change as a result.

This change also flushed out a number of typer bugs, fixed herein, because `doPure` is a lambda.  Specifically, if you have a block receiver that is generic, you don't want to use a function type for the block that leaks type parameter references out of scope to auto-type the lambdas inputs and outputs.  That exposed some weaknesses in how we deal with chains of calls of block receivers like `myList.map { ... }.filter { ... }`.  Now, after typing a call we are sure to roll information from the call's computed type parameters back into the block.

### DIFF
--- a/be-java/src/commonTest/kotlin/lang/temper/be/java/JavaBackendTest.kt
+++ b/be-java/src/commonTest/kotlin/lang/temper/be/java/JavaBackendTest.kt
@@ -883,17 +883,17 @@ class JavaBackendTest {
             |                              }
             |                              static Generator<Optional<? super Object>> fn__0() {
             |                                  class Local_1 {
-            |                                      int caseIndex_38 = 0;
+            |                                      int caseIndex_41 = 0;
             |                                  }
             |                                  final Local_1 local$1 = new Local_1();
-            |                                  Function<Generator<Optional<? super Object>>, Result<Optional<? super Object>>> convertedCoroutine_42 = generator_37 -> {
-            |                                      int caseIndexLocal_39 = local$1.caseIndex_38;
-            |                                      local$1.caseIndex_38 = -1;
-            |                                      switch (caseIndexLocal_39) {
+            |                                  Function<Generator<Optional<? super Object>>, Result<Optional<? super Object>>> convertedCoroutine_45 = generator_40 -> {
+            |                                      int caseIndexLocal_42 = local$1.caseIndex_41;
+            |                                      local$1.caseIndex_41 = -1;
+            |                                      switch (caseIndexLocal_42) {
             |                                          case 0:
             |                                              {
             |                                              console_5.log("foo");
-            |                                              local$1.caseIndex_38 = 1;
+            |                                              local$1.caseIndex_41 = 1;
             |                                              return new ValueResult<>(Optional.empty());
             |                                          }
             |                                          case 1:
@@ -907,7 +907,7 @@ class JavaBackendTest {
             |                                          }
             |                                      }
             |                                  };
-            |                                  return Core.safeAdaptGeneratorFn(convertedCoroutine_42 :: apply);
+            |                                  return Core.safeAdaptGeneratorFn(convertedCoroutine_45 :: apply);
             |                              }
             |                              static {
             |                                  console_5 = Core.getConsole(Logger.getLogger("my_test_library.test"));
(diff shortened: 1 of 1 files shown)


builtin/src/commonMain/kotlin/lang/temper/builtin/DoPureFn.kt
```diff
@@ -84,9 +95,19 @@ object DoPureFn : SpecialFunction, NamedBuiltinFun {
...
                     if (returnName != null) {
+                        val body = parts.body
+                        val singleStmt: Tree? = singleStmtIgnoringVoids(body)
+                        if (singleStmt != null && singleStmt is CallTree && isAssignment(singleStmt) &&
+                            (singleStmt.child(1) as? LeftNameLeaf)?.content == returnName
+                        ) {
+                            val expr = singleStmt.child(2)
+                            Replant(freeTree(expr))
+                            return@replaceMacroCallWith
+                        }
```
mikesamuel, line 106: By not inlining a block when we've got one expression, we can avoid a lot of spurious test differences

frontend/src/commonMain/kotlin/lang/temper/frontend/implicits/ImplicitsModule.kt
```diff
@@ -118,6 +118,7 @@ object ImplicitsModule {
             ),
         )
         val endStage = Stage.Run // TODO Go back to Stage.GenerateCode once `console` is a stable value.
+        // do not commit: can this go away?
```
mikesamuel, line 121: TODO: Mike, try this.  Also, in Implicits can getConsole just return new GlobalConsole?
Letting us do away with `let console` in Implicits entirely.

frontend/src/commonMain/kotlin/lang/temper/frontend/syntax/SyntaxMacroStage.kt
```diff
@@ -125,7 +128,16 @@ private fun declareModuleConsole(root: BlockTree) {
...
+                V(BuiltinFuns.vDoPure)
+                Fn {
+                    V(outTypeSymbol)
+                    V(Value(ReifiedType(WellKnownTypes.consoleType2)))
+                    Block {
+                        Call {
+                            Rn(getConsoleBuiltinName)
+                        }
+                    }
+                }
```
mikesamuel, line 140: This is the bulk of the change

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/MakeResultsExplicit.kt
```diff
@@ -194,16 +204,10 @@ internal class MakeResultsExplicit private constructor(
                 pos,
                 listOf(outputName) +
                     if (returnTypeTree != null) {
-                        val nameForReturnType = shareReferenceTo(
```
mikesamuel, line 197: Turns out this was unnecessary.  We've only got this if we cut it out of \outType metadata.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -1114,6 +1128,18 @@ internal class Typer(
                                 childI.content.typeTag.name.displayName.asciiUnTitleCase(),
                             ),
                         )
+                    childI is FunTree && childI.parts != null -> {
+                        val parts = childI.parts!!
+                        InputBound.LambdaBound(
```
mikesamuel, line 1133: The new LambdaBound helps when we have some information about input and output types from the block but not enough for a function type.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -3113,9 +3152,40 @@ private enum class LateType2CheckResult {
     Immediate,
 }
 
-private fun TypeContext2.valueFormalTypeAt(variants: Iterable<Callee>, index: Int): Type2? {
-    val ts = variants.mapNotNull {
-        it.sig.valueFormalForActual(index)?.type
+private fun TypeContext2.valueFormalTypeAt(variants: Iterable<Callee>, index: Int, typeActuals: List<Type2>?): Type2? {
```
mikesamuel, line 3155: Reworked this to detect the case where there's an out-of-scope use of a receiver's type parameter.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/TyperTest.kt
```diff
@@ -1848,6 +1848,54 @@ class TyperTest {
...
+            |
+            |    export let x = doPure { (): C => makeC() };
+            |
+            |    x
+            |/// ╹     : C
+        """.trimMargin(),
+    )
+
+    @Test
+    fun chainedLambdas() = assertTypes(
```
mikesamuel, line 1866: Adopted some tests from the function functional tests.

fundamentals/src/commonMain/kotlin/lang/temper/type/Types.kt
```diff
@@ -832,9 +832,9 @@ val StaticType.mentionsInvalid: Boolean get() = when (this) {
 }
 
 /** Determine if any component is invalid. */
-val Type2.mentionsInvalid: Boolean get() =
-    this.definition == WellKnownTypes ||
```
mikesamuel, line 836: Don't know what this was trying to do.  Obv a bug.

fundamentals/src/commonMain/kotlin/lang/temper/type2/InferBounds.kt
mikesamuel, line 1: It occurred to me that one way to learn things about a partially typed lambda block is to just imagine the block receiver calls its block.  That let's us add constraints for information we have.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -67,7 +67,7 @@ class PyBackendTest {
             |class IntMaker:
             |    radix_7: 'int3'
             |    __slots__ = ('radix_7',)
-            |    def int64_to_int(this_0, int_9: 'int64_21') -> 'int3':
+            |    def int64_to_int(this_0, int_9: 'int64_22') -> 'int3':
```
tjpalmer, line 70: Been biased toward reducing our uniquifying to actually needed cases sometime.

fundamentals/src/commonMain/kotlin/lang/temper/type/Types.kt
tjpalmer: 😮

PR #364, opened 2026-02-27
Call trait methods from class impls

tjpalmer:
- This is one of the focus issues right now, but others remain to be fixed
- Also ignore warnings before ignoring unit never type fallback

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -304,14 +304,7 @@ class RustTranslator(
         }
     }
 
-    /**
-     * Generate helpers for using named fields for making new class instances.
-     * These are called `...Maker` rather than `...Builder` because it's less
-     * likely to interfere with user-named Builder types. Temper has a style of
-     * naming things `...Builder`, so Temper users might reach for that naming
-     * easily in their own types, and we want to reduce collision risk.
-     */
-    private fun processForClassMaker(
+    private fun processForClassBuilder(
         fn: TmpL.FunctionDeclarationOrMethod,
         enclosingType: Rust.Type? = null,
         generics: List<Rust.GenericParam> = listOf(),
@@ -322,8 +315,8 @@ class RustTranslator(
         fn.parameters.parameters.count { it.name != fn.parameters.thisName } <= 1 && return
         // And for now, skip those with rest parameters. TODO Extract to list value?
         fn.parameters.restParameter != null && return
-        // Build the maker.
-        // Here we make `WhateverMaker` for requireds and/or `WhateverMakerOptions` structs for optionals.
+        // Build the builder.
+        // Here we make `WhateverBuilder` for requireds and/or `WhateverOptions` structs for optionals.
         // Alternatively, could make a Java-style builder, which is common in Rust, but it takes less advantage of
         // standard static checking that we get with pub struct fields.
         val pos = fn.pos
@@ -334,7 +327,7 @@ class RustTranslator(
             else -> enclosingType as Rust.Id
         }
         // TODO Ensure unique names.
-        val builderId = "${targetId.outName.outputNameText}Maker".toId(targetId.pos)
+        val builderId = "${targetId.outName.outputNameText}Builder".toId(targetId.pos)
         val optionsId = "${targetId.outName.outputNameText}Options".toId(targetId.pos)
         // Figure out if we have requireds and/or optionals.
         // We need to separate these in Rust because requireds often can't Default.
@@ -402,7 +395,7 @@ class RustTranslator(
                         items = buildList {
                             Rust.Function(
                                 pos,
-                                id = "make".toId(pos),
+                                id = "build".toId(pos),
                                 // Pass self by move on purpose in these, so we can avoid cloning.
                                 // We make builder types Clone in case anyone badly wants copies on their own.
                                 params = listOf(self),
@@ -429,7 +422,7 @@ class RustTranslator(
                 items = buildList {
                     Rust.Function(
                         pos,
-                        id = "make".toId(pos),
+                        id = "build".toId(pos),
                         params = listOf(self),
                         returnType = rustReturnType,
                         block = Rust.Block(
@@ -439,7 +432,7 @@ class RustTranslator(
                                 optionals.isEmpty() -> callNew(requireds)
                                 // Delegate to the with-optionals builder.
                                 else -> self.deepCopy().methodCall(
-                                    key = "make_with",
+                                    key = "build_with",
                                     args = listOf(makePath(pos, "std", "default", "Default", "default").call()),
                                 )
                             },
@@ -449,7 +442,7 @@ class RustTranslator(
                         val optionsArg = "options".toId(pos)
                         Rust.Function(
                             pos,
-                            id = "make_with".toId(pos),
+                            id = "build_with".toId(pos),
                             generics = optionalGenerics.deepCopy(),
                             params = listOf(
                                 self.deepCopy(),
@@ -1940,7 +1933,7 @@ class RustTranslator(
             enclosingTypePub != null &&
             enclosingTypePub.scope == null
         ) {
-            processForClassMaker(constructor, enclosingType, generics, returnType = returnType)
+            processForClassBuilder(constructor, enclosingType, generics, returnType = returnType)
         }
         // Actually translate the constructor now, with all our adjustments.
         return translateFunctionDeclarationOrMethod(
--- a/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
+++ b/be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
@@ -1193,11 +1193,11 @@ class RustBackendTest {
                 |#[derive(Clone)]
                 |pub struct C(std::sync::Arc<std::sync::RwLock<CStruct>>);
                 |#[derive(Clone)]
-                |pub struct CMaker {
+                |pub struct CBuilder {
                 |    pub x: std::sync::Arc<String>, pub y: std::sync::Arc<String>
                 |}
-                |impl CMaker {
-                |    pub fn make(self) -> C {
+                |impl CBuilder {
+                |    pub fn build(self) -> C {
                 |        C::new(self.x, self.y)
                 |    }
                 |}
@@ -1750,7 +1750,7 @@ class RustBackendTest {
     }
 
     @Test
-    fun needlesslyGenericMaker() {
+    fun needlesslyGenericBuilder() {
         assertGenerateWanted(
             temper = """
                 |export class Ha<T>(public i: Int, public j: Int) {}
@@ -1768,11 +1768,11 @@ class RustBackendTest {
                 |#[derive(Clone)]
                 |pub struct Ha<T: Clone + std::marker::Send + std::marker::Sync + 'static>(std::sync::Arc<HaStruct<T>>);
                 |#[derive(Clone)]
-                |pub struct HaMaker {
+                |pub struct HaBuilder {
                 |    pub i: i32, pub j: i32
                 |}
-                |impl HaMaker {
-                |    pub fn make(self) -> Ha<T> {
+                |impl HaBuilder {
+                |    pub fn build(self) -> Ha<T> {
                 |        Ha::new(self.i, self.j)
                 |    }
                 |}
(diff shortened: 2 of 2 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -43,14 +43,12 @@ import lang.temper.type2.withType
...
-internal fun allowWarnings(pos: Position): Rust.AttrInner = Rust.AttrInner(
-    pos,
-    Rust.Call(
-        pos,
-        "allow".toId(pos),
-        listOf("dependency_on_unit_never_type_fallback", "warnings").map { it.toId(pos) },
-    ),
-)
+internal fun allowWarnings(pos: Position) = run {
+    // Separate them with `warnings` first, to prevent warnings against the other on older rust.
```
tjpalmer, line 47: At least, I seemed to see effects per the comment above, but I didn't spend a lot of time on this aspect.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -94,6 +98,7 @@ class RustTranslator(
     private var insideMutableType = false
     private val failVars = mutableSetOf<ResolvedName>()
     private val functionContextStack = mutableListOf<FunctionContext>()
+    private val logSink = LogSink.devNull // TODO what?
```
tjpalmer, line 101: I need some log sink, and this works for now, but maybe should reconsider or even just pass log sinks into translators. I've wanted that more than once before.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -309,8 +322,8 @@ class RustTranslator(
         fn.parameters.parameters.count { it.name != fn.parameters.thisName } <= 1 && return
         // And for now, skip those with rest parameters. TODO Extract to list value?
         fn.parameters.restParameter != null && return
-        // Build the builder.
-        // Here we make `WhateverBuilder` for requireds and/or `WhateverBuilderOptions` structs for optionals.
+        // Build the maker.
+        // Here we make `WhateverMaker` for requireds and/or `WhateverMakerOptions` structs for optionals.
```
tjpalmer: I think Builder is more idiomatic, but I think people will cope.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1073,59 +1087,108 @@ class RustTranslator(
...
 
+    private fun buildForwarderFromClassToTrait(
+        pos: Position,
+        decl: TmpL.TypeDeclaration,
+        superShape: VisibleMemberShape,
+        methodKind: MethodKind,
+    ): List<Rust.Item> = run {
+        // We're here because this class has no matching member, so walk its supertypes to match the super method.
+        // The method we inherit closest might be on a different branch.
+        val overrides = findOverrides(decl.typeShape, superShape, typeContext, logSink)
```
tjpalmer, line 1091: I factored out `findOverrides` from existing logic. I think it meets the needs here.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1073,59 +1087,108 @@ class RustTranslator(
...
+        buildForwarderToTrait(pos, targetType, superShape, methodKind) result@{ traitType, methodId, argIds ->
+            Rust.Call(
+                pos,
+                callee = traitType.extendWith(methodId.deepCopy()),
+                args = buildList {
+                    add("self".toKeyId(pos))
+                    addAll(argIds)
+                },
+            )
+        }
```
tjpalmer, line 1123: This is now factored to share logic with trait-wrapper-to-trait forwarding that I put in on the previous pr. The main difference is what happens inside the forwarding method, so that's what the callback is for.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1073,59 +1087,108 @@ class RustTranslator(
...
+        buildForwarderToTrait(pos, shape.enclosingType, shape, methodKind) result@{ traitType, methodId, argIds ->
+            Rust.Call(
+                pos,
+                callee = traitType.extendWith(methodId.deepCopy()),
+                args = buildList {
+                    add("self".toKeyId(pos).member("0", notMethod = true).deref().ref())
+                    addAll(argIds)
+                },
+            )
+        }
```
tjpalmer, line 1146: This retains behavior from before while factoring most of the logic to share with the new class-to-trait forwarding above.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1073,59 +1087,108 @@ class RustTranslator(
...
-                                args = buildList {
-                                    add(selfArg)
-                                    for (argName in argNames) {
-                                        add(argName.toId(pos))
-                                    }
-                                },
-                            )
-                        },
-                    ),
+                    block = Rust.Block(pos, result = traitType?.let { buildResult(it, methodId, argIds) }),
```
tjpalmer, line 1188: Actual method content is now handled per above callbacks.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -67,7 +67,8 @@ class RustBackendTest {
             |      src: {
             |        lib.rs: {
             |          content: ```
-            |            #![allow(dependency_on_unit_never_type_fallback, warnings)]
+            |            #![allow(warnings)]
+            |            #![allow(dependency_on_unit_never_type_fallback)]
```
tjpalmer, line 71: Here's the new split with warnings ignored first.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1187,11 +1193,11 @@ class RustBackendTest {
...
                 |pub struct C(std::sync::Arc<std::sync::RwLock<CStruct>>);
                 |#[derive(Clone)]
-                |pub struct CBuilder {
+                |pub struct CMaker {
                 |    pub x: std::sync::Arc<String>, pub y: std::sync::Arc<String>
                 |}
-                |impl CBuilder {
-                |    pub fn build(self) -> C {
+                |impl CMaker {
+                |    pub fn make(self) -> C {
```
tjpalmer: New "Maker" vs old "Builder". Yes, we still need to address the general name handling problem sometime. But this seems less likely to conflict given our naming styles in temper.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1270,6 +1277,19 @@ class RustBackendTest {
...
+                |  public set prop(value: String): Void {}
+                |  public set thing(value: String): Void {}
+                |  public whatever(): String { "sure" }
+                |}
+                |// E provides indirection on type bindings to D.
+                |interface E<T> extends D<T> {}
+                |// Alternate prop/thing set/get vs D above.
+                |class F extends B<C<A>> & E<Int> {
+                |  public get thing(): String { "Hello!" }
+                |}
```
tjpalmer, line 1292: Lots of tricky weaving above. Probably should add things like this to a funtest sometime, but this lets me test things for rust for now, at least. It sure explodes in generate code lines below, though.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1420,6 +1440,9 @@ class RustBackendTest {
                 |    fn clone_boxed(& self) -> B<T> {
                 |        B::new(self.clone())
                 |    }
+                |    fn whatever(& self) -> std::sync::Arc<String> {
+                |        BTrait::whatever(self)
+                |    }
```
tjpalmer: Explicitly call B's `whatever`, since that's the resolution found.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1442,6 +1468,209 @@ class RustBackendTest {
...
+                |    fn whatever(& self) -> std::sync::Arc<String> {
+                |        ATrait::whatever( & ( * self.0))
+                |    }
+                |    fn prop(& self) -> std::sync::Arc<String> {
+                |        ATrait::prop( & ( * self.0))
+                |    }
+                |    fn set_prop(& self, value: std::sync::Arc<String>) {
+                |        ATrait::set_prop( & ( * self.0), value)
+                |    }
+                |}
```
tjpalmer, line 1529: This trait wrapper to trait stuff isn't new, but `interface D` is new in the test case, so it's extra lines in the codegen here.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1442,6 +1468,209 @@ class RustBackendTest {
...
+                |    }
+                |    fn whatever(& self) -> std::sync::Arc<String> {
+                |        BTrait::whatever(self)
+                |    }
+                |    fn prop(& self) -> std::sync::Arc<String> {
+                |        DTrait::prop(self)
+                |    }
+                |    fn set_prop(& self, value: std::sync::Arc<String>) {
+                |        DTrait::set_prop(self, value)
+                |    }
```
tjpalmer, line 1637: Call A for greeting implicitly, B for whatever, and own F impl for thing. Others are picked up by D.

This looks correct to me by eyeball check vs the temper code above, but I haven't tested the new code to compile and run with rust yet.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1539,7 +1768,7 @@ class RustBackendTest {
     }
 
     @Test
-    fun needlesslyGenericBuilder() {
+    fun needlesslyGenericMaker() {
```
tjpalmer: This case still is for testing #271, and I still haven't fixed that issue.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/OverrideLinker.kt
```diff
@@ -61,8 +74,7 @@ internal fun linkOverrides(
             true
         }
     }
-
-    member.overriddenMembers = overriddenMembers.toSet()
+    return overriddenMembers.toSet()
 }
```
tjpalmer, line 78: Factored out a new public function for use in be-rust. I can't just use `overriddenMembers` (as also discussed in comments), because I need to know the overrides for methods that don't exist directly in the base type.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
tjpalmer: > Explicitly call B's `whatever`, since that's the resolution found.

Actually, that just results in infinite recursion when the impl is for B, because it just comes back here. Instead just allow the BTrait default implementation to be called implicitly.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -321,7 +334,7 @@ class RustTranslator(
             else -> enclosingType as Rust.Id
         }
         // TODO Ensure unique names.
-        val builderId = "${targetId.outName.outputNameText}Builder".toId(targetId.pos)
+        val builderId = "${targetId.outName.outputNameText}Maker".toId(targetId.pos)
```
mikesamuel: Do you prefer Maker to Builder?

tjpalmer: > Do you prefer Maker to Builder?

I prefer Builder because it's more idiomatic. I was just trying to be friendlier with Temper conventions.

That said, I have a new idea. I was trying to make one branch for all the outstanding issues again, but since I've gotten the pivot, I can't finish it immediately, anyway. I'll back out the Builder changes from this pr, so I can reassess later.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -310,7 +316,7 @@ class RustTranslator(
         // And for now, skip those with rest parameters. TODO Extract to list value?
         fn.parameters.restParameter != null && return
         // Build the builder.
-        // Here we make `WhateverBuilder` for requireds and/or `WhateverBuilderOptions` structs for optionals.
+        // Here we make `WhateverBuilder` for requireds and/or `WhateverOptions` structs for optionals.
```
tjpalmer, line 319: Noticed in passing that this comment was wrong vs current behavior.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
tjpalmer: I've backed out the Builder renaming and written issue #367 about a different strategy to be both idiomatic and more often avoid naming conflicts.

PR #368, opened 2026-03-04
Supply core library for required backends

tjpalmer:
- This change will be explicitly tested in an external project once this pr goes in

### DIFF
--- a/cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
+++ b/cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
@@ -73,6 +73,7 @@ class BuildTest {
     }
 
     @Test
+    @Timeout(JAVA_TIMEOUT_SECONDS) // Just keep seeing this one specifically time out for some reason.
     fun csharpBackendDirs() = runBuildTest("CSharpBackendDirs", path = "/buildDirs") { topDir ->
         runBuild(backends = listOf(CSharpBackend.Factory.backendId), workRoot = topDir)
         topDir.withTextOf("temper.keep/csharp/apple/name-selection.json") { text ->


cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
```diff
@@ -73,6 +73,7 @@ class BuildTest {
     }
 
     @Test
+    @Timeout(JAVA_TIMEOUT_SECONDS) // Just keep seeing this one specifically time out for some reason.
```
tjpalmer, line 76: I got tired of this timing out in ci.

tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Build.kt
```diff
@@ -393,6 +390,12 @@ fun doOneBuild(build: Build): BuildResult {
         },
     )
 
+    // For supplying core library, buckets don't really matter.
+    val flatBackendIds = backendOrganization.backendBuckets.flatten()
+    supplyCoreLibrary(harness.outputFileSystem, flatBackendIds, cancelGroup, harness.cliConsole)
+
+    build.beforeStartTranslation?.await()
```
tjpalmer, line 397: This just moved down to after the backend organization instead of before.

cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
mikesamuel, line 1: Sometimes I think one test just runs early and so gets debited all the module loading and cache filling that's shared with subsequent test cases.

tjpalmer: > Sometimes I think one test just runs early and so gets debited all the module loading and cache filling that's shared with subsequent test cases.

Yeah, could be. If we start seeing another one be slow, we could just give extra time to the whole class maybe.

PR #369, opened 2026-03-06
Apply backend bindings for watch

tjpalmer:
- Tested with an external backend, where this was failing previously:

```
Watch starting build #0 at 2026-03-06T19:01:44.840Z
Finished build #0 in 3.588s
Watch starting build #1 at 2026-03-06T19:02:07.253Z
Finished build #1 in 329.622ms
Watch starting build #2 at 2026-03-06T19:02:36.885Z
Finished build #2 in 223.697ms
```

### DIFF
--- a/tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Build.kt
+++ b/tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Build.kt
@@ -115,14 +115,7 @@ fun prepareBuild(
         return null
     }
 
-    val pluggedInConfig = moduleConfig.copy(
-        moduleCustomizeHook = { module, isNew ->
-            moduleConfig.moduleCustomizeHook.customize(module, isNew)
-            for (backendId in supportedBackends) {
-                module.addEnvironmentBindings(lookupFactory(backendId)!!.environmentBindings)
-            }
-        },
-    )
+    val pluggedInConfig = plugInBackendConfigs(moduleConfig)
     return Build(
         harness = h,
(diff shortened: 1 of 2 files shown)


tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Build.kt
```diff
@@ -131,6 +124,19 @@ fun prepareBuild(
     )
 }
```
tjpalmer, line 125: I also looked into making Watcher call prepareBuild, but there's a larger gap for that change. The down side here is if people add other things in the middle of prepareBuild, but I think this is ok for now, anyway.

tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Build.kt
```diff
@@ -131,6 +124,19 @@ fun prepareBuild(
...
+    moduleConfig: ModuleConfig,
+    backends: List<BackendId> = supportedBackends,
+): ModuleConfig = ModuleConfig(
+    moduleCustomizeHook = { module, isNew ->
+        moduleConfig.moduleCustomizeHook.customize(module, isNew)
+        for (backendId in backends) {
+            module.addEnvironmentBindings(lookupFactory(backendId)!!.environmentBindings)
+        }
+    },
+)
```
tjpalmer, line 138: I could make a superficial test of just this function, but testing for temper watch generally is awkward because of how much infrastructure is currently pulled in there.

PR #371, opened 2026-03-12
cli/repl: Fix importing from within repl

mikesamuel:
Recent changes to Build.kt lost the mayRun bit which broke importing in the REPL.

This commit fixes the root problem: deriving one ModuleConfig from another.

And it changes a ReplTest case to read a static property which holds a class instance, something that is unlikely to work unless modules imported by the REPL are going through RuntimeEmulationStage.

### DIFF
--- a/cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
+++ b/cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
@@ -1150,15 +1150,19 @@ class ReplTest {
             fs.write(filePath("config.temper.md"), "# hello".toByteArray())
             fs.write(
                 filePath("src.temper"),
-                """export let message(): String { "Hello, World!" }""".toByteArray(),
+                """
+                    |export class Message(public text: String) {
+                    |  public static hello: Message = new Message("Hello, World!");
+                    |}
+                """.trimMargin().toByteArray(),
             )
         }
         repl.processLine("""help("${AvailableImports.NAME}")""")
         assertPendingContains(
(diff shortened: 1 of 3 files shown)


cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
```diff
@@ -1150,15 +1150,19 @@ class ReplTest {
             fs.write(filePath("config.temper.md"), "# hello".toByteArray())
             fs.write(
                 filePath("src.temper"),
-                """export let message(): String { "Hello, World!" }""".toByteArray(),
+                """
+                    |export class Message(public text: String) {
+                    |  public static hello: Message = new Message("Hello, World!");
```
mikesamuel, line 1155: This is unlikely to be properly initialized unless the build the REPL does for external modules is getting *mayRunModuleConfig*.

tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Build.kt
```diff
@@ -128,7 +128,7 @@ fun prepareBuild(
 fun plugInBackendConfigs(
     moduleConfig: ModuleConfig,
     backends: List<BackendId> = supportedBackends,
-): ModuleConfig = ModuleConfig(
+): ModuleConfig = moduleConfig.copy(
```
mikesamuel, line 131: This is the operative change: reuse all of moduleConfig, don't just recreate one with one piece and the rest defaults.

PR #372, opened 2026-03-13
Small changes for correctness

ShawSumma:
I used the AI to identify and fix small problems in the backends, mostly focusing on correctness, but also doing some inlining.

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
@@ -683,29 +683,12 @@ private class EqNeGeneric(
 private val eqGeneric = EqNeGeneric("EqGeneric", BuiltinOperatorId.EqGeneric, RustOperator.Equals)
 private val eqIntInt = Infix("EqIntInt", BuiltinOperatorId.EqIntInt, RustOperator.Equals)
 private val eqStrStr = CmpStrStr("EqStrStr", BuiltinOperatorId.EqStrStr, RustOperator.Equals)
-private val float64Abs = MethodCall("Float64::abs", "abs")
-private val float64Acos = MethodCall("Float64::acos", "acos")
-private val float64Asin = MethodCall("Float64::asin", "asin")
-private val float64Atan = MethodCall("Float64::atan", "atan")
-private val float64Atan2 = MethodCall("Float64::atan2", "atan2")
-private val float64Ceil = MethodCall("Float64::ceil", "ceil")
-private val float64Cos = MethodCall("Float64::cos", "cos")
-private val float64Cosh = MethodCall("Float64::cosh", "cosh")
-private val float64Exp = MethodCall("Float64::exp", "exp")
 private val float64Expm1 = MethodCall("Float64::expm1", "exp_m1")
-private val float64Floor = MethodCall("Float64::floor", "floor")
 private val float64Log = MethodCall("Float64::log", "ln")
-private val float64Log10 = MethodCall("Float64::log10", "log10")
 private val float64Log1p = MethodCall("Float64::log1p", "ln_1p")
 private val float64Max = FunctionCall("Float64::max", "temper_core::float64::max")
 private val float64Min = FunctionCall("Float64::min", "temper_core::float64::min")
 private val float64Near = FunctionCall("Float64::near", "temper_core::float64::near")
-private val float64Round = MethodCall("Float64::round", "round")
-private val float64Sin = MethodCall("Float64::sin", "sin")
-private val float64Sinh = MethodCall("Float64::sinh", "sinh")
-private val float64Sqrt = MethodCall("Float64::sqrt", "sqrt")
-private val float64Tan = MethodCall("Float64::tan", "tan")
-private val float64Tanh = MethodCall("Float64::tanh", "tanh")
 
 private object Float64E : Constant("Float64::e") {
     override fun value(pos: Position) = makePath(pos, "std", "f64", "consts", "E")
@@ -754,10 +737,6 @@ private object Int64ToInt32Unsafe : Cast("Int64::toInt32Unsafe") {
     override fun buildType(pos: Position) = "i32".toId(pos)
 }
 
-private val int32Max = MethodCall("Int32::max", "max")
-private val int32Min = MethodCall("Int32::min", "min")
-private val int64Max = MethodCall("Int64::max", "max")
-private val int64Min = MethodCall("Int64::min", "min")
 internal val intToString = FunctionCall("Int32::toString", "temper_core::int_to_string")
 private val int64ToFloat64 = FunctionCall("Int64::toFloat64", "temper_core::int64_to_float64")
 private val int64ToInt32 = FunctionCall("Int64::toInt32", "temper_core::int64_to_int32")
@@ -792,8 +771,6 @@ private val listedIsEmpty =
 private val listedJoin = FunctionCall("Listed::join", "temper_core::listed::join", hasGeneric = true, fnIndex = -1)
 private val listedLength = FunctionCall(listedTypes.map { "$it::length" }, "$LISTED_TRAIT_NAME::len", hasGeneric = true)
 private val listedMap = FunctionCall("Listed::map", "temper_core::listed::map", hasGeneric = true, fnIndex = -1)
-private val listedMapDropping =
-    FunctionCall("Listed::mapDropping", "temper_core::listed::map_dropping", hasGeneric = true, fnIndex = -1)
 private val listedReduce =
     FunctionCall("Listed::reduce", "temper_core::listed::reduce", hasGeneric = true, fnIndex = -1)
 private val listedReduceFrom =
@@ -996,8 +973,6 @@ private val stringBuilderAppendCodePoint =
 private val stringBuilderToString =
     FunctionCall("StringBuilder::toString", "temper_core::string::builder::to_string")
 
-private val stringIsEmpty = MethodCall("String::isEmpty", "is_empty")
-
 private object StringIndexNone : Constant("StringIndex::none") {
     override fun value(pos: Position) = "()".toId(pos)
 }
@@ -1034,30 +1009,13 @@ private val connectedReferences = listOf(
     dequeRemoveFirst,
     DoneResult,
     Empty,
-    float64Abs,
-    float64Acos,
-    float64Asin,
-    float64Atan,
-    float64Atan2,
-    float64Ceil,
-    float64Cos,
-    float64Cosh,
-    float64Exp,
     float64Expm1,
-    float64Floor,
     float64Log,
-    float64Log10,
     float64Log1p,
     float64Min,
     float64Max,
     float64Near,
-    float64Round,
     float64Sign,
-    float64Sin,
-    float64Sinh,
-    float64Sqrt,
-    float64Tan,
-    float64Tanh,
     Float64E,
     Float64Pi,
     float64ToInt,
@@ -1067,10 +1025,6 @@ private val connectedReferences = listOf(
     float64ToString,
     GetConsole,
     ignore,
-    int32Max,
-    int32Min,
-    int64Max,
-    int64Min,
     IntToFloat64,
     IntToInt64,
     intToString,
@@ -1096,7 +1050,6 @@ private val connectedReferences = listOf(
     listedJoin,
     listedLength,
     listedMap,
-    listedMapDropping,
     listedReduce,
     listedReduceFrom,
     listedSlice,
@@ -1150,7 +1103,6 @@ private val connectedReferences = listOf(
     stringToInt,
     stringToInt64,
     stringToString,
-    stringIsEmpty,
     StringBuilderConstructor,
     stringBuilderAppend,
     stringBuilderAppendBetween,
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -2090,9 +2090,12 @@ class RustTranslator(
         ).wrapArcType()
     }
 
-    private fun translateGarbage(garbage: TmpL.Garbage): Rust.Call {
-        val pos = garbage.pos
-        return "panic!".toId(pos).call(listOf(Rust.StringLiteral(pos, "Garbage")))
+    private fun translateGarbage(garbage: TmpL.Garbage): Rust.Expr = run {
+        translateUnsupported(garbage.pos, garbage.toString())
+    }
+
+    private fun translateGarbageStatement(garbage: TmpL.GarbageStatement): Rust.ExprStatement = run {
+        translateUnsupportedStatement(garbage)
     }
 
     private fun translateGetProperty(expression: TmpL.GetProperty, avoidClone: Boolean): Rust.Expr {
@@ -2615,7 +2618,7 @@ class RustTranslator(
         typePub: Rust.VisibilityPub? = null,
     ): List<Rust.Item> {
         return when (member) {
-            is TmpL.GarbageStatement -> return listOf()
+            is TmpL.GarbageStatement -> translateGarbageStatement(member).toItem() // won't compile but that's ok
             // So far only bother with returnType forwarding for instance methods. Will we need more later?
             is TmpL.Getter -> translateGetter(member, block = block, forTrait = forTrait, returnType = returnType)
             is TmpL.Setter -> translateSetter(member, block = block, forTrait = forTrait) // don't expect return type
@@ -2906,7 +2909,7 @@ class RustTranslator(
                 val ref = statement.left
                 val subject = when (val subj = ref.subject) {
                     is TmpL.Expression -> translateExpression(subj, avoidClone = true)
-                    is TmpL.TypeName -> translateTypeName(subj)
+                    is TmpL.TypeName -> translateTypeName(subj) // wrong but also shouldn't happen
                 }
                 val setter = "set_${translatePropertyId(ref.property)}"
                 subject.methodCall(setter, listOf(value))
@@ -2934,13 +2937,13 @@ class RustTranslator(
         try {
             return when (statement) {
                 is TmpL.Assignment -> return translateAssignment(statement)
-                is TmpL.BoilerplateCodeFoldEnd -> return listOf()
-                is TmpL.BoilerplateCodeFoldStart -> return listOf()
+                is TmpL.BoilerplateCodeFoldEnd -> translateUnsupportedStatement(statement)
+                is TmpL.BoilerplateCodeFoldStart -> translateUnsupportedStatement(statement)
                 is TmpL.BreakStatement -> translateBreakStatement(statement)
                 is TmpL.ContinueStatement -> translateContinueStatement(statement)
-                is TmpL.EmbeddedComment -> return listOf()
+                is TmpL.EmbeddedComment -> translateUnsupportedStatement(statement)
                 is TmpL.ExpressionStatement -> translateExpressionStatement(statement)
-                is TmpL.GarbageStatement -> Rust.ExprStatement(statement.pos, translateGarbage(statement))
+                is TmpL.GarbageStatement -> translateGarbageStatement(statement)
                 is TmpL.HandlerScope -> error("handled elsewhere")
                 is TmpL.LocalDeclaration -> return translateModuleOrLocalDeclaration(statement)
                 is TmpL.LocalFunctionDeclaration -> error("handled elsewhere")
@@ -3069,6 +3072,7 @@ class RustTranslator(
         return when (type) {
             is TmpL.FunctionType -> translateFunctionType(type)
             is TmpL.TypeIntersection -> when {
+                // This branch is never expected, so some this translation is untested.
                 type.types.isEmpty() -> ANY_NAME.toId(pos)
                 type.types.size == 1 -> translateType(type.types.first(), inExpr = inExpr, isFlex = isFlex)
                 else -> Rust.ImplTraitType(
@@ -3335,6 +3339,18 @@ class RustTranslator(
             ),
             args = listOf(),
         )
+    }
+
+    private fun translateUnsupported(pos: Position, diagnostic: String): Rust.Expr = run {
+        "panic!".toId(pos).call(listOf(Rust.StringLiteral(pos, "Unsupported: $diagnostic")))
+    }
+
+    private fun translateUnsupported(tree: TmpL.Tree): Rust.Expr = run {
+        translateUnsupported(tree.pos, tree.toString())
+    }
+
+    private fun translateUnsupportedStatement(statement: TmpL.Statement): Rust.ExprStatement = run {
+        Rust.ExprStatement(statement.pos, translateUnsupported(statement))
     }
 
     private fun translateValueReference(expression: TmpL.ValueReference): Rust.Expr {
--- a/be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/listed.rs
+++ b/be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/listed.rs
@@ -143,22 +143,6 @@ mod listed {
         Arc::new(result)
     }
 
-    pub fn map_dropping<T, O>(
-        listed: &dyn ListedTrait<T>,
-        transform: &dyn Fn(T) -> Option<O>,
-    ) -> List<O>
-    where
-        T: Clone + Sync + Send,
-    {
-        let mut result = vec![];
-        for i in 0..listed.len() {
-            if let Some(value) = transform(listed.get(i)) {
-                result.push(value);
-            }
-        }
-        Arc::new(result)
-    }
-
     pub fn reduce<T>(listed: &dyn ListedTrait<T>, accumulate: &dyn Fn(T, T) -> T) -> T
     where
         T: Clone + Sync + Send,


mikesamuel:
Could you add to functional-test-suite/src/commonMain/resources/types/float/ops/ ?

That'd make sure the corner cases are exercised for all backends.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Float64.cs
```diff
@@ -73,7 +73,7 @@ public static bool Near(
             var rel = relTol ?? 1e-9;
             var abs = absTol ?? 0.0;
             var margin = Math.Max(Math.Max(Math.Abs(x), Math.Abs(y)) * rel, abs);
-            return Math.Abs(x - y) < margin;
+            return Math.Abs(x - y) <= margin;
```
tjpalmer, line 76: This does match [Python's `math.isclose`](https://docs.python.org/3/library/math.html#math.isclose) better. I'll plan to see if everything else also gets updated in this pr.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Float64.cs
```diff
@@ -109,7 +109,7 @@ public static double ToFloat64(this string s)
                 ? double.PositiveInfinity
                 : trimmed == "-Infinity"
                     ? double.NegativeInfinity
-                    : double.Parse(s);
+                    : double.Parse(trimmed);
```
tjpalmer, line 112: [In Remarks here](https://learn.microsoft.com/en-us/dotnet/api/system.double.parse?view=net-10.0#system-double-parse(system-string)), it says that `double.Parse` allows for surrounding whitespace, so it doesn't matter much whether `s` or `trimmed` goes in. Maybe worth a comment either way, and maybe `trimmed` is less surprising. So probably an ok change to make, even if `s` also works.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/StringUtil.cs
```diff
@@ -62,7 +62,7 @@ public static int CountBetween(string s, int left, int right)
             for (int i = Math.Max(0, left); i < limit; ++i)
             {
                 count += 1;
-                if (i + 1 < right)
+                if (i + 1 < limit)
```
tjpalmer, line 65: Because the loop prevents `i` from going past `limit`, either works fine here, so again, a needless change but also not a breaking one. And maybe `limit` is less surprising to some people.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/StringUtil.cs
```diff
@@ -195,7 +195,7 @@ public static int RequireNoStringIndex(int i)
         {
             if (i < 0)
             {
-                return -1;
+                return i;
             }
```
tjpalmer, line 199: I personally feel a lot better always using `-1` for no string index rather than arbitrary negative values. Is there a good reason for this change that I'm missing? If not, I'd like to revert this.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/float.js
```diff
@@ -19,7 +19,7 @@ export const float64Near = (x, y, relTol, absTol) => {
     absTol = 0;
   }
   const margin = Math.max(Math.max(Math.abs(x), Math.abs(y)) * relTol, absTol);
-  return Math.abs(x - y) < margin;
+  return Math.abs(x - y) <= margin;
```
tjpalmer, line 22: So we've updated cs and js so far.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/listed.js
```diff
@@ -238,7 +238,7 @@ export const listBuilderReverse = (ls) => {
  * @param {T} newValue
  */
 export const listBuilderSet = (ls, i, newValue) => {
-  if (0 <= i && i <= ls.length) {
+  if (0 <= i && i < ls.length) {
     ls[i] = newValue;
   }
```
tjpalmer, line 243: I guess the old version allowed for expanding the array on `i == ls.length`, which isn't what I read in the docs:

> Panics on index out of bounds.

So I agree this is a good fix.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/string.js
```diff
@@ -159,7 +159,7 @@ export const stringHasAtLeast = (s, begin, end, minCount) => {
  * @returns {number}
  */
 export const stringNext = (s, i) => {
-    let iNext = Math.min(s.length, i);
+    let iNext = Math.min(s.length, Math.max(0, i));
```
tjpalmer, line 162: Because this is our spec:

```ts
  public next(index: StringIndex): StringIndex;
```

It's illegal to pass in negative numbers, and we currently have no specific behavior for illegal negative values. Old behavior would return the same invalid index passed in. New behavior treats it as index 0. I think the old behavior is better, but maybe a comment would be good.

And maybe better yet would be an exception on bad value, but that might slow down the common case of valid parameter values.

Overall, maybe add a comment, but I'd like to revert this change.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/string.js
```diff
@@ -173,7 +173,7 @@ export const stringNext = (s, i) => {
  * @returns {number}
  */
 export const stringPrev = (s, i) => {
-    let iPrev = Math.min(s.length, i);
+    let iPrev = Math.min(s.length, Math.max(0, i));
```
tjpalmer, line 176: I haven't reviewed carefully here, but I suspect the old behavior is better again, as for stringNext.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaLocalRemover.kt
```diff
@@ -479,7 +493,7 @@ class LuaLocalRemover {
                     stmt.exprs.exprs.map(::scan),
                 ),
             )
-            stmt.targets.targets.forEach(::define)
+            stmt.targets.targets.forEach { define(it, asEnvField = false) }
```
tjpalmer, line 496: I can't easily just review this by looking through code. I'll rely on whatever analysis and testing you've done. Could you report that here?

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaSupportNetwork.kt
```diff
@@ -91,7 +91,7 @@ internal fun operatorToName(
     BuiltinOperatorId.Print -> "print"
     BuiltinOperatorId.StrCat -> "concat"
     BuiltinOperatorId.Listify -> "listof"
-    BuiltinOperatorId.Async -> "TODO" // TODO
+    BuiltinOperatorId.Async -> "async"
```
tjpalmer, line 94: Presuming you've done testing for the various additions here.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaSupportNetwork.kt
```diff
@@ -289,6 +289,92 @@ internal object LuaSupportNetwork : SupportNetwork {
...
+                    builtin.builtinOperatorId,
+                )
+                BuiltinOperatorId.GtStrStr -> inlineBinaryOp(
+                    builtin.builtinOperatorId.toString(), BinaryOpEnum.Gt, LuaOperatorDefinition.Gt,
+                    builtin.builtinOperatorId,
+                )
+                BuiltinOperatorId.GeStrStr -> inlineBinaryOp(
+                    builtin.builtinOperatorId.toString(), BinaryOpEnum.GtEq, LuaOperatorDefinition.Ge,
+                    builtin.builtinOperatorId,
+                )
```
tjpalmer, line 377: Inlining this seems reasonable. Are funtests good enough to test these changes?

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
```diff
@@ -93,7 +93,7 @@ function temper.codepoint_fallback(s, i)
         b1 = string.byte(s, i + 1)
         -- TODO: these checks should really be (b1 & 192) == 128
         if b1 == nil or b1 >= 192 then return 0xFFFD; end
-        return (b0 - 192)*64 + b1
+        return (b0 - 192)*64 + b1%64
```
tjpalmer, line 96: I haven't looked closely at this change.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
```diff
@@ -1985,4 +2026,17 @@ do
...
+    local head = 1
+    for i = 1, #list do
+        local ok, val = pcall(f, list[i])
+        if ok then
+            ret[head] = val
+            head = head + 1
+        end
+    end
+    return ret
+end
```
tjpalmer, line 2040: Going to trust your analysis and testing on these lua changes.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -216,7 +216,7 @@ def generic_lt_eq(left: C, right: C) -> bool:
 
 
 def generic_lt(left: C, right: C) -> bool:
-    "Checks if two left <=right, caring about the sign of zeros of floats."
+    "Checks if left < right, caring about the sign of zeros of floats."
     if isinstance(left, float) and isinstance(right, float):
         return float_lt(left, right)
     return left < right
```
tjpalmer, line 222: Doc changes do match the code better in the above cases.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -251,9 +251,12 @@ def arith_int_mod(dividend: int, divisor: int) -> int:
     arith_int_mod(5, 3) == 2
     arith_int_mod(-5, -3) == -2
     arith_int_mod(5, -3) == 2
-    arith_int_mod(-5, -3) == -2
+    arith_int_mod(-5, 3) == -2
```
tjpalmer, line 254: I guess it was a repeat otherwise, and I haven't reviewed intended semantics, but the behavior matches:

```py
>>> t.arith_int_mod(-5, 3)
-2
>>> t.arith_int_mod(-5, -3)
-2
```

So looks like a good change.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -251,9 +251,12 @@ def arith_int_mod(dividend: int, divisor: int) -> int:
...
     """
-    return dividend - divisor * int(dividend / divisor)
+    q = dividend // divisor
+    if (dividend ^ divisor) < 0 and q * divisor != dividend:
+        q += 1
+    return dividend - divisor * q
```
tjpalmer, line 259: This change should handle large integers that fail in float div, so this is good. Technically could cache `q * divisor` in a local, but not vital at the moment.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -387,7 +390,7 @@ def __init__(self, capacity: int):
 
     def __bool__(self) -> bool:
         "Test if any bit is set."
-        return bool(rb"\0" in self._bytearray)
+        return any(b != 0 for b in self._bytearray)
```
tjpalmer, line 393: I expect the original is correct and much faster, so unless I've missed something, we should revert this.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -496,7 +499,10 @@ def int64_div(a: int, b: int) -> int:
     # Mostly concerned with b == -1, but maybe evil `a` snuck in from outside?
     if a <= -0x8000_0000_0000_0000 and b < 0:
         return int64_clamp(int(a / b))
-    return int(a / b)
+    r = a // b
+    if (a ^ b) < 0 and r * b != a:
+        r += 1
+    return r
```
tjpalmer, line 505: I haven't reviewed this, but if done correctly, I presume it handles large integers better.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -549,7 +555,7 @@ def float64_near(
     y: float,
     rel_tol: Optional[float] = Unset,
     abs_tol: Optional[float] = Unset,
-) -> float:
+) -> bool:
```
tjpalmer, line 558: Fixed return types look good.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
tjpalmer, line 567: And this clearly is using Python's `isclose`, so that's cs, js, and py so far.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -722,7 +728,7 @@ def require_string_index(i: int) -> int:
 def require_no_string_index(i: int) -> int:
     "Checked cast from i to NoStringIndex, a negative int"
     if i < 0:
-        return -1
+        return i
```
tjpalmer, line 731: Again, at the moment, I'd prefer reverting this to `-1`.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -381,7 +381,7 @@ internal fun Rust.GenericParam.toArg(): Rust.Id {
     return when (this) {
         is Rust.Id -> this
         is Rust.TypeParam -> id
-        is Rust.PathSegments -> TODO() // needed?
+        is Rust.PathSegments -> segments.last() as Rust.Id
```
tjpalmer, line 384: If we aren't testing this, maybe should revert to `TODO()`, unless we have a good analysis on why this is correct.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustOperator.kt
```diff
@@ -7,6 +7,8 @@ enum class RustOperator(
     val operatorDefinition: RustOperatorDefinition,
 ) {
     Assign("=", RustOperatorDefinition.Assignment),
+    LogicalAnd("&&", RustOperatorDefinition.LogicalAnd),
+    LogicalOr("||", RustOperatorDefinition.LogicalOr),
```
tjpalmer, line 11: I guess fine. We weren't using these, I guess?

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -146,13 +146,13 @@ private fun supportCodeByOperatorId(builtinOperatorId: BuiltinOperatorId?): Supp
         BuiltinOperatorId.NeIntInt -> neIntInt
         BuiltinOperatorId.NeStrStr -> neStrStr
         BuiltinOperatorId.NeGeneric -> neGeneric
-        BuiltinOperatorId.CmpFltFlt -> TODO()
-        BuiltinOperatorId.CmpIntInt -> TODO()
-        BuiltinOperatorId.CmpStrStr -> TODO()
+        BuiltinOperatorId.CmpFltFlt -> cmpFltFlt
+        BuiltinOperatorId.CmpIntInt -> CmpIntInt
+        BuiltinOperatorId.CmpStrStr -> CmpStrStrOrdering
```
tjpalmer, line 151: I guess we weren't/aren't actually using any of these? I guess meh on the changes if still not testing them.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -583,6 +583,53 @@ private object CmpGeneric : MethodCall(
...
+        returnType: Type2,
+        translator: RustTranslator,
+    ): Rust.Expr {
+        // Print takes a single string argument and outputs it.
+        val arg = arguments[0].expr as Rust.Expr
+        return "println!".toId(pos).call(listOf(Rust.StringLiteral(pos, "{}"), arg))
+    }
+}
+
+private val print: SupportCode = PrintCode
```
tjpalmer, line 631: I haven't reviewed if/how these are used at all.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -636,12 +683,29 @@ private class EqNeGeneric(
...
 private val eqStrStr = CmpStrStr("EqStrStr", BuiltinOperatorId.EqStrStr, RustOperator.Equals)
+private val float64Abs = MethodCall("Float64::abs", "abs")
+private val float64Acos = MethodCall("Float64::acos", "acos")
+private val float64Asin = MethodCall("Float64::asin", "asin")
+private val float64Atan = MethodCall("Float64::atan", "atan")
+private val float64Atan2 = MethodCall("Float64::atan2", "atan2")
+private val float64Ceil = MethodCall("Float64::ceil", "ceil")
+private val float64Cos = MethodCall("Float64::cos", "cos")
+private val float64Cosh = MethodCall("Float64::cosh", "cosh")
+private val float64Exp = MethodCall("Float64::exp", "exp")
```
tjpalmer: We were already passing float ops tests. Is the goal here to use explicitly scoped calls instead of methods? Given our existing control, I expect these changes are unneeded, and I think I'd prefer to revert.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -1056,6 +1150,7 @@ private val connectedReferences = listOf(
     stringToInt,
     stringToInt64,
     stringToString,
+    stringIsEmpty,
```
tjpalmer: Yeah, I think a lot of these additions are unneeded, and I think I'd prefer to revert them.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -2351,6 +2351,18 @@ class RustTranslator(
...
+        val operator = when (expr.op.tmpLOperator) {
+            TmpLOperator.Bang -> RustOperator.BoolComplement
+        }
+        return Rust.Operation(
+            expr.pos,
+            left = null,
+            operator = Rust.Operator(expr.op.pos, operator),
+            right = translateExpression(expr.operand),
+        )
+    }
```
tjpalmer, line 2367: If we were getting by, does this matter?

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -2588,8 +2600,8 @@ class RustTranslator(
         return when (member) {
             is TmpL.Getter -> translateGetterId(member)
             is TmpL.Setter -> translateSetterId(member)
-            is TmpL.NormalMethod -> translateNormalishMethodId(member)
-            else -> TODO()
+            is TmpL.Method -> translateNormalishMethodId(member)
+            else -> error("unexpected member type for translateMethodId: $member")
```
tjpalmer, line 2607: Is this tested somewhere? The new `error` call is likely helpful, but I feel less confident about the NormalMethod -> Method change.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -2603,7 +2615,7 @@ class RustTranslator(
         typePub: Rust.VisibilityPub? = null,
     ): List<Rust.Item> {
         return when (member) {
-            is TmpL.GarbageStatement -> TODO()
+            is TmpL.GarbageStatement -> return listOf()
```
tjpalmer: I'd likely prefer creating a string literal with garbage info instead of an empty list, so I'd prefer to revert unless we're including the garbage info.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -2892,7 +2904,10 @@ class RustTranslator(
         return when (statement.left.property) {
             is TmpL.ExternalPropertyId -> {
                 val ref = statement.left
-                val subject = translateExpression((ref.subject as? TmpL.Expression) ?: TODO(), avoidClone = true)
+                val subject = when (val subj = ref.subject) {
+                    is TmpL.Expression -> translateExpression(subj, avoidClone = true)
+                    is TmpL.TypeName -> translateTypeName(subj)
+                }
```
tjpalmer, line 2913: I'm not fresh enough on this to be sure of the effect. For type name, we'd presumably also need `::` instead of `.`? Do we have tests for any of this?

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -2919,27 +2934,33 @@ class RustTranslator(
...
             return when (statement) {
                 is TmpL.Assignment -> return translateAssignment(statement)
-                is TmpL.BoilerplateCodeFoldEnd -> TODO()
-                is TmpL.BoilerplateCodeFoldStart -> TODO()
+                is TmpL.BoilerplateCodeFoldEnd -> return listOf()
+                is TmpL.BoilerplateCodeFoldStart -> return listOf()
                 is TmpL.BreakStatement -> translateBreakStatement(statement)
                 is TmpL.ContinueStatement -> translateContinueStatement(statement)
-                is TmpL.EmbeddedComment -> TODO()
+                is TmpL.EmbeddedComment -> return listOf()
```
tjpalmer: Again, I'd like to revert to TODO unless we include more info in the output.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -2919,27 +2934,33 @@ class RustTranslator(
...
                 is TmpL.ExpressionStatement -> translateExpressionStatement(statement)
-                is TmpL.GarbageStatement -> TODO()
+                is TmpL.GarbageStatement -> Rust.ExprStatement(statement.pos, translateGarbage(statement))
                 is TmpL.HandlerScope -> error("handled elsewhere")
                 is TmpL.LocalDeclaration -> return translateModuleOrLocalDeclaration(statement)
-                is TmpL.LocalFunctionDeclaration -> TODO() // handled elsewhere
+                is TmpL.LocalFunctionDeclaration -> error("handled elsewhere")
```
tjpalmer, line 2949: Given gargage improvements, this error call and the garbage change above look good.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -3047,11 +3068,18 @@ class RustTranslator(
...
             is TmpL.FunctionType -> translateFunctionType(type)
-            is TmpL.TypeIntersection -> TODO()
+            is TmpL.TypeIntersection -> when {
+                type.types.isEmpty() -> ANY_NAME.toId(pos)
+                type.types.size == 1 -> translateType(type.types.first(), inExpr = inExpr, isFlex = isFlex)
+                else -> Rust.ImplTraitType(
+                    pos,
+                    bounds = type.types.map { translateType(it, inExpr = inExpr, isFlex = isFlex) as Rust.Path },
+                )
+            }
```
tjpalmer, line 3082: I don't have enough fresh in my mind to know if I like this, but maybe it's ok. At the same time, we aren't supposed to be making intersection types anymore, so maybe the TODO is better?

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -3047,11 +3068,18 @@ class RustTranslator(
...
             is TmpL.TypeUnion -> translateTypeUnion(type, isFlex = isFlex)
             is TmpL.GarbageType -> "()".toId(pos)
             is TmpL.NominalType -> translateTypeNominal(type, inExpr = inExpr, isFlex = isFlex)
-            is TmpL.BubbleType -> TODO()
+            is TmpL.BubbleType -> "()".toId(pos)
```
tjpalmer, line 3086: This probably works for the legacy behavior we've locked ourselves into for now.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/listed.rs
```diff
@@ -143,6 +143,22 @@ mod listed {
...
+        T: Clone + Sync + Send,
+    {
+        let mut result = vec![];
+        for i in 0..listed.len() {
+            if let Some(value) = transform(listed.get(i)) {
+                result.push(value);
+            }
+        }
+        Arc::new(result)
+    }
```
tjpalmer: Somehow we got by without this, and I haven't reviewed why this is needed now. Maybe good, though?

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -529,7 +529,31 @@ val TmpL.Type.withoutBubbleOrNull: TmpL.Type get() = this.withoutAtom {
...
+        if (hasDifferences) {
+            when (typesWithout.size) {
+                0 -> TmpL.NeverType(pos)
+                1 -> typesWithout.first().deepCopy()
+                else -> TmpL.TypeIntersection(pos, typesWithout.map { it.deepCopy() })
+            }
+        } else {
+            this
+        }
+    }
```
tjpalmer, line 556: Again, maybe ok, but I don't know that we actually need this anymore.

common/src/commonMain/kotlin/lang/temper/common/Memoized.kt
```diff
@@ -10,6 +12,7 @@ class Memoized<T>(
 ) {
     private var cached: List<T>? = null
 
+    @Synchronized
     operator fun invoke(): T = (
```
tjpalmer, line 16: I haven't analyzed if this matters. If it does, maybe it helps prevent random errors, but I just don't know.

compile/src/commonMain/kotlin/lang/temper/compile/fetch/GitCache.kt
```diff
@@ -116,7 +116,7 @@ internal fun makeWritable(root: Path) {
     var exception: Throwable? = null
     repeat(2) {
         runCatching {
-            Files.walk(root).forEach { it.toFile().setWritable(true, true) }
+            Files.walk(root).use { it.forEach { path -> path.toFile().setWritable(true, true) } }
```
tjpalmer, line 119: At the moment, I don't see why this is better.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
```diff
@@ -1121,13 +1121,17 @@ private fun checkForImportCycles(
...
         fun lookForCycle(possibleCycle: Cons.NotEmpty<PendingImport>): Cons.NotEmpty<PendingImport>? {
             val import = possibleCycle.head
             val exporter = import.exporter
             val exporterLoc = exporter.loc
             if (exporterLoc == loc) {
                 return possibleCycle
             }
+            if (!visited.add(exporterLoc)) {
+                return null
+            }
```
tjpalmer, line 1134: I would like analysis and/or testing for this change. I haven't analyzed myself at the moment.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/InlineToRepairUnrealizedGoals.kt
```diff
@@ -310,7 +310,6 @@ private class InlineToRepairUnrealizedGoals(
             val calleeTree = callToInline.definition
             val formalNameToArg = callToInline.formalNameToFunArg
             val thisBindings = callToInline.thisTypeBindings
-            callToInline.thisTypeBindings
```
tjpalmer, line 313: This looks like good cleanup at a glance.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -3234,7 +3234,7 @@ private fun mergeTypeArgs(from: StaticType?, into: StaticType): StaticType {
                 type.definition === WellKnownTypes.nullTypeDefinition -> null
                 else -> type
             }
-            is OrType -> type.members.firstNotNullOf { findNominal(it) }
+            is OrType -> type.members.firstNotNullOfOrNull { findNominal(it) }
```
tjpalmer, line 3237: I haven't analyzed to know if this is a good change or not.

lexer/src/commonMain/kotlin/lang/temper/lexer/Lexer.kt
```diff
@@ -1107,7 +1107,7 @@ class Lexer(
                         // If a '<' is adjacent to a '<' or '=', treat it as part of a
                         // larger shift or comparison operator, or close tag start marker.
                         val before = text[end - 1]
-                        val after = if (end + 1 < limit) { text[end - 1] } else { '\u0000' }
+                        val after = if (end + 1 < limit) { text[end + 1] } else { '\u0000' }
```
tjpalmer, line 1110: The old code looks sus, but I don't know why it wouldn't have triggered problems. No test changes from this?

result-helpers/src/commonMain/kotlin/lang/temper/result/junit/JUnitResultParser.kt
```diff
@@ -125,7 +125,7 @@ data class FailureInfo(
 ) {
     // Where luaunit uses type instead of message, so use that as a fallback,
     // and we do some conversion there but not full.
-    val cause: String get() = message ?: type!!
+    val cause: String get() = message ?: type ?: "unknown"
```
tjpalmer: If we were getting by, maybe the non-null assertion was correct, so maybe should revert?

tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Runner.kt
```diff
@@ -157,5 +157,6 @@ internal class SynchronizedPrintBuffer {
     @Synchronized
     fun append(str: String) { buffer.append(str) }
 
+    @Synchronized
     override fun toString(): String = "$buffer"
```
tjpalmer, line 161: Again, some analysis here would be good, which I haven't done right now.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
ShawSumma: This is that we don't want it to overwrite the value*64 before making the pattern x*64 + y%64

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
ShawSumma: This checks is there are only zeros, the other checks if there are any zero bytes

b"\x00\x77" would have been false, as it had a zero byte, but should be true, as it's not all zeros

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
ShawSumma: That's probably wise

lexer/src/commonMain/kotlin/lang/temper/lexer/Lexer.kt
ShawSumma: Nope, no changes, it's one of those "how did it ever work?!" things.

tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Runner.kt
ShawSumma: It makes sure the buffer is never read at the same the above append is happening, as that would be bad and give you a toString() that never actually existed.

tjpalmer: > It makes sure the buffer is never read at the same the above append is happening, as that would be bad and give you a toString() that never actually existed.

I guess the question is if it's ever used from multithreaded contexts. I still haven't looked, though.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaLocalRemover.kt
ShawSumma: It literally just sets `locals[target.target.id] = luaName` instead of `locals.remove(target.target.id)`, as that was causing issues

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaSupportNetwork.kt
ShawSumma: Yes, tho in the future an int-specialized version would be good.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaSupportNetwork.kt
ShawSumma: Yes, tho I'm adding the test Mike suggested

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
ShawSumma: It would be not that much better, as q changing makes it annoying, and python locals are quite slow compared to a multiply

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
ShawSumma: Yes, it fixes the same problem that some of the other div-related stuff did when returning 64 bit ints using float div.

tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Runner.kt
ShawSumma: For the time being it feels like it could be, as the `@Synchronized` on the other method is preexisting. If it weren't used multithreaded then both should be removed.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
```diff
@@ -1470,7 +1500,7 @@ function temper.string_hasindex(str, i)
 end
 
 function temper.string_indexof(str, target, i)
-    return string.find(str, target, i, true) or 0
+    return string.find(str, target, i, true) or -1
```
tjpalmer, line 1503: For lua, I think we're using 0 as no string index, since indices start at 1?

ShawSumma: I liked having it be -1 everywhere, tho maybe this needs more thought

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
tjpalmer: > This checks is there are only zeros, the other checks if there are any zero bytes

Ah. Logic fail on my part. And I can't invent a nicer alternative off the top of my head (although possibly some options if we were on numpy).

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
tjpalmer: I saw a revert to -1 elsewhere, but maybe missed this one?

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
tjpalmer: We should have better error messages, but I think setting external static properties already doesn't work in temper, and I'm not sure Mike intends to support it (and I think I can support that decision):

```ts
$ class A {
    public static var hi: Int = 5;
  }
interactive#0: A__0
$ A.hi = 6
1: A.hi = 6
   ⇧
[interactive#1:1+0-1]@G: Expected subtype of Function, but got Type
1: A.hi = 6
          ⇧
[interactive#1:1+7-8]@G: Expected subtype of Symbol, but got Int32
interactive#1: fail
$ A.hi
interactive#2: 5
```

tjpalmer: Meanwhile, beyond the temper errors, here's what that assignment translates to with the new code, which is super wrong:

```rs
std::any::TypeId::of::<Blah>().set_hi(6);
```

But maybe can leave it with a comment that it's known wrong.

tjpalmer: For reference, more correct would be `Blah::set_hi(6)`, but we also don't generate the setter, which is also ok.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -2339,8 +2339,8 @@ class RustTranslator(
         return Rust.Operator(
             op.pos,
             operator = when (op.tmpLOperator) {
-                TmpLOperator.AmpAmp -> TODO() // RustOperator.LogicalAnd
-                TmpLOperator.BarBar -> TODO() // RustOperator.LogicalOr
+                TmpLOperator.AmpAmp -> RustOperator.LogicalAnd
+                TmpLOperator.BarBar -> RustOperator.LogicalOr
```
tjpalmer, line 2346: I see here's where they were being ignored before. Looks like we don't actually translate to using these. Instead we assign bool values from if/else sequences. But probably good to implement these anyway, just in case, and even if not easy to test, the implementations look ok.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/listed.rs
tjpalmer: We got by without it because nothing has ever tested mapDropping. The above is also incorrect, because it's based on Option instead of Result. I'll back out the mapDropping changes from this branch, because it would increase scope. At least be-csharp also doesn't implement mapDropping. I haven't tested all backends yet.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
tjpalmer: Yeah, `print` also is a relic that we no longer produce. Not sure about the other compares, but as they're not being used, I'm feeling too busy to study and remove them from this pr. Just leaving this comment here, so people can know that they haven't ever really been tested.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/StringUtil.cs
mikesamuel: I agree with Tom on this.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustOperator.kt
mikesamuel: Temper internally turns && and || into nested conditionals.  That way our flow-sensitive analyses don't need to account for short-circuiting because it's implicit in the CFG already.

But I thought we recovered it in TmpL.  Dunno though.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
mikesamuel: Yeah.  Unless this is fixing a bug related to getters/setters or constructors, this seems fishy.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
mikesamuel: Embedded comments are just for the docgen use case.

common/src/commonMain/kotlin/lang/temper/common/Memoized.kt
mikesamuel: I think this class is unused and is redundant with Kotlin's *Lazy*.

compile/src/commonMain/kotlin/lang/temper/compile/fetch/GitCache.kt
mikesamuel: Yeah, we don't actually need to close a thing, unless I'm missing something.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
mikesamuel: Does this affect `==` with NoStringIndex

tjpalmer: > Does this affect `==` with NoStringIndex

I put an approve, but maybe this is worth testing somewhere sometime if we don't yet.

PR #375, opened 2026-03-14
Define missing shift and bitwise operators

mikesamuel:
Previously, Temper was missing some bit-twiddling operators:

- Shift operators: `<<`, `>>`, and `>>>`.
- Bitwise negation: `~`
- Exclusive or: `^`

In keeping with JavaScript for syntactic choices, this commit defines these operators with that punctuation.

Unlike JavaScript, our semantics are the same as C-like languages that do not have undefined behaviour on over/under flow, e.g. Java and C#.

This commit also defines a functional test that exercises these operators.

Some bugs were flushed out as a result:

- C#'s *System.Text.Convert.ToString* when used to format numbers as non-decimal, formats as if cast to unsigned: no `-` sign for negative numbers.
- be-js does not always render *Int64*s as bigints.
- Lua's number formatting ran into trouble with *math.mininteger* because of the 2's complement fixed point around negation with the smallest representible integer.

## Caveat

~~Lua, Rust, and C++ do not currently pass.
This commit will remain as part of a draft PR until they do.~~

### DIFF
--- a/be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
+++ b/be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
@@ -1196,7 +1196,7 @@ do
                 if last_digit_value ~= 0 then
                     last_digit_value = base - last_digit_value
                 end
-                local negatable = (num + last_digit_value) // base
+                local negatable = temper_int.int_div(num + last_digit_value, base)
                 local digit_index = last_digit_value + 1
                 local last_digit = string_sub(digits, digit_index, digit_index)
                 return "-" .. temper_tostring(-negatable, base) .. last_digit
@@ -1205,7 +1205,7 @@ do
             end
         elseif num >= base then
             local mod = num % base + 1
-            return temper_tostring(math_floor(num // base), base) .. string_sub(digits, mod, mod)
+            return temper_tostring(temper_int.int_div(num, base), base) .. string_sub(digits, mod, mod)
         else
             local mod = num % base + 1
             return string_sub(digits, mod, mod)
--- a/be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/intnew.lua
+++ b/be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/intnew.lua
@@ -68,6 +68,11 @@ local function int64_toint32unsafe(n)
     return result
 end
 
+-- Simple int floor div.
+function temper.int_div(a, b)
+    return a // b
+end
+
 -- Int32
 
 function temper.int32_add(a, b)
--- a/be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/intold.lua
+++ b/be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/intold.lua
@@ -169,6 +169,11 @@ local function trunc(x)
     return result
 end
 
+-- Simple int floor div.
+function temper.int_div(a, b)
+    return math.floor(a / b)
+end
+
 -- Int32
 
 function temper.int32_add(a, b)


be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpSupportNetwork.kt
```diff
@@ -789,7 +794,7 @@ private class WrapAsOptional(
 
 private val intToFloat64 = Cast(listOf("Int32::toFloat64"), StandardNames.keyDouble)
 private val intToInt64 = Cast(listOf("Int32::toInt64"), StandardNames.keyLong)
-private val intToString = StaticCall("Int32::toString", StandardNames.systemConvertToString)
+private val intToString = StaticCall("Int32::toString", StandardNames.temperCoreConvertToString)
```
mikesamuel, line 797: System.Text.Convert.ToString(int, int) formats -1 for radix 16 as "FFFFFFFF".
Our semantics is to use a sign.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
```diff
@@ -118,6 +118,103 @@ public static long DivSafe(this long a, long b)
             return a == long.MinValue && b == -1 ? long.MinValue : a / b;
         }
 
+        public static int UShr(this int a, int b)
+        {
+            b = b & 0x1F;
+            return b == 0 ? a : (a >> b) & (0x7FFF_FFFF >> (b - 1));
```
mikesamuel, line 124: b == 0 is a special case.  It can return a negative.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
```diff
@@ -118,6 +118,103 @@ public static long DivSafe(this long a, long b)
...
+                signed = true;
+                un = ((uint) ~n) + 1;
+            } else {
+                un = (uint) n;
+            }
+            uint uRadix = (uint) radix;
+            // do...while means for 0 we get "0"
+            do
+            {
+                int digitValue = (int) (un % uRadix);
```
mikesamuel, line 151: This outputs digits from least to most significant, so we reverse below.

be-js/src/commonMain/kotlin/lang/temper/be/js/js.out-grammar
```diff
@@ -368,8 +368,7 @@ StringLiteral.renderTo =
 `;
 
 NumericLiteral(value%`Number`);
-NumericLiteral.renderTo =
-    `tokenSink.emit(OutputToken("$value", OutputTokenType.NumericValue))`;
+NumericLiteral.renderTo = `tokenSink.emit(kotlinNumberToJsNumberToken(value))`;
```
mikesamuel, line 371: This allows putting the `n` suffix on BigInt values.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
```diff
@@ -1138,7 +1143,18 @@ do
 
     local function temper_tostring(num, base)
         if num < 0 then
-            return "-" .. temper_tostring(-num, base)
+            if num == math.mininteger then
```
mikesamuel, line 1194: This was infinitely recursing.  I need to test this more thoroughly, but I thought @ShawSumma  might have a better idea.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/intold.lua
```diff
@@ -27,6 +27,10 @@ if bitops == nil then
     bitops.band = bitwise(AND)
     bitops.bxor = bitwise(XOR)
     bitops.bor = bitwise(OR)
+    function bitops.bnot(a)
+        return bitops.bxor(a, -1)
+    end
```
mikesamuel, line 32: Need more here, especially shift operators which should probably be implemented in terms of mult and div by powers of 2.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -255,6 +255,54 @@ def arith_int_mod(dividend: int, divisor: int) -> int:
     """
     return dividend - divisor * int(dividend / divisor)
 
+def arith_bit_shl32(n: int, shift: int) -> int:
+    """
+    Performs an Int32 left shift.
+    """
+    # ctypes.c_int32 but that does not expose shift operators
```
mikesamuel, line 262: Apparently there is a shim for ctypes on PyPy but in any case, it doesn't actually let you invoke C shifting.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustOperator.kt
mikesamuel, line 1: Yeah.  I haven't done enough Rust bit twiddling.  TBD: debugging of this by getting the functional test running.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppSupportNetwork.kt
```diff
@@ -41,8 +41,13 @@ internal object CppSupportNetwork : SupportNetwork {
...
         BuiltinOperatorId.BooleanNegation -> Like.unary("!")
-        BuiltinOperatorId.BitwiseAnd -> Like.binary("&")
-        BuiltinOperatorId.BitwiseOr -> Like.binary("|")
+        BuiltinOperatorId.BitwiseAnd32, BuiltinOperatorId.BitwiseAnd64 -> Like.binary("&")
+        BuiltinOperatorId.BitwiseOr32, BuiltinOperatorId.BitwiseOr64 -> Like.binary("|")
+        BuiltinOperatorId.BitwiseXor32, BuiltinOperatorId.BitwiseXor64 -> Like.binary("^")
+        BuiltinOperatorId.BitwiseNegation32, BuiltinOperatorId.BitwiseNegation64 -> Like.unary("~")
+        BuiltinOperatorId.BitwiseShl32, BuiltinOperatorId.BitwiseShl64 -> Like.binary("<<")
+        BuiltinOperatorId.BitwiseShr32, BuiltinOperatorId.BitwiseShr64 -> Like.binary(">>")
+        BuiltinOperatorId.BitwiseShrUnsigned32, BuiltinOperatorId.BitwiseShrUnsigned64 -> null // TODO
```
mikesamuel: I need to double check where we are on numerics in be-cpp.  TBD

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -1090,14 +1133,197 @@ object BuiltinFuns {
...
             ) { a, b -> a or b },
             LongLongToLongFun(
                 "|",
-                BuiltinOperatorId.BitwiseOr,
+                BuiltinOperatorId.BitwiseOr64,
             ) { a, b -> a or b },
         ),
-    )
+    ).also {
+        helpSnippet(it, "Bitwise or operator", "builtin/|")
```
mikesamuel, line 1144: This makes the help available to the REPL help function.

be-cppv/src/commonMain/resources/lang/temper/be/cppv/temper-core/int.hpp
```diff
@@ -97,6 +98,12 @@ T neg(T i) {
   return sub(T(0), i);
 }
 
+template<typename T>
+T ushr(T i, int32_t j) {
+  constexpr int shift_size_mask = (sizeof(T) * CHAR_BIT) - 1;
+  return (T) (((std::make_unsigned_t<T>) i) >> (j & shift_size_mask));
+}
```
mikesamuel, line 106: Take this with a grain of salt.  I'm trying to do a no-UB right logical shift in the template style of the other numeric operators.

mikesamuel: I tweaked this to be C++-11 compatible.

functional-test-suite/src/commonMain/resources/types/int/shifty/shifty.temper.md
mikesamuel, line 1: I vetted the output by just writing Java versions of the helper functions in jshell and having it produce the output.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
```diff
@@ -1138,10 +1143,21 @@ do
...
+                local digit_index = last_digit_value + 1
+                local last_digit = string_sub(digits, digit_index, digit_index)
+                return "-" .. temper_tostring(-negatable, base) .. last_digit
+            else
+                return "-" .. temper_tostring(-num, base)
+            end
         elseif num >= base then
             local mod = num % base + 1
-            return temper_tostring(math_floor(num / base), base) .. string_sub(digits, mod, mod)
+            return temper_tostring(math_floor(num // base), base) .. string_sub(digits, mod, mod)
```
mikesamuel: `/` instead of `//` can lead to loss of precision.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppSupportNetwork.kt
```diff
@@ -41,8 +41,14 @@ internal object CppSupportNetwork : SupportNetwork {
...
-        BuiltinOperatorId.BitwiseAnd -> Like.binary("&")
-        BuiltinOperatorId.BitwiseOr -> Like.binary("|")
+        BuiltinOperatorId.BitwiseAnd32, BuiltinOperatorId.BitwiseAnd64 -> Like.binary("&")
+        BuiltinOperatorId.BitwiseOr32, BuiltinOperatorId.BitwiseOr64 -> Like.binary("|")
+        BuiltinOperatorId.BitwiseXor32, BuiltinOperatorId.BitwiseXor64 -> Like.binary("^")
+        BuiltinOperatorId.BitwiseNegation32, BuiltinOperatorId.BitwiseNegation64 -> Like.unary("~")
+        BuiltinOperatorId.BitwiseShl32, BuiltinOperatorId.BitwiseShl64 -> Like.binary("<<")
+        BuiltinOperatorId.BitwiseShr32, BuiltinOperatorId.BitwiseShr64 -> Like.binary(">>")
+        BuiltinOperatorId.BitwiseShrUnsigned32 -> Like.core("ushr32")
+        BuiltinOperatorId.BitwiseShrUnsigned64 -> Like.core("ushr64")
```
mikesamuel, line 51: Casts to unsigned, shifts, casts back

be-cppv/src/commonMain/resources/lang/temper/be/cppv/temper-core/int.hpp
```diff
@@ -97,6 +99,12 @@ T neg(T i) {
   return sub(T(0), i);
 }
 
+template<typename T>
+T ushr(T i, int32_t j) {
+  constexpr int shift_size_mask = (sizeof(T) * CHAR_BIT) - 1;
+  return (T) (((typename std::make_unsigned<T>::type) i) >> (j & shift_size_mask));
```
mikesamuel, line 105: make_unsigned_t is not C++ 11

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpOperator.kt
```diff
@@ -15,6 +15,11 @@ enum class CSharpOperator(
     LogicalAnd("&&", CSharpOperatorDefinition.LogicalAnd),
     InclusiveOr("|", CSharpOperatorDefinition.InclusiveOr),
     And("&", CSharpOperatorDefinition.And),
+    ExclusiveOr("^", CSharpOperatorDefinition.ExclusiveOr),
+    LeftShift("<<", CSharpOperatorDefinition.Shift),
+    RightShift(">>", CSharpOperatorDefinition.Shift),
```
mikesamuel, line 20: C# has a `>>>` operator but it's a recent addition.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpSupportNetwork.kt
```diff
@@ -798,7 +803,7 @@ private val int64ToFloat64 = StaticCall(listOf("Int64::toFloat64"), StandardName
 private val int64ToFloat64Unsafe = Cast(listOf("Int64::toFloat64Unsafe"), StandardNames.keyDouble)
 private val int64ToInt32 = StaticCall(listOf("Int64::toInt32"), StandardNames.temperCoreCoreToInt)
 private val int64ToInt32Unsafe = Cast(listOf("Int64::toInt32Unsafe"), StandardNames.keyInt)
-private val int64ToString = StaticCall("Int64::toString", StandardNames.systemConvertToString)
+private val int64ToString = StaticCall("Int64::toString", StandardNames.temperCoreConvertToString)
```
mikesamuel, line 806: https://learn.microsoft.com/en-us/dotnet/api/system.convert.tostring?view=net-10.0#code-try-4 shows that for non-decimal, it's always the unsigned bit sequence.

For short:

> ```cs
> //       Base 2 conversion:
> //          -32768    -->  0x1000000000000000
> ```

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
mikesamuel, line 124: You can emulate zero-extending shift by masking with positive 0x7FFF_FFFF... shifted by one less.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
```diff
@@ -118,6 +118,103 @@ public static long DivSafe(this long a, long b)
...
+        /** Like System.Convert.ToString but outputs the signed form. */
+        public static string ConvertToString(int n, int radix)
+        {
+            StringBuilder sb = new StringBuilder();
+            uint un;
+            bool signed = false;
+            if (n < 0)
+            {
+                signed = true;
+                un = ((uint) ~n) + 1;
```
mikesamuel, line 143: Same as 2's complement negation but does not require stepping out of unsigned value land.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
```diff
@@ -118,6 +118,103 @@ public static long DivSafe(this long a, long b)
...
+            } else {
+                un = (uint) n;
+            }
+            uint uRadix = (uint) radix;
+            // do...while means for 0 we get "0"
+            do
```
mikesamuel, line 149: Outputs digits in least significant first order.  We reverse it at the end.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaSupportNetwork.kt
```diff
@@ -108,16 +109,34 @@ class JavaSupportNetwork private constructor(private val javaLang: JavaLang) : S
             BuiltinOperatorId.ModIntInt, BuiltinOperatorId.ModIntInt64 -> modIntInt
             BuiltinOperatorId.ModIntIntSafe, BuiltinOperatorId.ModIntInt64Safe -> modIntIntSafe
             BuiltinOperatorId.ModFltFlt -> modDubDub
-            BuiltinOperatorId.BitwiseAnd -> bitwiseAnd
```
mikesamuel, line 111: We used Java as our standard for what bitwise operations on a non-UB, fixed-size ints, non-dynlang do, so it's no surprise this is a direct mapping to existing operators.

be-js/src/commonMain/kotlin/lang/temper/be/js/Js.kt
```diff
@@ -3613,7 +3613,7 @@ object Js {
         override fun renderTo(
             tokenSink: TokenSink,
         ) {
-            tokenSink.emit(OutputToken("$value", OutputTokenType.NumericValue))
+            tokenSink.emit(kotlinNumberToJsNumberToken(value))
```
mikesamuel, line 3616: Make sure that longs have the `n` suffix so they're represented using JS bigints.

be-js/src/commonMain/kotlin/lang/temper/be/js/JsSupportNetwork.kt
```diff
@@ -1590,18 +1590,58 @@ private val builtinOperatorIdToSupportCode = BuiltinOperatorId.entries.mapNotNul
...
+                Js.InfixExpression(pos, a, Js.Operator(pos.leftEdge, "^"), b)
+            }
+        }
+
+        BuiltinOperatorId.BitwiseShl32 -> runtimeLibraryBackedSupportCode(id) { pos, args ->
+            arity2(args)?.let { (a, b) ->
+                Js.InfixExpression(pos, a, Js.Operator(pos.leftEdge, "<<"), b)
+            }
+        }
+        BuiltinOperatorId.BitwiseShl64 -> runtimeLibraryReference(id) // BigInt << needs careful {under,over}flow
```
mikesamuel, line 1630: Our semantics are chosen so that programs in languages that you pick for things that need to be fast, can be fast.

So as an ALU only samples 5 or 6 lines from the shift amount, we have to emulate that to get the same semantics on bigint runtimes.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/tsconfig.json
```diff
@@ -8,19 +8,20 @@
...
-        "string.js"
+      "async.js",
+      "bitvector.js",
+      "check-type.js",
+      "core.js",
+      "date.js",
+      "deque.js",
+      "float.js",
+      "index.js",
+      "int.js",
```
mikesamuel, line 19: added

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/intnew.lua
```diff
@@ -14,6 +14,23 @@ function bitops.bxor(a, b)
     return a ~ b
 end
 
+function bitops.bnot(a)
+    return a ~ -1
```
mikesamuel, line 18: I wasn't sure how widely supported prefix `~` is.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/intnew.lua
```diff
@@ -14,6 +14,23 @@ function bitops.bxor(a, b)
...
+
+function bitops.lshift(a, b)
+    return a << b
+end
+
+function bitops.rshift(a, b)
+    return a >> b
+end
+
+function bitops.arshift(a, b)
```
mikesamuel, line 29: Not sure we need an unreliable arshift.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -549,8 +557,111 @@ private object GetConsole : RustInlineSupportCode("::getConsole") {
...
+                    ),
+                    Rust.Operator(bRPos, RustOperator.As),
+                    Rust.Id(bRPos, OutName("u32", null)),
+                ),
+            ),
+            pos = pos,
+        )
+    }
+}
+private val bitwiseShl32 = BitwiseShift("wrapping_shl", BIT_SIZE_I32, SHIFT_MASK_I32, BuiltinOperatorId.BitwiseShl32)
```
mikesamuel, line 604: Rust has shift operators, but they unreliably trap on over/under-flow.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -123,10 +123,20 @@ private val fLongLongToLong = Signature2(
     hasThisFormal = false,
 )
 
+private val fLongIntToLong = Signature2(
```
mikesamuel, line 126: The Int64 shifts shouldn't require the shift amount be same width since only a handful of its bits are actually fed into the ALU shift/rot unit.

be-cppv/src/commonMain/kotlin/lang/temper/be/cppv/RunCpp.kt
```diff
@@ -35,7 +35,8 @@ fun runCpp(
 }
 
 object GppCommand : ToolSpecifics {
-    override val cliNames = listOf("g++")
+    override val cliNames = listOf("g++-16", "g++-15", "g++-14", "g++")
```
ShawSumma, line 38: This is odd to me, but should be fine for now. I am not sure why we need g++ as opposed to clang++ for systems that don't have a g++. Maybe a way to regex cli names like `g\+\+-\d+` or something alike in the future.

mikesamuel: All I know is clang doesn't have the include path setup properly.  I haven't investigated why.  Just we seem to be using g++ flags here, so I figured I'd stick with g++.

tjpalmer: I agree we can refine this in the future. I'd still like to support cmake out of the box at some point also.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaSupportNetwork.kt
tjpalmer: > We used Java as our standard for what bitwise operations on a non-UB, fixed-size ints, non-dynlang do, so it's no surprise this is a direct mapping to existing operators.

And I'm still mostly fairly happy with biasing toward java semantics in many cases.

be-js/src/commonMain/kotlin/lang/temper/be/js/JsSupportNetwork.kt
tjpalmer: > Our semantics are chosen so that programs in languages that you pick for things that need to be fast, can be fast.

I'd still like to narrow performance gap between backends where possible, but we sometimes have to live with limits, I agree. Might be worth a general comment in docs that Int64 arithmetic can be especially slow on some backends. That is, using Int64 for large identifiers (and very simple arithmetic) is typically more reliable for speed than fancy Int64 math.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
tjpalmer: > `/` instead of `//` can lead to loss of precision.

That only works on newer luas, though. This will now fail to parse on older luas and luajit. Maybe should merge anyway and fix later?

Overall, we probably need to make a strategy for running funtests on multiple luas. And we don't really want different backends in terms of temper code gen. Like we might need separate test runtime choice from overall backend.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/intnew.lua
tjpalmer: > I wasn't sure how widely supported prefix `~` is.

I've typically manually adjusted my local testing for lua 5.1, luajit, lua 5.2, and lua 5.4 maybe? There's also 5.5 now, and I don't know if it breaks on anything.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
tjpalmer: > Rust has shift operators, but they unreliably trap on over/under-flow.

Yeah, overall I don't mind rust's strategy of having a variety of method names for different semantics.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
tjpalmer: > Like we might need separate test runtime choice from overall backend.

This also applies to cppv and perhaps others (different module styles for js? dotnet versions?), where we want to generate one set of code for a backend but test in different envs.

mikesamuel: Ugh.  Matrix testing is inevitable.  I didn't expect it to be Lua that pushed us there.

Yeah.  I am horribly ignorant of the different versions.

PR #376, opened 2026-03-15
feat: add std/io and std/keyboard modules across 6 backends

notactuallytreyanastasio:
## Summary

Adds `std/io` (sleep, readLine) and `std/keyboard` (nextKeypress) modules to Temper's standard library, wired across all six backends (JS, Python, Lua, Rust, Java, C#).

These are the first I/O primitives in Temper that enable interactive, real-time programs without host-language wrappers. The snake game now works on all 6 backends using arrow keys and WASD, no Enter key required.

## What changed since review

Addressed all of Mike's feedback:

1. **readLine simplified** — No longer does raw TTY / single-keypress reading. All 6 backends now do actual buffered line input. `readLine()` reads lines, period.
2. **Process exits removed** — Removed all `System.exit`, `process.exit`, `os.kill`, `Environment.Exit` calls from readLine. A library shouldn't terminate the host.
3. **Java completeExceptionally** — Fixed the swallowed exception bug.
4. **C# net8.0 reverted** — Back to net6.0. The version bump was incidental.
5. **New std/keyboard module** — Keypress input split into its own module (`std/keyboard`) with clean boundaries, separate from `std/io`. This addresses Mike's concern about scoping raw terminal access.
6. **Windows support** — 5/6 backends support Windows for keyboard input (Java via PowerShell, JS via Node setRawMode, Python via msvcrt, C# via Console.ReadKey, Rust via crossterm). Lua is Unix-only.

### std/io
```temper
@connected("stdSleep")
export let sleep(ms: Int): Promise<Empty> { panic() }

@connected("stdReadLine")
export let readLine(): Promise<String?> { panic() }
```

### std/keyboard
```temper
@connected("stdNextKeypress")
export let nextKeypress(): Promise<String?> { panic() }
```

Returns key names as strings: `"a"`, `"ArrowUp"`, `"Enter"`, `"Escape"`, etc. Returns null on EOF.

## Architecture

The snake game uses two concurrent async blocks sharing mutable direction state:

```temper
let {nextKeypress} = import("std/keyboard");
let {sleep} = import("std/io");

var inputDirection: Direction = new Right();

// Input loop — runs on its own thread/coroutine
async { ... =>
  while (true) {
    let key = await nextKeypress();
    // update inputDirection
  }
}

// Game tick — runs concurrently
async { ... =>
  while (game.status is Playing) {
    game = changeDirection(game, inputDirection);
    game = tick(game);
    console.log(render(game));
    await sleep(200);
  }
}
```

`readLine()` stays library-safe (buffered, blocking, no terminal manipulation). `nextKeypress()` is the explicit opt-in for "I own the terminal" programs.

## Per-Backend keyboard implementations

| Backend | Unix | Windows |
|---------|------|---------|
| Java | stty raw + escape sequence parsing | PowerShell `[Console]::ReadKey($true)` |
| JS | `process.stdin.setRawMode(true)` | Same (Node handles it) |
| Python | `tty.setraw()` + `termios` | `msvcrt.getwch()` |
| Lua | `stty raw` via cooperative scheduler | Not supported (returns null) |
| C# | `Console.ReadKey(true)` | Same (cross-platform) |
| Rust | `crossterm` crate | Same (cross-platform) |

## Test plan

- [ ] `controlFlowIoSleep` functional test passes on all 6 backends
- [ ] `JsRunFileLayoutTest.someModules` passes (snapshots updated for io.js/keyboard.js)
- [ ] `ReplTest.translateToLua` passes (regex updated for scheduler call)
- [ ] Snake game plays with arrow keys + WASD on JS, Python, Lua, Rust, Java, C#
- [ ] Snake moves continuously on timer while waiting for keypress input

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -1924,10 +1924,55 @@ public static void runAsync(Supplier<Generator<Optional<? super Object>>> genera
...
+        });
+        return future;
+    }
+
+    public static java.util.concurrent.CompletableFuture<String> stdReadLine() {
+        java.util.concurrent.CompletableFuture<String> future = new java.util.concurrent.CompletableFuture<>();
+        ForkJoinPool.commonPool().execute(() -> {
+            try {
+                if (System.console() != null) {
+                    // TTY: use stty raw mode for single keypress
```
mikesamuel: Might we need a different API for non-GUI keypress access.

    interface TerminalDirectAccess {
      nextKeypress(): Promise<Int32>;
    }

    let onTerminalKeypress(fn (TerminalDirectAccess): Generator<Empty>)

I wonder if https://github.com/mabe02/lanterna might help for non-GUI keypress access.

----

Doing this tty dodge is probably not Windows compatible, and if done in two threads, could lead to odd results.

`System.console().readLine()` might do a bit better specifically for the readLine use case.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -1924,10 +1924,55 @@ public static void runAsync(Supplier<Generator<Optional<? super Object>>> genera
...
+                    String[] cmd = {"/bin/sh", "-c", "stty raw -echo </dev/tty"};
+                    Runtime.getRuntime().exec(cmd).waitFor();
+                    int ch = System.in.read();
+                    String[] restore = {"/bin/sh", "-c", "stty sane </dev/tty"};
+                    Runtime.getRuntime().exec(restore).waitFor();
+                    if (ch == 3) { // Ctrl+C
+                        System.exit(1);
```
mikesamuel: Temper is a language for libraries which should leave exiting up to the program.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -1924,10 +1924,55 @@ public static void runAsync(Supplier<Generator<Optional<? super Object>>> genera
...
+                    }
+                    future.complete(String.valueOf((char) ch));
+                } else {
+                    java.io.BufferedReader reader = new java.io.BufferedReader(
+                        new java.io.InputStreamReader(System.in));
```
mikesamuel: This defaults to the system encoding which might not be the same as the terminal encoding.

It also doesn't release the file handle.

If there's no System.console(), then I think maybe we just need to default to assuming that there's no input to read.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -1924,10 +1924,55 @@ public static void runAsync(Supplier<Generator<Optional<? super Object>>> genera
...
+                    String line = reader.readLine();
+                    future.complete(line);
+                }
+            } catch (Exception e) {
+                future.complete(null);
```
mikesamuel: completeExceptionally(e)

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/io.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @returns {Promise<string | null>}
+ */
+export function stdReadLine() {
+  return new Promise(resolve => {
+    if (typeof process !== 'undefined' && process.stdin) {
+      process.stdin.resume();
+      process.stdin.setEncoding('utf8');
+      if (process.stdin.isTTY && process.stdin.setRawMode) {
+        process.stdin.setRawMode(true);
```
mikesamuel: Could this raw mode for the keypress access be done inside a try/finally that scopes the change to just the direct access operation.  Perhaps an API ike that discussed under Core.java would scope the effect.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaSupportNetwork.kt
```diff
@@ -91,7 +91,7 @@ internal fun operatorToName(
     BuiltinOperatorId.Print -> "print"
     BuiltinOperatorId.StrCat -> "concat"
     BuiltinOperatorId.Listify -> "listof"
-    BuiltinOperatorId.Async -> "TODO" // TODO
+    BuiltinOperatorId.Async -> "async_launch"
```
mikesamuel, line 102: @ShawSumma we forgot to actually implement async support in be-lua.  Could you give some eyes to this.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/TemperLang.Core.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net48;net6.0</TargetFrameworks>
+    <TargetFrameworks>net48;net8.0</TargetFrameworks>
```
mikesamuel: We need to be intentional about changes to supported versions.

@tjpalmer Do you have a sense of the difference between dotnet 6.0 and donet 8.0?

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -1351,3 +1351,51 @@ def _utf8_byte_of(code_point: int, byte_offset: int, n_bytes: int) -> int:
...
+
+def std_sleep(ms: int) -> 'Future[None]':
+    """Sleep for ms milliseconds, returning a Future."""
+    f: Future[None] = new_unbound_promise()
+
+    def _do_sleep():
+        _time.sleep(ms / 1000.0)
+        f.set_result(None)
+
+    _executor.submit(_do_sleep)
```
mikesamuel, line 1417: Ok.  I see this is run on the executor, so not blocking the main thread.

mikesamuel (commented):
Overall, impressive what you managed to put together quickly.

I think we need a bit more clarity on exactly what readLine does and shouldn't do.

io.temper.md says

> Read one line from standard input. Returns null on EOF.

But iiuc, your immediate use case needs interactive (before <kbd>ENTER</kbd>) keypress information.

Temper is a language for libraries, not whole programs, so we need to be careful to limit the scope of global changes like raw TTY access that might adversely affect the larger program.

Maybe splitting the problem into two APIs would allow for better scoping.
Some ideas under be-java around ways to do that.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
notactuallytreyanastasio: You're right on all counts. readLine should just read lines — the raw TTY / single-keypress behavior was a shortcut to get the snake game working and it leaked into what should be a straightforward API. This is for sure a boundaries oversight/mistake.

Plan:
1. **Simplify `readLine` across all 6 backends** to do actual line reading (no raw mode, no `stty`, no `setRawMode`). For Java specifically, use `System.console().readLine()` when a console is available, fall back to `BufferedReader` otherwise.
2. **Remove `System.exit()` / `process.exit()` / etc. from all backends** — a library shouldn't be terminating the host program.
3. **Defer the keypress API** — the `TerminalDirectAccess` / `onTerminalKeypress` idea makes sense as a separate, scoped API. That can be a follow-up where we design it properly with try/finally scoping for raw mode. The snake game can move to that API once it exists.

This means the snake game will be broken on this branch until we have a keypress API, but readLine will be correct and safe for library use. Do you think a keypress API is too ambitious? I would love to achieve it, because this entire mission is snake-driven. I _did_ consider another approach: what if to play snake, you had to write a lambda function that would determine the snake's next movement patterns in an array or something? I thought this was amusing, but would be tricky to hot-code-reload it for each backend etc. That said, I think we ignore that zany idea. What would concrete steps be towards implementing a keypress API that would be acceptable in your view?

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
notactuallytreyanastasio: Agreed — will remove all `System.exit` / `process.exit` / `os.kill` / `Environment.Exit` calls from readLine across every backend. A library has no business killing the host process.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
notactuallytreyanastasio: Noted — need to figure out the right encoding and resource handling here. Will flag this when we get to the readLine simplification so we don't just cargo-cult the same mistakes into the cleanup.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
notactuallytreyanastasio: Yep, will fix.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/io.js
notactuallytreyanastasio: Same plan as discussed on Core.java — readLine will drop raw mode entirely. When we build the keypress API, we'll scope raw mode with proper cleanup (e.g. restore in a finally or on callback completion).

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/TemperLang.Core.csproj
notactuallytreyanastasio: This was incidental — should still work with net6.0. I'll revert it and test the C# snake against 6.0. Happy to leave the 6.0 vs 8.0 discussion for a separate thread.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
notactuallytreyanastasio: Leaning toward lanterna for the Java keypress backend — it gives us cross-platform terminal handling (Windows via JNA) without shelling out to stty, and it handles arrow keys / special key sequences properly. Seems like the right foundation for a scoped `onTerminalKeypress` style API.

notactuallytreyanastasio: It requires not-java-8 soooo letting that go

PR #382, opened 2026-03-18
warn on over/under flow when parsing integer values.

mikesamuel:


log/src/commonMain/kotlin/lang/temper/log/MessageTemplate.kt
```diff
@@ -58,6 +58,12 @@ enum class MessageTemplate(
     // Lexical problems
     BadEmoji("Emoji not allowed", CompilationPhase.Lex),
     InvalidIdentifier("Not a valid identifier", CompilationPhase.Lex),
+    Int32OutOfBounds("Int32 value, %s, out of representable bounds", CompilationPhase.Lex),
+    Int64OutOfBounds("Int64 value, %s, out of representable bounds", CompilationPhase.Lex),
+    MaybePromoteInt32ToInt64(
+        "You can turn an Int32 literal into an Int64 by putting I64 at the end",
+        CompilationPhase.Lex,
```
tjpalmer, line 65: I think we were promoting `i64`, not `I64`.

mikesamuel: 🤦 Will fix.

PR #384, opened 2026-03-19
Upgrade C++ Backend 

ShawSumma:
This gets the C++11 backend to a state that is able to run significant chunks of temper code. I've added it to the default backends too


tjpalmer (commented):
Note that the latest cpp work was previously under cppv, which is also used in the external c++03 backend. The cppv backend also has support specifically for c++23 and other things also, so we probably should correlate plans before merging this or anything else on c++.

tjpalmer:
Closing because #433.

PR #386, opened 2026-03-19
Finish changing and testing near maybe

tjpalmer:
- I think this adds the backends that weren't already updated to `<=` for `Float64::near`
- Also adjust funtest to check for an edge case
- I don't have set up to test all backends properly at the moment, but this passes on lua, interp, js, and rust, at least, I think

### DIFF
--- a/functional-test-suite/src/commonMain/resources/types/float/ops/ops.temper.md
+++ b/functional-test-suite/src/commonMain/resources/types/float/ops/ops.temper.md
@@ -77,8 +77,11 @@ But also a bit of testing of `near` params.
 We have both `relTol` and `absTol` tolerance options, in that order. Either or
 both can be specified, although we try only one or the other here.
 
-    console.log("one.near(one + 0.1, absTol = 0.11): ${
-      one.near(one + 0.1, null, 0.11).toString()
+    console.log("one.near(one + 0.25, absTol = 0.25): ${
+      one.near(one + 0.25, null, 0.25).toString()
+    }");
+    console.log("one.near(one + 0.1 + 1e-9, absTol = 0.1): ${
+      one.near(one + 0.1 + 1e-9, null, 0.1).toString()
     }");
     console.log("one.near(one - 0.1, absTol = 0.11): ${
       one.near(one - 0.1, null, 0.11).toString()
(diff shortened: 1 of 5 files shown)


functional-test-suite/src/commonMain/resources/types/float/ops/ops.temper.md
```diff
@@ -77,8 +77,11 @@ But also a bit of testing of `near` params.
 We have both `relTol` and `absTol` tolerance options, in that order. Either or
 both can be specified, although we try only one or the other here.
 
-    console.log("one.near(one + 0.1, absTol = 0.11): ${
-      one.near(one + 0.1, null, 0.11).toString()
+    console.log("one.near(one + 0.25, absTol = 0.25): ${
+      one.near(one + 0.25, null, 0.25).toString()
+    }");
```
tjpalmer, line 82: I think I was getting floating point problems with `1.0 + 0.1`, but I think `1.0 + 0.25` is working for an equality check.

functional-test-suite/src/commonMain/resources/types/float/ops/ops.temper.md
```diff
@@ -77,8 +77,11 @@ But also a bit of testing of `near` params.
...
+    console.log("one.near(one + 0.1 + 1e-9, absTol = 0.1): ${
+      one.near(one + 0.1 + 1e-9, null, 0.1).toString()
     }");
```
tjpalmer, line 85: And I could probably get tighter, but this is mostly a slightly out of margin case. Maybe `1.0 + 0.1` would work consistently across backends without the extra margin if they're all reliably ieee754?

functional-test-suite/src/commonMain/resources/types/float/ops/ops.temper.md
```diff
@@ -127,7 +130,8 @@ one.sinh(): ✅
 (three * pi / four).tan(): ✅
 one.tanh(): ✅
 one.near(one + 0.1): false
-one.near(one + 0.1, absTol = 0.11): true
+one.near(one + 0.25, absTol = 0.25): true
+one.near(one + 0.1 + 1e-9, absTol = 0.1): false
```
mikesamuel, line 134: 🎉

PR #402, opened 2026-04-08
implicits: define StringBuilder.end

mikesamuel:
The new StringBuilder.end getter returns a StringIndex such that `this.toString().slice(String.begin, this.end)` is the string appended thus far as long as the string builder is not cleared in between.

This will allow keeping track of substrings in a string being built that might require additional processing, for example, translation string substitution.


be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
```diff
@@ -1373,6 +1373,20 @@ function temper.stringbuilder_clear(builder)
...
 end
 
+function temper.stringbuilder_end(builder)
+    if #builder == 0 then
+        return 1
+    end
+
+    local compacted = table.concat(builder, "")
+
+    -- Compacting the stringbuilder amortizes the cost of this operation
```
mikesamuel, line 1383: I use this strategy in Implicits and it works well wherever we're actually using lists.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -2945,7 +2945,19 @@ internal class Typer(
                                 it.definition as TypeShape
                             }.toSet()
                         }
-                        BecauseNoSuchMember(t.pos, memberSymbol, typeShapes)
+                        if (typeShapes.isEmpty() && thisType is InvalidType?) {
```
mikesamuel, line 2948: Mentioning MissingType was just annoying me.

be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
```diff
@@ -2329,6 +2329,43 @@ class JsBackendTest {
...
+        ),
+        want = """
+            |{
+            |  "js": {
+            |    "my-test-library": {
+            |      "src": {
+            |        "foo.js": {
+            |          "content":
+            |          ```
+            |          [""][0].length;
```
tjpalmer, line 2356: Thanks for this test. I wasn't sure I was imagining it correctly.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
tjpalmer: > I use this strategy in Implicits and it works well wherever we're actually using lists.

I'm sort of more of a fan of caching the length somewhere, since otherwise repeatedly calling `end` could keep allocating more memory, but I'm ok moving forward with this for now. And I think lua is flexible enough, although I haven't tried benchmarking to see if this slows anything down in any cases:

```lua
> local a = {"hi", "there"}; a.length = 3; print(#a)
2
```

Wrong `.length` there, by the way. I was just being sloppy. But correct `#a` of 2.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -129,6 +129,12 @@ def str_cat(*parts: str) -> str:
     return "".join(map(str, parts))
 
 
+def string_builder_end(string_builder: list[str]) -> int:
+    compacted = "".join(string_builder)
+    string_builder[:] = (compacted,)
+    return len(compacted)
```
tjpalmer, line 135: For python, harder to cache a length without a layer of nesting:

```py
>>> a = []
>>> a.end = 5
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'list' object has no attribute 'end'
```

frontend/src/commonMain/resources/implicits/Implicits.temper
```diff
@@ -778,6 +778,44 @@ export class StringBuilder {
...
+      else -> do {
+        // Compact and test
+        // Ideal if these things happen atomically.
+        let s = toString();
+        substrings.clear();
+        substrings.add(s);
+        s.end
+      }
+    }
+  }
```
tjpalmer, line 817: And we already have a layer of nesting here, so could potentially cache length info here, too, without much extra cost. But again, I'm ok just moving forward for now.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/TyperTest.kt
```diff
@@ -782,7 +782,7 @@ class TyperTest {
         |    (new B() as A).hi;
         """.trimMargin(),
         wantErrors = listOf(
-            "1+4-9: No member hi in MissingType!",
+            "1+4-6: No type for subject of .hi!",
```
tjpalmer, line 785: Yeah, I think I like the new text better.

functional-test-suite/src/commonMain/resources/types/string/indices/indices.temper.md
```diff
@@ -118,3 +118,22 @@ try it.
...
+      let str = stringBuilder.toString();
+      // Now I can use the string index with str.
+      console.log(str.slice(String.begin, afterComma));
+      console.log(str.slice(afterComma, str.end));
+    }
+
+```log
+Hello,
+ World!
+```
```
tjpalmer, line 139: Nice test.

tjpalmer (approved):

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
mikesamuel: I can't reply to the one about Lua for some reason, but unsure what `#a` is supposed to show.  The length is the sum of the lengths of elements.  I'm not using table length except to check whether it's empty or already consolidated.

I suppose we can rely on length caching since without orphaned surrogates, the sum of lengths is the length of the sum.

be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
mikesamuel: Yeah.  Building on the length expander simplifies the implementation, but it's good to have confirmation.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
tjpalmer: > I can't reply to the one about Lua for some reason, but unsure what `#a` is supposed to show.

I was just showing that the spare property doesn't change the array-like treatment. That's all. Not showing calculating of the string builder length.

On replying, github is awesome enough to show two versions of each comment for code reviews, and you can reply to one but not the other. Not sure if that's what you hit against or not. It often trips me up.

PR #403, opened 2026-04-09
Support empty rust enums; test mutual recursion

tjpalmer:
- Fix RustTranslator crash on sealed interfaces without subtypes
- Funtest mutual recursion
- The two topics are unrelated, but I hadn't put cross-backend testing on mutual recursion anywhere, so I snuck that in for future testing

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -296,7 +296,6 @@ class RustTranslator(
         // If we use a separate pre-naming stage for be-rust, this might matter less.
         decls@ for (topLevel in module.topLevels) {
             when (topLevel) {
-                is TmpL.TypeDeclaration -> {} // TODO
                 is TmpL.ModuleFunctionDeclaration -> {
                     decls[topLevel.name.name] = DeclInfo(topLevel, typeFrom = topLevel.sig)
                 }
@@ -974,7 +973,7 @@ class RustTranslator(
         ).toItem().also { moduleItems.add(it) }
         // Implement traits including AnyValue.
         val typeRef = id.makeTypeRef(generics)
-        implTraits(pos, decl, typeRef, generics) // , enumId)
+        implTraits(pos, decl, typeRef, generics)
         Rust.Call(
             pos,
             callee = "temper_core".toKeyId(pos).extendWith("impl_any_value_trait_for_interface!"),
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -60,6 +60,7 @@ import lang.temper.type2.ValueFormalKind
 import lang.temper.type2.hackMapOldStyleToNew
 import lang.temper.type2.withNullity
 import lang.temper.type2.withType
+import lang.temper.value.DeclTree
 import lang.temper.value.TBoolean
 import lang.temper.value.TClass
 import lang.temper.value.TClosureRecord
@@ -295,6 +296,7 @@ class RustTranslator(
         // If we use a separate pre-naming stage for be-rust, this might matter less.
         decls@ for (topLevel in module.topLevels) {
             when (topLevel) {
+                is TmpL.TypeDeclaration -> {} // TODO
                 is TmpL.ModuleFunctionDeclaration -> {
                     decls[topLevel.name.name] = DeclInfo(topLevel, typeFrom = topLevel.sig)
                 }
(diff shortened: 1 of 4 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1377,7 +1378,7 @@ class RustTranslator(
             pos,
             id = enumId,
             generics = generics,
-            items = decl.typeShape.sealedSubTypes!!.map { sub ->
+            items = (decl.typeShape.sealedSubTypes ?: listOf()).map { sub ->
```
tjpalmer, line 1382: And here's where we had a null pointer exception during temper build before this pr.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -932,6 +933,33 @@ class RustBackendTest {
                 |    }
                 |}
                 |temper_core::impl_any_value_trait!(Carrot, [Apple]);
+                |pub enum AllAloneEnum {}
```
tjpalmer: Empty enums are ok in rust.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -932,6 +933,33 @@ class RustBackendTest {
...
+                |        AllAloneTrait::clone_boxed( & ( * self.0))
+                |    }
+                |}
+                |temper_core::impl_any_value_trait_for_interface!(AllAlone);
+                |impl std::ops::Deref for AllAlone {
+                |    type Target = dyn AllAloneTrait;
+                |    fn deref(& self) -> & Self::Target {
+                |        & ( * self.0)
+                |    }
+                |}
```
tjpalmer: So much code we generate for one little line in Temper. Maybe I can make better macros someday.

functional-test-suite/src/commonMain/resources/functions/as-values/as-values.temper.md
```diff
@@ -36,3 +36,34 @@ Named functions can be stored in properties with a functional interface type.
...
+      }
+    }
+
+    let isOdd(i: Int): Boolean {
+      if (i == 0) {
+        false
+      } else {
+        isEven(i - 1)
+      }
+    }
```
tjpalmer, line 64: This would have failed in be-rust before #400, but I only added a rust backend test at that time.

Meanwhile, as I mention here, this isn't an ideal place to add this tangent test, but maybe also not the worst place.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -729,7 +731,7 @@ class RustTranslator(
                 type = typeRef.deepCopy(),
                 items = buildList {
                     // Some need as_enum.
-                    if (supShape.sealedSubTypes != null) {
+                    if (supDecl?.parts?.metadataSymbolMap?.contains(sealedTypeSymbol) == true) {
```
tjpalmer, line 734: Checking sealedSubTypes is unreliable to know when we need an `as_enum` method. This was causing rust compiler problems after code gen.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1745,6 +1747,38 @@ class RustBackendTest {
...
+                |pub struct AllAlone(std::sync::Arc<dyn AllAloneTrait>);
+                |impl AllAlone {
+                |    pub fn new(selfish: impl AllAloneTrait + 'static) -> AllAlone {
+                |        AllAlone(std::sync::Arc::new(selfish))
+                |    }
+                |}
+                |impl AllAloneTrait for AllAlone {
+                |    fn as_enum(& self) -> AllAloneEnum {
+                |        AllAloneTrait::as_enum( & ( * self.0))
+                |    }
```
tjpalmer, line 1765: So we get an as_enum for AllAloneTrait.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1745,6 +1747,38 @@ class RustBackendTest {
...
+                |    fn clone_boxed(& self) -> AllAlone {
+                |        AllAloneTrait::clone_boxed( & ( * self.0))
+                |    }
+                |}
+                |impl ITrait for AllAlone {
+                |    fn clone_boxed(& self) -> I {
+                |        ITrait::clone_boxed( & ( * self.0))
+                |    }
+                |}
```
tjpalmer, line 1774: But not for ITrait.

functional-test-suite/src/commonMain/resources/interfaces/property-members/property-members.temper.md
```diff
@@ -81,3 +81,10 @@ Also because it's in the ballpark, test generic sealed interfaces here.
 ```log
 Sealed thing is foo.
 ```
+
+Also a sealed interface with no subtypes. This is just here to see if backends
+break. There's no way to instantiate one.
+
+But do extend another non-sealed interface to ensure we also handle that.
+
+    sealed interface AllAlone extends I {}
```
tjpalmer, line 90: And this checks across backends that at least this doesn't break things.

mikesamuel (approved):

PR #404, opened 2026-04-10
optimization pass works with effects lists

mikesamuel:
This reworks the contextual autoescaper accumulator optimization micro-pass to work with effects.

It stubs out support for event effects; I plan to do that in conjunction with implementing the first use of those effects.

I was debugging the pass using the REPL and realized the `help` is less helpful than it could be for some things, so I opportunistically debugged help.

Now, `help(+)` is hacked to recognize that `+` there really means the name, not `+` in prefix position without an operator.  Maybe we could recognize that `(+)` is always a reference to the operator not a prefix or infix use, but this just hacks better support for `help` into the REPL specifically.


builtin/src/commonMain/kotlin/lang/temper/builtin/TypeFuns.kt
```diff
@@ -586,6 +587,10 @@ internal sealed class AsLikeFunction(
  * console.log(isAFoo(new Bar()).toString());  //!outputs "false"
  * ```
  */
+@HelpSnippet(
```
mikesamuel, line 590: I was investigating why `is` was acting weird when evaluated at compile time with type parameters.

Added the REPL help snippet connection.

builtin/src/commonMain/kotlin/lang/temper/builtin/TypeFuns.kt
mikesamuel, line 1: View with whitespace.  if body is just indented.

cli/src/main/kotlin/lang/temper/cli/repl/AdjustInputForRepl.kt
mikesamuel, line 1: The ReplTest makes clear what's going on here.

help(is) should just work.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -1118,8 +1118,14 @@ internal class Typer(
                             )
                         }
                     }
+                    inferredType is AndType && inferredType.members.all { it is FunctionType } ->
```
mikesamuel, line 1121: Don't try to convert `+`'s type to a new style type when staging `help(nym`+`)`.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -1118,6 +1118,8 @@ internal class Typer(
...
```
mikesamuel, line 1121: Don't try to convert the type of the argument in <tt>help(nym\`+\`)</tt>  to a new style type.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -3880,6 +3880,28 @@ class GenerateCodeStageTest {
...
     )
+
+    @Test
+    fun isAppliedToParameterizedType() = assertModuleAtStage(
+        stage = Stage.Run,
+        moduleResultNeeded = true,
+        input = """
+            |sealed interface I<T> {}
+            |class A<T> extends I<T> {}
+            |class B<T> extends I<T> {}
```
mikesamuel, line 3891: This didn't really expose anything, but I figure it's not bad to have a test of this form.

fundamentals/src/commonMain/kotlin/lang/temper/value/Result.kt
```diff
@@ -116,7 +116,24 @@ data class Value<T : Any>(
 sealed class Fail(val info: LogEntry? = null) : Result() {
     override fun destructure(structureSink: StructureSink) = structureSink.nil()
 
-    override fun toString() = OutToks.failWord.text
+    override fun toString(): String {
```
mikesamuel, line 119: Stringifying the message when stringiifying Fail results helps debugging.

build-user-docs/src/main/kotlin/lang/temper/docbuild/ExtractRequiredInlinedSnippets.kt
```diff
@@ -92,3 +95,9 @@ private fun combineMarkdownLinesDroppingIncidentalSpace(
         .render(MarkdownContent(lines.joinToString(separator = "\n")).root)
         .trimEnd()
 }
+
+// Markers like `<a name="builtin&#45;is" class="snippet-anchor-name"></a>` allow for
+// internal navigation but are not useful.
+private val snippetAnchorHtmlPattern = Regex(
+    """\s*<a [^>]*class="$SNIPPET_ANCHOR_CLASSNAME"[^>]*></a>\s*""",
+)
```
tjpalmer, line 103: I'm unsure at the moment why they aren't useful, but I'm probably just missing context.

cli/src/main/kotlin/lang/temper/cli/repl/AdjustInputForRepl.kt
```diff
@@ -0,0 +1,52 @@
...
+import lang.temper.lexer.LexicalDefinitions.Companion.quotedNamePrefix
+import lang.temper.lexer.Operator
+import lang.temper.lexer.OperatorType
+import lang.temper.log.LogSink
+import lang.temper.log.unknownPos
+
+internal fun adjustInputForRepl(commandText: String): String {
+    if (commandText.trimStart().startsWith(ReplHelpFn.NAME)) {
+        // Lex up to 5 tokens to see if we match ["help", "(", something, ")", ";"]
+        // with or without the semicolon at the end.
```
tjpalmer, line 13: I'm ok with the hack being isolated to `repl` package.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
tjpalmer: > Don't try to convert `+`'s type to a new style type when staging `help(nym`+`)`.

I'm unsure why this matters. Do we not plan to drop old-style types at some point?

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -3880,6 +3880,28 @@ class GenerateCodeStageTest {
...
+            |
+            |let f<T>(x: I<T>): Boolean {
+            |  x is A<T>
```
tjpalmer, line 3894: We presumably can't check this at runtime for all cases, but if we have statically constrained generics, then we only technically need to check `is A`. I don't remember all our past discussions on this.

tjpalmer (approved):
I have less to say about the optimization pass itself. I think I'm still not following all the details with how things are being used there.

On first class operators, while I'm not a fan of emphasizing symbolic operators, it still might be ok and wouldn't be novel to allow operators as expressions, even beyond `(+)`, such as in `hi(1, +, 2)` or other nullary contexts. And things like `help(+)` (as in this pr) could even make it easier to learn how to do operator overloading.

Overall, this pr is fine by me.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
mikesamuel: It was causing a crash on the ReplTest.

mikesamuel:
> I have less to say about the optimization pass itself. I think I'm still not following all the details with how things are being used there.
> 
> On first class operators, while I'm not a fan of emphasizing symbolic operators, it still might be ok and wouldn't be novel to allow operators as expressions, even beyond `(+)`, such as in `hi(1, +, 2)` or other nullary contexts. And things like `help(+)` (as in this pr) could even make it easier to learn how to do operator overloading.
> 
> Overall, this pr is fine by me.

Yeah. I think tweaking the OPP to understand something preceded by `(` or `,` and followed by `)` or `,` would do it.
Let me noodle on that.

PR #405, opened 2026-04-10
Report invalid local function call

tjpalmer:
- Basically, try to avoid calls to badly typed callees from being ignored
- Handle an issue discovered in a different repo
- Change type check reporting to post-order traversal handling


frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -442,7 +442,8 @@ internal class TypeChecker(
                             }
                         }
                     }
-                    if (!reported && calleeType != InvalidType) { // Invalid callee types already reported
+                    // Most invalid callee types already reported in more specific ways, but not all.
+                    if (!reported && (calleeType != InvalidType || callee.functionContained == null)) {
```
tjpalmer, line 459: If I report all InvalidType, I get the poor error message here along with more useful error messages in various cases in GenerateCodeStageTest. I don't know for sure what the right check is in every case, but this one seems to handle the edge case in the new test.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -3352,6 +3352,25 @@ class GenerateCodeStageTest {
...
 
+    @Test
+    fun invalidNonNullCheck() = assertModuleAtStage(
+        stage = Stage.Run,
+        input = """
+            |export let Act = fn (i: Int): Void;
+            |export let hi(i: Int, act: Act?): Void {
+            |  if (i != 0 && act != null) {
+            |    act(i);
+            |  }
```
tjpalmer, line 3363: The above doesn't actually handle non-null auto-casting properly because we don't support `&&` for that yet. And this is an edge case we weren't otherwise reporting errors on.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -3352,6 +3352,25 @@ class GenerateCodeStageTest {
...
+            |}
+        """.trimMargin(),
+        want = """
+            |{
+            |  run: "void: Void",
+            |  errors: ["Expected function type, but got Invalid!"],
```
tjpalmer, line 3369: Got the error for this case now.

mikesamuel:
Maybe the list of allowed errors at https://github.com/temperlang/temper/blob/de7bec89975891d2480da3bd02e1837f9382b9b0/functional-test-suite/src/commonMain/resources/semantics/broken/broken.temper.md?plain=1#L9-L17 might need adjusting.

tjpalmer:
> Maybe the list of allowed errors at ... might need adjusting.

Looking into this. It's getting the new error instead of the old "No declaration for doesntExist" now. On the pos. The old one is better. I'll try to see if I can get priority to the old error.

docs/for-users/temper-docs/docs/reference/builtins.md
```diff
@@ -2123,7 +2123,7 @@ Unlike in some other languages, curly brackets (`{...}`) are **required** around
 ```temper
 for (var i = 0; i < 3; ++i)
   console.log(i.toString()) // Curly brackets missing
-// ❌ No type for subject of .toString!
+// ❌ Expected function type, but got Invalid!, No type for subject of .toString!, Expected function type, but got Invalid!
```
tjpalmer, line 2126: Here are the details on this case when I run it in context giving more info:

```ts
1: for (var i = 0; i < 3; 
   ┗━┛
[test/test.temper:1+0-3]@G: Expected function type, but got Invalid
2: console.log(i.toString())
               ⇧
[test/test.temper:2+14-15]@G: No type for subject of .toString
2: console.log(i.toString())
               ⇧
[test/test.temper:2+14-15]@G: No declaration for i
2: console.log(i.toString())
               ┗━━━━━━━━┛
[test/test.temper:2+14-24]@G: Expected function type, but got Invalid
```

The extra "got Invalid" errors aren't ideal, but I don't think they're horrible either. And I presume we don't see "No declaration for i" in the docs case because it has the same pos as the error before it.

docs/for-users/temper-docs/docs/reference/builtins.md
```diff
@@ -2338,7 +2338,7 @@ You can't do that in Temper; always put `{`...`}` around the bodies.
 
 ```temper
 if (true) console.log("Runs"); else console.log("Does not run");
-// ❌ Expected a TopLevel here!
+// ❌ Expected a TopLevel here!, Expected function type, but got Invalid!
```
tjpalmer, line 2341: Details:

```ts
1: onsole.log("Runs"); else console.log("Does not run");
                       ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
[test/test.temper:1+31-63]@P: Expected a TopLevel here
1: if (true) console.log(
   ┗┛
[test/test.temper:1+0-2]@G: Expected function type, but got Invalid
```

docs/for-users/temper-docs/docs/reference/builtins.md
```diff
@@ -2834,7 +2834,7 @@ Unlike in some other languages, curly brackets (`{...}`) are **required** around
 var i = 0;
 while (i < 3)
   console.log((i++).toString()) // Curly brackets missing
-// ❌
+// ❌ Expected function type, but got Invalid!
```
tjpalmer, line 2837: Not sure why no messages before. I get all these if I try this in GenerateCodeStageTest:

```ts
1: while (i < 3)
          ⇧
[test/test.temper:1+7-8]@G: No declaration for i
1: while (i < 3)
          ┗━━━┛
[test/test.temper:1+7-12]@G: No callee matches inputs [Invalid, Int32] among [(Int32?, Int32?) -> Boolean, (Int64?, Int64?) -> Boolean, (Float64?, Float64?) -> Boolean, (String?, String?) -> Boolean, <cmpT extends AnyValue>(cmpT?, cmpT?) -> Boolean]
1: while (i < 3)
   ┗━━━┛
[test/test.temper:1+0-5]@G: Expected function type, but got Invalid
2: console.log((i++).toString()) // C
                ⇧
[test/test.temper:2+15-16]@G: No declaration for i
2: console.log((i++).toString()) // C
                ⇧
[test/test.temper:2+15-16]@G: Missing type info for i
2: console.log((i++).toString()) // C
                ⇧
[test/test.temper:2+15-16]@G: No declaration for i
2: console.log((i++).toString()) // 
                ⇧
[test/test.temper:2+15]@G: Cannot assign to Invalid from Int32
2: console.log((i++).toString()) // 
                ⇧
[test/test.temper:2+15]@G: Expected subtype of Invalid, but got Int32
2: console.log((i++).toString()) // Cur
                ┗━┛
[test/test.temper:2+15-18]@G: No type for subject of .toString
2: console.log((i++).toString()) // Curly bracket
                ┗━━━━━━━━━━━┛
[test/test.temper:2+15-28]@G: Expected function type, but got Invalid
```

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -96,27 +96,11 @@ internal class TypeChecker(
     private val voidReturnDeclNames = mutableSetOf<TemperName>()
 
     fun check(root: BlockTree) {
-        // First make sure we track which things actually can be void. Could store all return values, but only bother
-        // with those that can be void, since that's all that needs the extra checks.
-        if (module.outputType?.isVoidAllowing == true) {
-            voidReturnDeclNames.add(module.outputName!!)
-        }
+        // First make sure we track which things actually can be void.
+        trackVoidReturnDeclNames(root)
```
tjpalmer, line 100: I changed the type checking below to post-order, but that meant we didn't have the void return types figured out in time. So made a separate tree visit before type checking now to track those. One more straw on the camel's back for speed, though, I presume.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -96,27 +96,11 @@ internal class TypeChecker(
...
-                is CallTree -> checkCall(t)
-                is DeclTree -> checkDecl(t)
-                is EscTree -> checkEsc(t)
-                is FunTree -> checkFun(t)
-                is StayLeaf -> checkStay(t)
-                is LeftNameLeaf -> checkLeftName(t)
-                is RightNameLeaf -> checkRightName(t)
-                is ValueLeaf -> checkValue(t)
-            }
+            // Recurse first on the expectation that deeper nodes have more fundamental errors.
```
tjpalmer, line 103: And here's my justification for changing to post-order. The specific motivator was a case from semantics broken funtest where post-order means we see an error on a missing identifier before we see the "invalid type" on the outer attempted function call.

I think my justification is ok, though. I've considered being more aggressive in pruning error messages against any overlap. But presumably if we did that, multiple specific smaller messages for inner trees would be more useful perhaps that some later outer message that's likely a consequence of the inner messages.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -132,10 +116,43 @@ internal class TypeChecker(
...
+        }
+        TreeVisit.startingAt(root).forEachContinuing tree@{ tree ->
+            tree is FunTree || return@tree
+            val returnType = (tree.typeInferences?.type as? FunctionType)?.returnType ?: return@tree
+            if (returnType.isVoidAllowing) {
+                val returnDeclName = tree.parts?.returnDecl?.parts?.name ?: return@tree
+                voidReturnDeclNames.add(returnDeclName.content)
+            }
+        }.visitPreOrder()
+    }
```
tjpalmer, line 154: This basically just extracts the logic from what was the main recursive type checking. A separate pass here is technically cleaner also, even if it likely has more overhead.

And maybe could name this more generically and include other preprocessing in the future here, too.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -244,12 +261,8 @@ internal class TypeChecker(
         // for now as well as anything else that might slip through in the future.
         val returnType = (t.typeInferences?.type as? FunctionType)?.returnType
         val returnDecl = t.parts?.returnDecl
-        val returnDeclName = returnDecl?.parts?.name
-        val returnDeclType = returnDeclName?.typeInferences?.type
+        val returnDeclType = returnDecl?.parts?.name?.typeInferences?.type
         checkSubType(returnDecl?.pos, returnType, returnDeclType)
-        if (returnType?.isVoidAllowing == true && returnDeclName != null) {
-            voidReturnDeclNames.add(returnDeclName.content)
-        }
```
tjpalmer, line 252: Here's what got extracted into the pre-pass.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -132,10 +116,43 @@ internal class TypeChecker(
...
+    /**
+     * Track which things actually can be void. Could store all return values, but only bother
+     * with those that can be void, since that's all that needs the extra checks.
+     */
+    private fun trackVoidReturnDeclNames(root: Tree) {
+        if (module.outputType?.isVoidAllowing == true) {
+            voidReturnDeclNames.add(module.outputName!!)
+        }
+        TreeVisit.startingAt(root).forEachContinuing tree@{ tree ->
+            tree is FunTree || return@tree
```
tjpalmer, line 147: I considered taking out this check since the next check below is likely good enough, but this one does slightly less work, I expect, and most tree nodes won't be funtrees.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
mikesamuel, line 1: 👍

mikesamuel (approved):

PR #407, opened 2026-04-21
contextual autoescaping spot optimization recognizes effects

mikesamuel:
Effects are important for localization and CST building.

This implements TODO()s related to those.

Necessary for temperlang/secure-composition#17


tjpalmer (approved):
Tests could be nice to make this more concrete, but I'm fine if you're fine.

PR #409, opened 2026-04-27
Help strings in byte buffer slices

tjpalmer:
- Support external and possible future needs for easy string storage in struct slices
- Allow encoding into and decoding from byte buffer slices in single statements and/or expressions
- Run direct mvn tests for be-java temper-core
- The unit tests are mostly written by AI, but they look reasonable to me

### DIFF
--- a/be-java/build.gradle
+++ b/be-java/build.gradle
@@ -68,9 +68,9 @@ tasks.register("testJavaTemperCore", Exec) {
     workingDir = file("src/commonMain/resources/lang/temper/be/java/temper-core")
     if (org.gradle.internal.os.OperatingSystem.current().isWindows()) {
         // Effort needed for finding mvn on windows.
-        commandLine "cmd", "/c", "mvn", "test"
+        commandLine "cmd", "/c", "mvn", "-B", "test"
     } else {
-        commandLine "mvn", "test"
+        commandLine "mvn", "test", "-B"
     }
 }
 
--- a/be-java/build.gradle
+++ b/be-java/build.gradle
@@ -74,6 +74,6 @@ tasks.register("testJavaTemperCore", Exec) {
     }
 }
 
-tasks.named('test') {
-    dependsOn tasks.named('testJavaTemperCore')
+tasks.named("test") {
+    dependsOn tasks.named("testJavaTemperCore")
 }
--- a/be-java/build.gradle
+++ b/be-java/build.gradle
@@ -74,6 +74,6 @@ tasks.register("testJavaTemperCore", Exec) {
     }
 }
 
-tasks.named("test") {
+tasks.named("check") {
     dependsOn tasks.named("testJavaTemperCore")
 }
--- a/be-java/build.gradle
+++ b/be-java/build.gradle
@@ -63,3 +63,17 @@ kotlin {
         }
     }
 }
+
+tasks.register("testJavaTemperCore", Exec) {
+    workingDir = file("src/commonMain/resources/lang/temper/be/java/temper-core")
+    if (org.gradle.internal.os.OperatingSystem.current().isWindows()) {
+        // Effort needed for finding mvn on windows.
+        commandLine "cmd", "/c", "mvn", "test"
+    } else {
+        commandLine "mvn", "test"
+    }
+}
+
+tasks.named('test') {
+    dependsOn tasks.named('testJavaTemperCore')
+}
--- a/be-java/src/commonMain/resources/lang/temper/be/java/temper-core/pom.xml
+++ b/be-java/src/commonMain/resources/lang/temper/be/java/temper-core/pom.xml
@@ -24,6 +24,15 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
   </properties>
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <version>5.9.2</version>
+      <type>jar</type>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
   <build>
     <plugins>
       <plugin>
@@ -114,6 +123,11 @@
         <artifactId>maven-jar-plugin</artifactId>
         <version>3.3.0</version>
         <configuration/>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.0.0</version>
       </plugin>
     </plugins>
   </build>
--- a/be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
+++ b/be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
@@ -7,6 +7,7 @@ import java.nio.charset.CharacterCodingException;
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.CoderResult;
+import java.nio.charset.StandardCharsets;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -423,6 +424,7 @@ public final class Core {
     /**
      * Encode a string into a section of a ByteBuffer.
      * The caller is responsible for knowing that the slice exists.
+     * Default to UTF8 for null decoder.
      */
     public static String decodeFromSlice(
         ByteBuffer source,
@@ -430,6 +432,9 @@ public final class Core {
         int sourceLength,
         CharsetDecoder decoder
     ) throws CharacterCodingException {
+        if (decoder == null) {
+            decoder = StandardCharsets.UTF_8.newDecoder();
+        }
         // Limit slice.
         int oldLimit = source.limit();
         int oldPosition = source.position();
@@ -449,6 +454,7 @@ public final class Core {
     /**
      * Encode a string into a section of a ByteBuffer, returning the number of bytes written.
      * The caller is responsible for knowing that the slice exists.
+     * Default to UTF8 for null encoder.
      */
     public static int encodeIntoSlice(
         String s,
@@ -458,6 +464,9 @@ public final class Core {
         CharsetEncoder encoder,
         byte padByte
     ) {
+        if (encoder == null) {
+            encoder = StandardCharsets.UTF_8.newEncoder();
+        }
         // Limit slice.
         int oldLimit = target.limit();
         int oldPosition = target.position();
--- a/be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
+++ b/be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
@@ -1,6 +1,12 @@
 package temper.core;
 
 import java.io.UnsupportedEncodingException;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.CoderResult;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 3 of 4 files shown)


be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
...
+    /**
+     * Encode a string into a section of a ByteBuffer.
+     * The caller is responsible for knowing that the slice exists.
+     * Default to UTF8 for null decoder.
+     */
+    public static String decodeFromSlice(
+        ByteBuffer source,
+        int sourceStart,
+        int sourceLength,
+        CharsetDecoder decoder
```
tjpalmer, line 433: The charset decoder allows configuring how to handle incomplete chars at the end.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
...
+    ) throws CharacterCodingException {
+        if (decoder == null) {
+            decoder = StandardCharsets.UTF_8.newDecoder();
```
tjpalmer, line 436: We likely either want platform encoding or utf8 in most cases. Probably depends on how often we expect to want structs for interchange vs just in process.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
...
+    /**
+     * Encode a string into a section of a ByteBuffer, returning the number of bytes written.
+     * The caller is responsible for knowing that the slice exists.
+     * Default to UTF8 for null encoder.
+     */
+    public static int encodeIntoSlice(
+        String s,
+        ByteBuffer target,
+        int targetStart,
+        int targetLength,
```
tjpalmer, line 463: Seems like length is better than end here since this is about fitting field definitions in data structs.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
...
+        CharsetEncoder encoder,
+        byte padByte
```
tjpalmer, line 465: Could make overloads that default coder to null and pad to zero, but I haven't done that yet.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
...
+            // Encode.
+            encoder.reset();
+            CoderResult result = encoder.encode(CharBuffer.wrap(s), target, true);
+            encoder.flush(target);
+            // Pad.
+            int written = target.position() - targetStart;
+            while (target.hasRemaining()) {
+                target.put(padByte);
+            }
+            return written;
```
tjpalmer, line 485: Returning bytes written requires comparing with wanted to know if it all fit. Maybe different overloads to report different things might be useful?

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/test/java/temper/core/SliceCoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+    void decodeFromSliceSuccess() throws CharacterCodingException {
+        String input = "Hello World";
+        ByteBuffer buffer = ByteBuffer.wrap("###Hello World###".getBytes(StandardCharsets.UTF_8));
+        CharsetDecoder decoder = StandardCharsets.UTF_8.newDecoder();
+        // Decode only "Hello World" (offset 3, length 11)
+        String result = Core.decodeFromSlice(buffer, 3, 11, decoder);
+        assertEquals("Hello World", result);
+        assertEquals(0, buffer.position(), "Position should be restored");
+        assertEquals(buffer.capacity(), buffer.limit(), "Limit should be restored");
+    }
```
tjpalmer, line 26: Again, these tests are almost just copied and pasted from AI rather than typed up by me, but I think they're pretty good. And I requested some specific kinds of tests. And some I edited to use the default utf8 on null coder.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/test/java/temper/core/SliceCoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Test
+    void decodeFromSliceInvalidSequence() {
+        ByteBuffer buffer = ByteBuffer.wrap(new byte[]{(byte)0xFF}); // Invalid UTF-8
+        CharsetDecoder decoder = StandardCharsets.UTF_8.newDecoder();
+        assertThrows(CharacterCodingException.class, () -> Core.decodeFromSlice(buffer, 0, 1, decoder));
```
tjpalmer, line 32: I think it's ok to expose this exception for now. Maybe we can have wrappers that wrap it in the future.

And I almost pulled out the new static methods from Core into a new SliceCoder class or some such, but in the end, I wasn't convinced now was the time to start that. Maybe with extra wrappers, that would have motivated me more. And maybe breaking apart Core generally would be good sometime.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/test/java/temper/core/SliceCoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Test
+    void decodeFromSliceLatin1() throws CharacterCodingException {
+        // 0xA3 in Latin-1 is '£'. In UTF-8, 0xA3 is an invalid start byte.
+        ByteBuffer buffer = ByteBuffer.wrap(new byte[]{(byte)0xA3});
+        CharsetDecoder decoder = StandardCharsets.ISO_8859_1.newDecoder();
+        String result = Core.decodeFromSlice(buffer, 0, 1, decoder);
+        assertEquals("£", result, "Should decode correctly using Latin-1");
+    }
```
tjpalmer, line 42: At first, I only got utf8 testing, so I requested others.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/test/java/temper/core/SliceCoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Test
+    void encodeIntoSlicePartialMultibyteFit() {
+        // ✨ = 3 bytes (E2 9C A8)
+        // 😀 = 4 bytes (F0 9F 98 80)
+        String text = "✨😀";
+        ByteBuffer buffer = ByteBuffer.allocate(10);
+        // Slice of 5 bytes at offset 0
+        // Result: "✨" (3 bytes) fits, "😀" (4 bytes) fails, 2 bytes padding
```
tjpalmer, line 81: I also specifically requested a case for a partially fitting multibyte char.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/test/java/temper/core/SliceCoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+            .onUnmappableCharacter(CodingErrorAction.REPLACE)
+            .replaceWith(new byte[]{(byte)'?'});
+        int written = Core.encodeIntoSlice(text, buffer, 0, 5, encoder, (byte)'.');
+        // '£' becomes 0xA3 (1 byte)
+        // '✨' is unmappable, becomes '?' (1 byte)
+        assertEquals(2, written, "Should have written 2 bytes (one real, one replacement)");
+        byte[] data = buffer.array();
+        assertEquals((byte)0xA3, data[0], "Latin-1 encoding for £");
+        assertEquals((byte)'?', data[1], "Replacement char for unmappable emoji");
+        assertEquals((byte)'.', data[2], "Padding");
```
tjpalmer, line 115: I guess all this demonstrates why both "how many bytes got used" and "did the entire string fit" are both potentially interesting questions, depending on someone's use case. I don't know how to return both without allocation. Or maybe someone could pass in an object to receive the info that could be reused throughout a loop.

But what I have so far is likely good enough for my current needs.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/pom.xml
```diff
@@ -115,6 +124,11 @@
         <version>3.3.0</version>
         <configuration/>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.0.0</version>
+      </plugin>
```
tjpalmer, line 131: I used the same versions that we generate into temper-built java projects.

be-java/build.gradle
```diff
@@ -63,3 +63,17 @@ kotlin {
...
+        // Effort needed for finding mvn on windows.
+        commandLine "cmd", "/c", "mvn", "-B", "test"
+    } else {
+        commandLine "mvn", "test", "-B"
+    }
+}
+
+tasks.named("check") {
+    dependsOn tasks.named("testJavaTemperCore")
+}
```
tjpalmer, line 79: [Test results here.](https://github.com/temperlang/temper/actions/runs/25022567205/job/73286326249#step:10:6819)

```
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.144 s - in temper.core.SliceCoderTest
```

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
         }
     }
 
+    /**
+     * Encode a string into a section of a ByteBuffer.
```
mikesamuel: This comment matches that for encodeIntoSlice

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
...
+            encoder = StandardCharsets.UTF_8.newEncoder();
+        }
+        // Limit slice.
+        int oldLimit = target.limit();
+        int oldPosition = target.position();
+        target.position(targetStart);
+        target.limit(targetStart + targetLength);
+        try {
+            // Encode.
+            encoder.reset();
```
mikesamuel, line 477: This reset seems like it would preclude using the same encoder with the same buffer across two adjacent slices (adjacent in memory or adjacent in a byte streaming situation).

Assuming that's not a desired use case,
why take an encoder instead of a Charset if we can just create an encoder from a charset above?

That would avoid any callers potentially getting confused and reusing an encoder possibly across threads.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
...
+        CharsetDecoder decoder
+    ) throws CharacterCodingException {
+        if (decoder == null) {
+            decoder = StandardCharsets.UTF_8.newDecoder();
+        }
+        // Limit slice.
+        int oldLimit = source.limit();
+        int oldPosition = source.position();
+        source.position(sourceStart);
+        source.limit(sourceStart + sourceLength);
```
mikesamuel: Seems like position and limit should be inside the try.

As is, if source.limit fails with an exception, oldPosition will not be restored.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -414,6 +421,75 @@ private static void denySurrogate(int codePoint) {
...
+        int targetLength,
+        CharsetEncoder encoder,
+        byte padByte
+    ) {
+        if (encoder == null) {
+            encoder = StandardCharsets.UTF_8.newEncoder();
+        }
+        // Limit slice.
+        int oldLimit = target.limit();
+        int oldPosition = target.position();
```
mikesamuel, line 472: Same issue here.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/test/java/temper/core/SliceCoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Test
+    void encodeIntoSliceTruncation() {
+        ByteBuffer buffer = ByteBuffer.allocate(5);
+        String text = "Too Long String";
+        // Buffer limit of 2 bytes should stop the encoder
+        int written = Core.encodeIntoSlice(text, buffer, 0, 2, null, (byte)0);
+        assertEquals(2, written);
+        assertEquals(0, buffer.position());
```
mikesamuel, line 69: Should this assert that (byte)'T' and (byte)'o' are in buffer?

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
tjpalmer: > This comment matches that for encodeIntoSlice

Will fix.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
tjpalmer: > This reset seems like it would preclude using the same encoder with the same buffer across two adjacent slices (adjacent in memory or adjacent in a byte streaming situation).

I'm not sure this is an important use case, but maybe you're right that we should accommodate that. I was mostly resetting just to keep things clean. But if I'm putting off some responsibility to the caller, maybe I should put off more. Although it might require bother between repeated calls, which is how I'm using this right now. Maybe there could be yet another parameter to avoid reset???

> Assuming that's not a desired use case, why take an encoder instead of a Charset if we can just create an encoder from a charset above?

Mostly, I'm just trying to limit garbage creation. Could especially matter when loops. And we're allocating strings obviously anyway, but might still be a reasonable reduction in total number of objects created.

Another reason to pass in encoders/decoders is to allow things like this, which I'm *not* currently using:

```java
var encoder = charset.newEncoder();
encoder.onMalformedInput(...);
encoder.onUnmappableCharacter(...);
Core.encodeIntoSlice(...);
```

That's unavailable with just passing in a Charset.

> That would avoid any callers potentially getting confused and reusing an encoder possibly across threads.

That's a good point, but I guess I'm focusing on enabling rather than on just making it easy to get right. Which might be an argument for not requiring the reset every time. Although one more param could be annoying, might should consider it now or later (could use the current to forward to a different overload later).

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
tjpalmer: > Seems like position and limit should be inside the try.
> 
> As is, if source.limit fails with an exception, oldPosition will not be restored.

That's sort of on the caller if they don't coordinate well, but I think you're right that this would be better. I'll move the slicing inside the try (and leave the old gathering outside).

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/test/java/temper/core/SliceCoderTest.java
tjpalmer: > Should this assert that (byte)'T' and (byte)'o' are in buffer?

I saw that this wasn't being checked and decided not to worry it. Didn't check everything. But I can add it if you want.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
tjpalmer: Other than the reset topic, I've addressed requests in my latest commit.

mikesamuel (approved):

PR #411, opened 2026-05-01
dataFile builtin allows declaring a data file.

mikesamuel:
This commit adds a frontend mechanism for module level metadata by storing it as declaration metadata on a declaration that serves no purpose other than to hold that metadata, and it tweaks *TmpLTranslator* to skip translating that declaration but to associate the metadata with the *TmpL.Module*.

A new builtin, `dataFile(path, mimeType, data)`, uses that mechanism to add a module level annotation for the file.

A future PR will add a data backend that just bundles up data files and makes them available.

The plan is for the contextual autoescaper optimization pass to also generate a data file with translation strings extracted from message tags.

Backends like be-java should eventually use the be-data output to prepare their own resources directory.


fundamentals/src/commonMain/kotlin/lang/temper/type/MetadataValueMultimapImpl.kt
mikesamuel, line 1: Being able to switch between value oriented maps and edge oriented maps is convenient in TmpLTranslator.

be/src/commonTest/kotlin/lang/temper/be/tmpl/TmpLBackendTest.kt
```diff
@@ -4165,6 +4165,40 @@ class TmpLBackendTest {
...
+    fun moduleLevelMetadata() = assertGeneratedCode(
+        // We shouldn't generate a declaration for any module metadata.
+        // But we should store the metadata with the Module node.
+        // Backends like be-data do additional stuff with that, so we
+        // can test there that this works end-to-end.
+        inputJsonPathToContent = """
+            |{
+            |  foo: {
+            |    "foo.temper": ```
+            |      dataFile("foo.json", "application/json", '{ "my-favorite-number": 123 }');
```
tjpalmer, line 4178: So this pr doesn't yet generate the file, right? I guess that's what the future be-data is for, although I'm sure different other backends will want to include data files in different ways, which I think you're also planning for.

Just the plain ability to include raw data files into code is something we'll likely want sometime.

docs/for-users/temper-docs/docs/reference/builtins.md
```diff
@@ -1763,6 +1763,36 @@ console.log("done"); //!outputs "done"
...
+### Macro `datafile`
+The `datafile` macro declares a data file that can be bundled with the library translation.
+It takes three inputs:
+
+- The path to the data file relative to the module directory.
+  For example, if the current module is `my-library//foo` and the path is `bar.json`
+  then in `my-library`, the resource path `foo/bar.json` would allow retrieving the data
+  file in some target-language specific way.
+- The mime-type of the data.
+- The data itself.
```
tjpalmer, line 1779: Always a string for now?

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -3921,6 +3922,32 @@ class GenerateCodeStageTest {
...
+                repeat(1000) { append("ll") }
+                append("o, World!")
+            },
+        )});
+        """.trimMargin(),
+        want = """
+            |{
+            |  generateCode: {
+            |    body: ```
+            |      @topLevelMetadata @stay @declareDataFile((["hello.txt", "text/plain", "Hellllllllllllllllll⋯llllllllllllo, World!"])) @reach(\none) let moduleMetadata#0: Empty;
```
tjpalmer, line 3944: Nice to see the repeat collapsed.

tjpalmer (approved):

be/src/commonTest/kotlin/lang/temper/be/tmpl/TmpLBackendTest.kt
mikesamuel: Exactly.

mikesamuel: The frontend generates trees, and only backends generate files.  

The TmpL "backend" generates different trees, but only has a *Backend* class for testing purposes.

docs/for-users/temper-docs/docs/reference/builtins.md
mikesamuel: Yeah.  I figure we can add overloads for bytes later and maybe for things like JSON asts.

mikesamuel: It might be good to be able to read files in the work root at compile time and echo it as a data file.

PR #412, opened 2026-05-01
be-data: emits data files

mikesamuel:
The new *DataBackend* ignores temper sources and just looks for module decorations that specify data files and generates *OutputFileSpecification*s for them.

```
🐚$ temper repl
$ help(dataFile)
dataFile: Function
  Macro "datafile"
  The "datafile" macro declares a data file that can be bundled with the library translation.
  It takes three inputs:
  - The path to the data file relative to the module directory.
  For example, if the current module is "my-library//foo" and the path is "bar.json"
  then in "my-library", the resource path "foo/bar.json" would allow retrieving the data
  file in some target-language specific way.
  - The mime-type of the data.
  - The data itself.
  dataFile(
    "./my-data.json",
    "application/json",
    '["My", "data", "goes", "here"]',
  )
  // ✅ null
interactive#0: void
$ dataFile("hello-world.txt", "text/plain", "Hello, World!");
interactive#1: void
$ translate(1, "data")
Translated data for interactive#1
  interactive/
    i0001/
      hello-world.txt: text/plain
        Hello, World!
interactive#2: void
$
```


be-data/src/commonMain/kotlin/lang/temper/be/data/DataBackend.kt
```diff
@@ -0,0 +1,295 @@
...
+                    if (md.key.symbol == declareDataFileSymbol) {
+                        val (path: FilePath?, mimeType: MimeType?, data: String?) = run unpack@{
+                            val ls = TList.unpackOrNull((md.value as? TmpL.ValueData)?.value)
+                            if (ls?.size == NUM_DATA_FILE_DECORATION_PARTS) {
+                                val (pathStr, mimeTypeStr, data) = ls.map {
+                                    TString.unpackOrNull(it)
+                                }
+                                var path: FilePath? = null
+                                if (pathStr.isNullOrEmpty()) {
+                                    logSink.log(DataBackendMessage.ExpectedPath, pos, listOf(ls[0]))
```
tjpalmer, line 93: I haven't really worked out a pattern I like for error reporting in backends. Maybe I'll copy your style here.

be-data/src/commonMain/kotlin/lang/temper/be/data/DataBackend.kt
```diff
@@ -0,0 +1,295 @@
...
+        get() = false
+
+    override fun inlineToTree(
+        pos: Position,
+        arguments: List<TypedArg<Nothing>>,
+        returnType: Type2,
+        translator: Nothing,
+    ): Nothing {
+        error("Should not be called")
+    }
```
tjpalmer, line 288: This is nice, too.

bundled-backends/src/commonTest/kotlin/lang/temper/bundledBackends/BundledBackendsTest.kt
```diff
@@ -81,6 +88,7 @@ class BundledBackendsTest {
     fun defaultSupported() {
         val wanted = setOf(
             CSharpBackend.Factory.backendId,
+            DataBackend.Factory.backendId,
```
tjpalmer, line 91: Do we want standalone data files by default?

common/src/commonMain/kotlin/lang/temper/common/MimeType.kt
```diff
@@ -20,5 +20,17 @@ data class MimeType(val major: String, val minor: String) {
...
+            val parts = s.split('/')
+            if (parts.size != 2) {
+                return RFailure("Invalid mime type `$s`, expected major/minor")
+            }
+            val (major, minor) = parts
+            if (major.isEmpty() || minor.isEmpty()) {
+                return RFailure("Empty part in mime type `$s`")
+            }
+            return RSuccess(MimeType(major, minor))
+        }
```
tjpalmer, line 46: Indirectly tested through other tests?

tjpalmer (approved):

bundled-backends/src/commonTest/kotlin/lang/temper/bundledBackends/BundledBackendsTest.kt
mikesamuel: This is a good question.  I don't know.  Every TmpL translation costs time.

common/src/commonMain/kotlin/lang/temper/common/MimeType.kt
mikesamuel: I can add tests.

common/src/commonTest/kotlin/lang/temper/common/MimeTypeTest.kt
mikesamuel, line 1: Added tests and checked some restrictions from the RFC.

PR #416, opened 2026-05-08
frontend: always inline `for ... of ...` where possible.

mikesamuel:
`@inlineUnrealizedGoal` is a decoration for function parameters that allows inlining.

Previously the micropass that did that only inlined when there was an unrealized goal: the block lambda passed to that parameter uses `break`, `continue`, or `return` (sugar for assign output var and `break`) but the jump target is in a containing function instead of in the lambda.

This was causing problems with flow sensitive analyses, including in the *OptimizeContextualAutoescapingBlocks* micropass as in:

      let listOfNums = html"""
        "<ul>
        :for (let num of [1, 2, 3]) {
          "  <li>${ num }</li>
        :}
        "</ul>
        ;

The basic blocks before the loop were not connecting to those in the loop because they were in different functions.

Now, *InlineToRepairUnrealizedGoals* always inlines where it is allowed.

This commit also reworks operator precedence around `for ... of ...` loops to allow `of` to distribute over comma operators which effectively reserves syntax like `for (let element, index of ls)` and `for (let key, value of map)` in case we want to allow multiple parameters to `forEach` perhaps via something like:

    @overload("forEach")
    let forEachIndexed(
      @inlineUnrealizedGoal body: fn (T, Int32): Void
    ): Void


frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/InlineToRepairUnrealizedGoals.kt
```diff
@@ -87,9 +89,10 @@ internal fun inlineToRepairUnrealizedGoals(
...
 ): List<TEdge> {
     val inliner = InlineToRepairUnrealizedGoals(root, logSink)
-    if (!inliner.findFunctionsWithUnrealizedGoals()) { return emptyList() }
-    if (!inliner.findCallsToInline()) { return emptyList() }
-    inliner.doInlining()
+    inliner.findFunctions()
+    if (inliner.findCallsToInline()) {
+        inliner.doInlining()
+    }
     inliner.reportRemainingUnrealizedGoals()
```
mikesamuel, line 96: Always report problems.

frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -399,23 +399,50 @@ class TypeStageTest {
...
-        """,
+            |{
+            |  define: {
+            |    body:
+            |    ```
+            |    let console#0;
+            |    console#0 = doPure(@stay fn: Console {
+            |        getConsole()
+            |    });
+            |    do_bind_forEach(list("foo"))(fn (x__0) {
```
mikesamuel, line 410: Before type information, not yet inlined.

lexer/src/commonMain/kotlin/lang/temper/lexer/Operator.kt
```diff
@@ -106,160 +119,154 @@ enum class Operator(
      */
     PreCase(1, "case", Prefix),
 
-    /**
-     * `of` expects a declaration on the left so binds looser than
-     * [HighColon] and [Eq] in `@decoration let name: Type = expr`
-     * and the [At] used to decorate.
-     */
-    Of(2, "of", Infix, Right),
```
mikesamuel, line 114: This moved up.  The rest is just from running `scripts/scr renumber-...`

tjpalmer (approved):
Glad to see the inlining here. I could go either way on the `of` precedence, but the change is fine.

PR #418, opened 2026-05-11
frontend: replumb DotHelper to allow for operator overloading

mikesamuel:
This is part of a rewrite that will get us our desired operator semantics:

- `@operator("==")` for methods allows for user-defined equality, for example
- `+` can be applied to user defined types like big ints, and decimals
- basic arithmetic builtins are still functions so dispatch to them is fast
- all overloading is done via method/extension function overloading without any need for cover functions.

*DotHelper* allows collecting information about type member usage which includes extension functions.

This commit adds support for `@operator` and allows it to be applied to both instance methods and other callables.  It changes *DotHelper* to abstract out the kind of invocation to allow both *DotMember*s and *OperatorMember*s.

The goal is for operator members to be erased to dot members or functions that are annotated.

This commit gets as far as testing the new operator but is mostly to just get the *DotHelper* bits in place so the rest of the change can be plumbed through in follow on commits.

The plan for the next commit is to replace one use of a cover function tentatively (`+`) with a suite of recognize operator functions plumbed through via a special operator sugar function that desugars to *DotHelper*.


frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
```diff
@@ -3481,4 +3481,71 @@ class DefineStageTest {
...
+            |  @operator("+")
+            |  public static unary(c: C): C { c }
+            |}
+        """.trimMargin(),
+        want = """
+            |{
+            |  define: {
+            |    body: ```
+            |## mixedAdd has arity 2 so gets an infix operator specifier
+            |      @fn @operator("_+_") let mixedAdd__0;
```
tjpalmer, line 3507: I think it's fine to expand internally. And maybe that was the other context I commented on already. If so, no big deal there.

fundamentals/src/commonMain/kotlin/lang/temper/type/Member.kt
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * The text of the operator surrounded by underscores where operands are allowed.
+     * This allows for describing these different kinds of operators, as shown below
+     * where `+` is the sample operator.
+     *
+     * - Infix: `_+_`, the operator punctuation (or word) appears between two operands.
+     * - Prefix: `+_`, the operator appears before its operand, like `-` for negation.
+     * - Postfix: `_+`, the operator appears after its operand, like `++` for post-increment
+     *   or superscript `T` for matrix transpose.
+     */
```
tjpalmer, line 63: Fixing my old comment after paying better attention to this being the snippet source. Is this referring to internal representation? Could it confuse about source representation when in the docs?

tjpalmer (approved):

frontend/src/commonMain/kotlin/lang/temper/frontend/syntax/DotOperationDesugarer.kt
```diff
@@ -229,14 +236,18 @@ internal class DotOperationDesugarer(
...
-                                    if (metadataProblem != null) {
-                                        logSink.log(
-                                            Log.Error,
-                                            MessageTemplate.UnexpectedMetadata,
-                                            metadataEdge.target.pos,
-                                            listOf(extensionSymbol, metadataProblem),
-                                        )
+                                    reportAnyMetadataProblem(metadataEdge, extensionSymbol, metadataProblem)
+                                }
+                                parts.metadataSymbolMultimap[operatorSymbol]?.forEach { metadataEdge ->
```
mikesamuel, line 241: Adding extension resolutions for functions with `@overload` metadata

interp/src/commonMain/kotlin/lang/temper/interp/OperatorImplementationDecorator.kt
```diff
@@ -0,0 +1,102 @@
...
+        var hasOperandBefore = op.startsWith("_").also {
+            if (it) { op = op.drop(1) }
+        }
+        var hasOperandAfter = op.endsWith("_").also {
+            if (it) { op = op.dropLast(1) }
+        }
+
+        // If we got an operator text instead of a specifier, infer the kind of operator
+        // from the annotated function's maximum declared arity.
+        if (op.isNotEmpty() && !hasOperandAfter && !hasOperandBefore) {
```
mikesamuel, line 67: Infer (infix vs prefix) based on arity.  Without the underscores is the documented way of using it, but having the metadata make clear the operator convention helps users of the metadata avoid ambiguity after the function tree is separated from the declaration.

fundamentals/src/commonMain/kotlin/lang/temper/type/Member.kt
mikesamuel: This is the documentation for an operator specifier which might show up in diagnostics.  It's allowed but the recommended input to the decorator is just the operator text.

tjpalmer: > This is the documentation for an operator specifier which might show up in diagnostics. It's allowed but the recommended input to the decorator is just the operator text.

Sounds good.

PR #421, opened 2026-05-15
Update java to 25 and kotlin to 2.3

tjpalmer:


tjpalmer:
Latest failure is the java 25 runtime error happening on java functional tests:

```java
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by org.fusesource.jansi.internal.JansiLoader in an unnamed module (file:/C:/Users/work2/Apps/apache-maven-3.8.6/lib/jansi-2.4.0.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled

WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by com.google.common.util.concurrent.AbstractFuture$UnsafeAtomicHelper (file:/C:/Users/work2/Apps/apache-maven-3.8.6/lib/guava-25.1-android.jar)
WARNING: Please consider reporting this to the maintainers of class com.google.common.util.concurrent.AbstractFuture$UnsafeAtomicHelper
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
Hello, World!
```

I'm unsure why the jansi would apply to running funtests. But it clearly triggered, as did the guava thing. Neither really should apply to running temper-built java. And js funtests don't have the same problem. 🤷

If I get back to this topic later, maybe I'll focus on just getting our bundle on java 25 with `--enable-native-access=ALL-UNNAMED` and rely on integration tests? Also could test the new native code caching just from that side.

PR #424, opened 2026-05-22
Make python params positional or named

tjpalmer:
- Give pretty names to python constructor params (which they had in times past but got broken)
- Make other python params positional-only, as that's been available since python 3.8, which is below our min supported version
- As usual, see the backend test first
- Maybe this leaves only Lua without easily named constructor params? (although I want to change how our rust builders work)


be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
```diff
@@ -431,26 +431,40 @@ ComprehensionIf.operatorDefinition = `null`;
...
 Arg.defaultValue.default = `null`;
 Arg.operatorDefinition = `null`;
 Arg.prefix.default = `ArgPrefix.None`;
 
-enum ArgPrefix = None | Star | DoubleStar;
+/**
+ * [Slash] is standalone rather than a prefix, and [Star] can go either way, but
+ * keeping them all together still helps with organization.
+ */
+enum ArgPrefix = None | Slash | Star | DoubleStar;
```
tjpalmer, line 462: See the comment above for why I went this way with positional-only slash indicators, and see the backend test for what this looks like.

I actually coded up [two](https://github.com/temperlang/temper/blob/ce9e2b7d5dbd27cea96c227370fc5354ed849db2/be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar#L437) [other](https://github.com/temperlang/temper/blob/ca13ab29d6bcf6b8d8133c2feb1c6a8cb2e1d21f/be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar#L435) versions of how to do slashes, but I think this way is the best I came up with.

be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
```diff
@@ -431,26 +431,40 @@ ComprehensionIf.operatorDefinition = `null`;
...
+       || (prefix%ArgPrefix == Star => `starToken`)
        || (prefix == DoubleStar => `starStarToken`)
        || ())
-    & arg%Identifier
+    & arg%Identifier?
     & ((annotation%Expr? => ":" & annotation) || ())
     & ((defaultValue%Expr? => "=" & defaultValue) || ());
 Arg requires `prefix == ArgPrefix.None || defaultValue == null`;
+Arg requires `arg != null || prefix == ArgPrefix.Slash || prefix == ArgPrefix.Star`;
+Arg requires `arg != null || (annotation == null && defaultValue == null)`;
```
tjpalmer, line 452: This now also allows standalone star to separate out named-only params, but we don't currently generate that.

And I'm mixed on enforcing these rules since it feels somewhat like overhead when we also have funtests and such, but I felt like trying to conform to validation being done here already.

be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
```diff
@@ -476,7 +490,7 @@ CallArg {
 // Prohibit **x=y
 CallArg requires `prefix == ArgPrefix.None || arg == null`;
 // This ArgPrefix is used by Arg.
-CallArg requires `prefix != ArgPrefix.Star`;
+CallArg requires `prefix != ArgPrefix.Slash`;
```
tjpalmer, line 488: You actually can use star prefix on call args, but slash is only used in defs (and only standalone).

be-py/src/commonMain/kotlin/lang/temper/be/py/PyArgRules.kt
```diff
@@ -29,12 +30,16 @@ fun argumentsIssues(args: Iterable<Py.Arg>): String? {
...
+        if (slash && arg.prefix == Py.ArgPrefix.Slash) {
+            return "only one slash arg allowed at most"
+        }
         if (star && arg.prefix != Py.ArgPrefix.DoubleStar) {
             return "$arg: only **keywords can follow *vararg"
         }
         if (doubleStar) {
             return "$arg: nothing can follow **keywords"
         }
+        slash = slash || arg.prefix == Py.ArgPrefix.Slash
```
tjpalmer, line 42: Again, aligning with existing validation, even though I think it's mostly overkill.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyAst.kt
```diff
@@ -23,7 +23,7 @@ fun Py.Arguments.Companion.simple(pos: Position, vararg name: String): Py.Argume
     Py.Arguments.simple(pos, name.toList())
 
 fun Py.Arguments.Companion.simple(pos: Position, names: List<String>): Py.Arguments =
-    Py.Arguments(pos, names.map { Py.Arg(pos, it) })
+    Py.Arguments(pos, args = names.map { Py.Arg(pos, it) })
```
tjpalmer, line 26: Not actually needed here. I needed it for an earlier version of positional-only params on this branch, and I never reverted, but it's also not terrible, so I'm inclined to keep it.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyImportsExports.kt
```diff
@@ -390,7 +390,7 @@ private fun Py.Arg.gatherImports(sink: ImportSink, exclude: Set<OutName>) {
 
 private fun Py.Arguments.gatherNames(names: MutableSet<OutName>) {
     args.forEach { arg ->
-        names.add(arg.arg.outName)
+        arg.arg?.also { names.add(it.outName) }
```
tjpalmer, line 393: The arg name itself is now optional, so a little more juggling is needed here.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslator.kt
```diff
@@ -371,7 +371,7 @@ class PyTranslator(
         ): List<Py.Stmt> = buildList {
             // TODO: substitute python parameter names for TmpL names
             // See be-java's javadoc(...) helpers.
-            val fnDocumentation = s.documentation?.prettyPleaseHelp()
+            val fnDocumentation = s.documentation.prettyPleaseHelp()
```
tjpalmer, line 374: Cleanup recommended by ij.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslator.kt
```diff
@@ -976,6 +976,17 @@ class PyTranslator(
...
+        val isConstructor = func is TmpL.Constructor
+        // Manage slash of positional args.
+        var slashAddedIfNeeded = false
+        fun addSlashIfNeeded() {
+            if (!slashAddedIfNeeded && args.isNotEmpty()) {
+                args.add(Py.Arg(params.pos, arg = null, prefix = Py.ArgPrefix.Slash))
+            }
+            // Track either way since we might just skip it if wanted early, such as for rest param at front.
+            slashAddedIfNeeded = true
+        }
```
tjpalmer, line 988: Here's new logic support for adding `/` for positional-only args.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslator.kt
```diff
@@ -1011,6 +1022,14 @@ class PyTranslator(
...
                 pyNames.name(name)
             }.asPyId(id.pos)
+            val prefix = when (kind) {
+                ArgKind.This, ArgKind.Required, ArgKind.Optional -> Py.ArgPrefix.None
+                ArgKind.Rest -> {
+                    // Slash has to go before rest arg.
+                    addSlashIfNeeded()
+                    Py.ArgPrefix.Star
+                }
+            }
```
tjpalmer, line 1032: Slash goes before any rest arg. (And were we planning to stop support rest args? Did we already stop but haven't changed tmpl yet?)

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslator.kt
```diff
@@ -1063,7 +1085,10 @@ class PyTranslator(
                     func.pos,
                     name = testName(func.name),
                     // Tests shouldn't have parameters nor need renames.
-                    args = Py.Arguments(func.pos, listOf(Py.Arg(func.pos, arg = pyIdent(func.pos, "self")))),
+                    args = Py.Arguments(
+                        func.pos,
+                        args = listOf(Py.Arg(func.pos, arg = pyIdent(func.pos, "self"))),
+                    ),
```
tjpalmer, line 1091: No real change here. Just another explicit `args =` that I'd added earlier on this branch.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -29,12 +29,12 @@ class PyBackendTest {
             |U_3 = TypeVar1('U_3', bound = Any0)
             |class Thing(Generic2[T_1]):
             |    __slots__ = ()
-            |    def identity(this_2, u_11: 'U_3', f_12: 'Callable3[[U_3], T_1]') -> 'T_1':
+            |    def identity(this_2, u_11: 'U_3', f_12: 'Callable3[[U_3], T_1]', /) -> 'T_1':
             |        return f_12(u_11)
-            |    def __init__(this_5) -> None:
+            |    def __init__(this, /) -> None:
```
tjpalmer, line 34: I changed all constructor params to render pretty and didn't bother to ask the `this`/`self` to stay ugly. And it should still stay positional-only, even if it would only matter in syntax like `Thing.__init__(this=...)`, which I tested as otherwise working.

And I'm also ok still saying `this` instead of `self` for now, but we could change to `self` in the future if we verify against other params being called `self`.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -67,16 +67,16 @@ class PyBackendTest {
...
-            |    def int64_to_int(this_0, int_9: 'int64_24') -> 'int3':
+            |    def int64_to_int(this_0, int_9: 'int64_24', /) -> 'int3':
             |        return int64_to_int32_35(int_9)
-            |    def string_to_int(this_1, string_12: 'str4') -> 'int3':
+            |    def string_to_int(this_1, string_12: 'str4', /) -> 'int3':
             |        return string_to_int32_36(string_12, this_1.radix_7)
-            |    def __init__(this_2, radix_15: 'int3') -> None:
-            |        this_2.radix_7 = radix_15
+            |    def __init__(this, /, radix: 'int3') -> None:
+            |        this.radix_7 = radix
```
tjpalmer, line 75: Example constructor with non-this arg. And maybe uninteresting for single-param case like this, but you still now can say `IntMaker(radix=...)`, which should still be fine since temper would allow `{ class: IntMaker, radix: ... }`.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -67,16 +67,16 @@ class PyBackendTest {
...
             |    @property
-            |    def radix(this_24) -> 'int3':
+            |    def radix(this_24, /) -> 'int3':
             |        return this_24.radix_7
-            |def crazy_sum(int_maker_16: 'IntMaker', int_17: 'int64_24', string_18: 'str4') -> 'int3':
+            |def crazy_sum(int_maker_16: 'IntMaker', int_17: 'int64_24', string_18: 'str4', /) -> 'int3':
```
tjpalmer, line 79: And we could/should presumably make all params pretty now, given the `/` to prevent named use, but I didn't do that in this pr, since I didn't want to think about new name conflicts.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyTranslatorTest.kt
```diff
@@ -219,9 +219,9 @@ class PyTranslatorTest : TranslatorTests(PyBackend.Python3.backendMeta, PySuppor
                     |class Thing_7:
                     |    prop_name_9: 'str0'
                     |    __slots__ = ('prop_name_9',)
-                    |    def __init__(blah_12: 'str0') -> None:
+                    |    def __init__(blah: 'str0', /) -> None:
```
tjpalmer, line 222: This case is from hand-crafted tmpl that gives `this` the name `blah`.

be/src/commonMain/kotlin/lang/temper/be/names/DescriptorChain.kt
```diff
@@ -89,8 +89,9 @@ fun DescriptorChain.idKind(): TmpL.IdKind = when (node) {
 
 /** Determine if this name is internal or is presented to external modules. */
 fun DescriptorChain.idReach(ignoreImport: Boolean = false): TmpL.IdReach = when (val n = this.node) {
+    is TmpL.Formal if parent?.node is TmpL.Constructor -> TmpL.IdReach.External
     is TmpL.FunctionDeclaration -> n.idReach()
-    is TmpL.Import -> if (ignoreImport) TmpL.IdReach.Internal else TmpL.IdReach.External
+    is TmpL.Import if !ignoreImport -> TmpL.IdReach.External
     else -> TmpL.IdReach.Internal
 }
```
tjpalmer, line 96: The constructor case above is what makes constructor params pretty now. I also refactored the logic a bit for the import case.

be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
mikesamuel: Ok, so slash needs to appear on its own in a parameter list.

What would it look like if instead of the below, *Arguments* had named arguments before positional arguments, so the `/` is still on its own in the comma separated list, but the count of children still matches the count of arguments (modulo `*` and `**`).

    Arguments ::= args%Arg*", ";

tjpalmer: > What would it look like if instead of the below, _Arguments_ had named arguments before positional arguments, so the `/` is still on its own in the comma separated list, but the count of children still matches the count of arguments (modulo `*` and `**`).
> 

Do you mean [like this that I linked to](https://github.com/temperlang/temper/blob/ce9e2b7d5dbd27cea96c227370fc5354ed849db2/be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar#L437) in a different comment?

If so, here's what was trouble there:

- I had to navigate separate param lists
- I couldn't figure out how to prevent a comma after an ending slash
- It doesn't jive well with the `*` vs `*args` option of python semi-rest/named-only-divider thinking
- It required changing more code to support it
- It didn't fit in as well with the existing star and starStar token usage

be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
mikesamuel: `**` too

```python3
$ python3
Python 3.14.3 (main, Feb  3 2026, 15:32:20) [Clang 17.0.0 (clang-1700.6.3.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def f(**x): print(repr(x))
...
>>> f(**{"foo": "bar"})
{'foo': 'bar'}
>>>
```

mikesamuel (approved):

PR #430, opened 2026-05-30
fix(IfTransform): when with is-Type arms now infers arm_types | Void

### LINKED ISSUE #427: when expression with is-Type arms infers Void instead of arm type
## Summary

```
when (s) {
  is Circle -> "circle";
  is Square -> "square";
}
```

## Repro

## Root cause

```kotlin
        controlFlow.pos,
        listOf(
            controlFlow,
            ControlFlow.Stmt(
            ),
        ),
    )
}
```

## The fix

```
let return__0;
```

to:

```
let return__0;
if (t#0) {
} else {
  if (t#1) {
    return__0 = "no"
  } else {
  }
}
```

## Note on completeness

## Prototype

notactuallytreyanastasio:
Fixes #427.

## What this changes

Removes 14 lines from `IfTransform.complicate` (`interp/src/commonMain/kotlin/lang/temper/interp/IfTransform.kt`, lines 221–234) that wrapped the nested if-chain for a `when` without `else` in a trailing `ControlFlow.StmtBlock([if-chain, Stmt(void)])`.

## Why it was wrong

Because `StmtBlock.wrap` is idempotent (a StmtBlock input is returned unchanged), `ControlFlowTransform.invoke` set the `StructuredFlow` to this `StmtBlock([If(...), Stmt(void)])`. `forwardMaximalPaths` walked exit paths and always found the trailing `Stmt(void)` as the terminal expression, never the arm bodies. `MakeResultsExplicit` then initialised the result variable to `void` unconditionally, and the Typer inferred `Void` for the entire expression.

## Why the fix works

The `void` inserted at lines 190–195 in the innermost synthetic else clause already marks the fallthrough ("no arm matched") exit path. Without the outer sequential `Stmt(void)`, `forwardMaximalPaths` finds the arm body expressions as terminal expressions alongside the innermost void. The inferred type becomes `arm_types | Void` instead of `Void`.

## Before / after

**Before** — `when (c) { is String -> "yes"; is Int -> "no"; }` at Type stage:
```
let return__0;
return__0 = void
```

**After**:
```
let return__0;
if (t#0) {
  return__0 = "yes"
} else {
  if (t#1) {
    return__0 = "no"
  } else {
    return__0 = void   ← fallthrough; no exhaustiveness checking yet
  }
}
```

## Scope

This does not implement exhaustiveness checking for sealed types. The fallthrough `void` path remains. Functions with an explicit non-void return type still require an `else` arm (or `else -> bubble()`) to compile without error. That is a separate feature.

## Tests

- `TypeStageTest.whenIsTypeNoElse` — regression test asserting the new Type stage output
- `control-flow/when-is-type-inference/when-is-type-inference.temper.md` — functional tests asserting runtime correctness of `describe`, `classify`, and recursive sealed dispatch
- Full frontend test suite passes (`./gradlew :frontend:jvmTest`)

Found while porting [chess9](https://github.com/notactuallytreyanastasio/chess9) to Temper.

### DIFF
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
@@ -2394,6 +2394,49 @@ class TypeStageTest {
             |}
         """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
     )
+
+    // https://github.com/temperlang/temper/issues/427
+    // A `when` expression without `else` whose arms return String should assign
+    // the return variable on each arm (String | Void), not initialise the whole
+    // result to void (Void). Before the IfTransform fix, the body was
+    // `let return__0; return__0 = void` — the arms were never assigned.
+    @Test
+    fun whenIsTypeNoElse() = assertModuleAtStage(
+        input = "when (c) { is String -> \"yes\"; is Int -> \"no\"; }",
+        moduleResultNeeded = true,
+        stage = Stage.Type,
+        want = """
+        {
+          type: {
+            body:
+            ```
+            let return__0;
+            var t#0, t#1;
+            if (!false) {
+              t#0 = is(c, String)
+            } else {
+              t#0 = false
+            };
+            if (t#0) {
+              return__0 = "yes"
+            } else {
+              if (!false) {
+                t#1 = is(c, Int32)
+              } else {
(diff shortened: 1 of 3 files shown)


notactuallytreyanastasio:
Closing per Tom's feedback: the Void inference for when-without-else is expected behavior until exhaustiveness checking is implemented for sealed types. No bug — users need an else arm for now.

PR #433, opened 2026-06-05
C++ Backend

ShawSumma:
Add C++ backend.

This PR gets the C++ backend to close to the level of the others.


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/RunCpp.kt
```diff
@@ -0,0 +1,189 @@
...
+    val gpp = this[GppCommand]
+    // The set of .cpp files to compile comes straight from the dependency graph the
+    // frontend already recorded — the current library's files plus those of every
+    // transitive dependency — rather than re-deriving it by scanning #include lines.
+    // This mirrors how the Rust backend hands its dependency set to Cargo.
+    val depCppFiles = transitiveCppDepFiles(libraryName, dependencies)
+
+    val gppArgs = buildList {
+        add("-I..")
+        add("-std=c++14")
```
tjpalmer, line 106: I have this configurable in cppv. It's ok to hardcode for now, but maybe could test different versions in a future pr.

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/nullable_param.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <type_traits>
+#include "temper_bubble.hpp"
+
+namespace temper {
+    namespace core {
+
+        template<class T>
+        struct NullableParam {
+            T value;
+            bool has_value;
```
tjpalmer, line 11: I'm unsure why NullableParam separate from nullable generally. And I'd like this to be compatible with `std::optional` for c++17 and later convenience. But I don't think we need to hold up this pr over that topic.

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/mapped.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+namespace temper {
+    namespace core {
+
+        namespace Mapped {
+
+            template<class Key, class Value>
+            struct Ordered {
+                std::vector<Key> order;
+                std::map<Key, Value> data;
```
tjpalmer, line 19: Maybe should use `std::unordered_map` since we're maintaining our own separate order, anyway?

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/promise.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+        // Async here is cooperative and single-threaded: instead of blocking a thread,
+        // a generator suspends at each `await` and registers a continuation that is run
+        // when the awaited promise settles. Continuations are dispatched through a FIFO
+        // queue so that independently launched async blocks interleave in a predictable
+        // order rather than recursing into one another.
+        inline std::deque<std::function<void()>>& async_task_queue() {
+            static std::deque<std::function<void()>> queue;
+            return queue;
+        }
```
tjpalmer, line 21: I also made my own separate async thing for rust for now. I have an abstraction that should allow for multi-threaded, but I've only implemented single-threaded for now. And I'd like to see if I could make it work in an executor-agnostic fashion for standard rust async, but I haven't worked on that at all, yet.

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/regex.hpp
```diff
@@ -0,0 +1,326 @@
+#pragma once
+#include <cstdint>
+#include <iomanip>
+#include <memory>
+#include <regex>
```
tjpalmer, line 5: Google says `std::regex` is bad, but I have no experience. If it gets us working for now, I'm happy with it being here for now.

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/temper_bubble.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include <exception>
+#include <string>
+
+namespace temper {
+    namespace core {
+
+        struct TemperBubble : std::exception {
+            std::string message;
```
tjpalmer, line 9: I'd still like a no exceptions option for c++, but this is fine for this pr.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -394,6 +394,155 @@ fun TmpL.Tree.referencedNames(): Sequence<ResolvedName> = sequence {
...
+ */
+fun mutatingMemberNames(
+    modules: Iterable<TmpL.Module>,
+    /**
+     * Member dotNames a backend wants treated as non-const for its own reasons (e.g. a
+     * representation that can't be const-qualified). Seeded into the analysis so the
+     * constraint propagates transitively through `this`-calls and override slots.
+     */
+    forcedMutating: Set<String> = emptySet(),
+): Set<String> {
```
tjpalmer, line 451: I wonder if I could use this idea in RustTranslator. Unsure off the top of my head, but I do calculate when a class is mutable or not. And not in an `@imu` sense but in an "is it imu right now by chance" sense. I abstract either way, but the abstraction is cheaper for imu.

functional-test-suite/src/commonMain/kotlin/lang/temper/tests/FunctionalTestStatus.kt
```diff
@@ -25,8 +25,10 @@ val functionalTestStatus: Map<Ft, List<IssueCheck>> = buildMap {
     issue(Ft.RegexZeroAdvance, lua(166))
     issue(Ft.NamesNonascii, lua(228))
     214.let { issue(Ft.TypesNetresponse, cpp(it), interp(it), lua(it)) }
-    onlyPasses(
+    onlyFails(
         cpp(198),
+        Ft.RegexMatch,
+        Ft.TypesNetresponse,
     )
```
tjpalmer, line 32: 🎉 Fantastic.

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/casting.hpp
```diff
@@ -0,0 +1,72 @@
...
+        // `x is SomeValueType` when x is a boxed AnyValue: a plain non-null check would
+        // wrongly accept a box holding a different type. Mirrors Rust's temper_core::is.
+        template<class T>
+        bool is_box(const std::shared_ptr<AnyValueBase>& src) {
+            return std::dynamic_pointer_cast<AnyValueBox<T>>(src) != nullptr;
+        }
+
+        template<class T>
+        std::shared_ptr<AnyValueBox<T>> checked_cast_box(const std::shared_ptr<AnyValueBase>& src) {
+            std::shared_ptr<AnyValueBox<T>> result = std::dynamic_pointer_cast<AnyValueBox<T>>(src);
```
tjpalmer, line 29: I guess we need rtti turned on for this, but I'm unsure we should try to get around that, given Temper's needs. Probably better than hand-baking our own rtti.

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/dense_bit_vector.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+namespace temper {
+    namespace core {
+        namespace DenseBitVector {
+
+            inline std::shared_ptr<std::vector<bool>> make(int32_t size) {
+                return std::make_shared<std::vector<bool>>(size, false);
+            }
+
+            inline bool get(std::shared_ptr<std::vector<bool>> bv, int32_t index) {
```
tjpalmer, line 16: Someone's gotta use `std::vector<bool>`, so it might as well be us.

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/int.hpp
```diff
@@ -0,0 +1,152 @@
...
+#include <cstdint>
+#include <limits>
+#include <string>
+#include "temper_bubble.hpp"
+#include "base_types.hpp"
+
+namespace temper {
+    namespace core {
+
+        namespace Int {
```
tjpalmer, line 12: I'd likely name this Int32, since that's our formal name, and Int is an alias. But I'm not worried for this pr.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -23,39 +28,638 @@ class CppBackendTest {
...
+    }
+}
+
+private fun assertGeneratedContains(
+    temper: String,
+    cppContains: List<String>,
+) {
+    val logSink = ListBackedLogSink()
+    val result = generateCode(
+        inputs = listOf(filePath("something", "something.temper") to temper.trimMargin()),
```
tjpalmer, line 657: So we've specifically made a module called `something`. What do things look like for top level?

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -23,39 +28,638 @@ class CppBackendTest {
                 |#include <my-test-library/something.hpp>
                 |namespace temper {
                 |  namespace my_test_library {
```
tjpalmer, line 30: A non-top-of-library-level submodule named `something` should have a namespace `something` for it also, and we *shouldn't* at all have user code under namespace `temper`. So I imagine this should be namespace `my_test_library::something`, not `temper::my_test_library`.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -23,39 +28,638 @@ class CppBackendTest {
...
-                |      temper::core::log(console_0, name__0);
+                |    static std::shared_ptr<temper::core::Console::Type> console_0;
+                |    static int32_t x;
+                |    void greet(std::string name) {
+                |      temper::core::Console::log(console_0, "Hi:");
+                |      temper::core::Console::log(console_0, name);
                 |      return;
                 |    }
-                |    temper::core::Void tmp_1() {
+                |    void temper_init_something() {
```
tjpalmer, line 38: I'd love to see auto static init for the common case. Maybe we could have preproc definitions for requesting manual init if needed. I have manual init in be-rust, because standard rust has no automatic init. I've considered testing for init inside every public function to automate this if not done manually, but haven't been convinced. There's also 3rd-party crates for generalized static init that I've considered pulling in.

I still want a manual init option, though, because that's my way of configuring things like async runner. I haven't worked out a final plan.

tjpalmer (approved):
The namespacing bothers me a lot (and see my comments in the backend test for more details on that), and I've stated a few other opinions here and there as well. I'm also unsure how much of this we could use for c++03, but I could still see about how to align as well as possible when getting back to that.

All that said, mostly I'm happy to see be-cpp so close to our usual finished level. Good work pushing on this.

And to clarify, I'd personally be fine merging this as is, even though I think some things should definitely and other things maybe should get adjusted in future PRs.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -394,6 +394,155 @@ fun TmpL.Tree.referencedNames(): Sequence<ResolvedName> = sequence {
...
+        ReceiverMember(dotName.dotNameText, parameters, body, overridden, isSetter = false, mayYield = mayYield)
+    is TmpL.Getter ->
+        ReceiverMember(dotName.dotNameText, parameters, body, overridden, isSetter = false, mayYield = false)
+    is TmpL.Setter ->
+        ReceiverMember(dotName.dotNameText, parameters, body, overridden, isSetter = true, mayYield = false)
+    else -> null
+}
+
+/**
+ * Member dotNames (methods, getters, setters) across [modules] whose receiver (`this`)
```
mikesamuel, line 431: This is something that you could only know for a particular version of a library and not across versions unless we make it explicit in the source, right?

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/Cpp.out-grammar
```diff
@@ -28,17 +28,33 @@ Namespace ::= "namespace" & (name%SingleName || ()) & "{" & body%Global*"" & "}"
...
+DerivedStructDef ::= "struct" & name%Name & ":" & bases%BaseSpec*"," & "{" & fields%StructPart*"" & "}" & ";";
+// A single base class is inherited `virtual` (to keep a shared `AnyValueBase` subobject
+// under diamond inheritance); with multiple bases each is plain `public`.
+BaseSpec ::= virt%VirtualMod? & "public" & base%Type;
+BaseSpec.virt.default = `null`;
+enum VirtualMod = Virtual;
+AnyStructDef = StructDef | DerivedStructDef;
+TemplateStructDef ::= "template" & "<" & typeParams%FuncParam*"," & ">" & def%AnyStructDef;
+TemplateFuncDef ::= "template" & "<" & typeParams%FuncParam*"," & ">" & def%FuncDef;
+TemplateFuncDecl ::= "template" & "<" & typeParams%FuncParam*"," & ">" & decl%FuncDecl;
```
mikesamuel, line 40: Would it make sense to factor out the `"<" & typeParams%FuncParam*"," & ">"` into a TemplateTypeParameters node?

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/regex.hpp
mikesamuel: Google uses PCRE internally.  It's partly performance, and partly it's better if everyone uses the same regex library so all code reviewers only have to learn one set of regex library quirks.

ShawSumma:
@mikesamuel had asked me to write about what AI use went into this PR.

I had used claude code to review and edit many of the changes in this but each had been then reviewed by me before PR.

PR #436, opened 2026-06-05
Remove mapDropping

tjpalmer:
- One way to fix #394
- I personally would like either to drop it or to base it on null instead of exceptions
- I can draft other options, but this was the easiest for quick proposal


frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -595,11 +595,11 @@ class GenerateCodeStageTest {
...
     @Ignore // TODO(mikesamuel): Fix typing of generic methods with explicit actuals
     @Test
-    fun mapDroppingTypes() = assertModuleAtStage(
+    fun mapTypeArg() = assertModuleAtStage(
         stage = Stage.GenerateCode,
         input = """
             |let ls: List<Int> = [1, 2];
-            |ls.mapDropping<String> { (x: Int): String => x.toString(10) }
+            |ls.map<String> { (x: Int): String => x.toString(10) }
         """.trimMargin(),
```
tjpalmer, line 603: I checked that the above gets an error:

```
Member map defined in List__0 | Listed__0 incompatible with usage!
```

If I remove the `<String>`, I get no error.

mikesamuel: Odd.

mikesamuel (approved):

PR #439, opened 2026-06-13
Panic for else on exhaustive sealed when

### LINKED ISSUE #427: when expression with is-Type arms infers Void instead of arm type

tjpalmer:
- Fix #427
- Panic when a sealed type is checked exhaustively rather than providing a void value
- For now, this exhaustiveness check relies on simple type checking conditions in the else-if chain, which should be the case for our current basic when-is block arms
- I didn't include the functional test from #430, but I manually checked it, and the TypeStageTest case here shows that we can expect correct semantics

```
> temper test -b py
Tests passed: 3 of 3
```

### DIFF
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
@@ -2394,6 +2394,74 @@ class TypeStageTest {
             |}
         """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
     )
+
+    @Test
+    fun sealedWhen() = assertModuleAtStage(
+        stage = Stage.Type,
+        pseudoCodeDetail = PseudoCodeDetail.default.copy(showInferredTypes = true),
+        input = """
+            |export sealed interface Shape {}
+            |export class Circle() extends Shape {}
+            |export class Square() extends Shape {}
+            |export let describe(s: Shape): String {
+            |  when (s) {
+            |    is Circle -> "circle";
+            |    is Square -> "square";
+            |  }
+            |}
+        """.trimMargin(),
+        want = """
+            |{
+            |  type: {
+            |    body:
+            |      ```
+            |      @typeDecl(Shape) @stay @sealedType let `test//`.Shape ⦂ Type;
+            |      `test//`.Shape = type (Shape);
+            |      @typeDecl(Circle) @stay let `test//`.Circle ⦂ Type;
+            |      `test//`.Circle = type (Circle);
+            |      @typeDecl(Square) @stay let `test//`.Square ⦂ Type;
+            |      `test//`.Square = type (Square);
+            |      @fn let `test//`.describe ⦂(fn (Shape): String), @typePlaceholder(Shape) typePlaceholder#0: Empty;
+            |      typePlaceholder#0 = {class: Empty__0};
+            |      @fn @visibility(\public) @stay @fromType(Circle) let constructor__0 ⦂(fn (Circle): Void);
+            |      constructor__0 = (@stay fn constructor(@impliedThis(Circle) this__0: Circle) /* return__0 */: Void {
+            |          return__0 = void
+            |      });
+            |      @fn @visibility(\public) @stay @fromType(Square) let constructor__1 ⦂(fn (Square): Void);
+            |      constructor__1 = (@stay fn constructor(@impliedThis(Square) this__1: Square) /* return__1 */: Void {
+            |          return__1 = void
+            |      });
+            |      `test//`.describe = (@stay fn describe(s__0 /* aka s */: Shape) /* return__2 */: String {
+            |          var t#0 ⦂ Boolean, t#1 ⦂ Boolean;
+            |          void;
+            |          fn__0: do {
+            |            if (!false) {
+            |              t#0 = is(s__0, Circle)
+            |            } else {
+            |              t#0 = false
+            |            };
+            |            if (t#0) {
+            |              return__2 = "circle"
+            |            } else {
+            |              if (!false) {
+            |                t#1 = is(s__0, Square)
+            |              } else {
+            |                t#1 = false
+            |              };
+            |              if (t#1) {
+            |                return__2 = "square"
+            |              } else {
+            |                return__2 = panic ⋖ String ⋗()
+            |              }
+            |            };
+            |          }
+            |      })
+            |
+            |      ```
+            |  }
+            |}
+        """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
+    )
 }
 
 private object ImpureIgnoreFn : NamedBuiltinFun, CallableValue {
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
@@ -2407,6 +2407,7 @@ class TypeStageTest {
             |  when (s) {
             |    is Circle -> "circle";
             |    is Square -> "square";
+            |    // defaults to panic here because above is exhaustive
             |  }
             |}
         """.trimMargin(),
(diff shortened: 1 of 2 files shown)


frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2394,6 +2394,74 @@ class TypeStageTest {
...
+        input = """
+            |export sealed interface Shape {}
+            |export class Circle() extends Shape {}
+            |export class Square() extends Shape {}
+            |export let describe(s: Shape): String {
+            |  when (s) {
+            |    is Circle -> "circle";
+            |    is Square -> "square";
+            |  }
+            |}
```
tjpalmer, line 2412: Look, Ma, no else!

frontend/src/commonTest/kotlin/lang/temper/frontend/TypeStageTest.kt
```diff
@@ -2394,6 +2394,74 @@ class TypeStageTest {
...
+            |              if (!false) {
+            |                t#1 = is(s__0, Square)
+            |              } else {
+            |                t#1 = false
+            |              };
+            |              if (t#1) {
+            |                return__2 = "square"
+            |              } else {
+            |                return__2 = panic ⋖ String ⋗()
+            |              }
```
tjpalmer, line 2456: We can see the value assignments as well as the panic above. Maybe best even still to retain the panic in generated code for languages with exhaustiveness checking, especially for those that might be open in the runtime.

If we target some language later that *disallows* `else` in exhaustive cases, we'll have to deal with that when we come to it.

interp/src/commonMain/kotlin/lang/temper/interp/IfTransform.kt
```diff
@@ -188,11 +196,21 @@ internal object IfTransform : ControlFlowTransform("if") {
...
                         null
                     } else {
-                        // Having a `void` here makes sure that static checks like UseBeforeInit
-                        // get a diagnostic position for error logging.
-                        ControlFlow.Stmt(
-                            macroCursor.referenceToVoid(macroCursor.macroEnvironment.pos.rightEdge),
-                        )
+                        when {
+                            // If someone has checked all subtypes for some supertype, panic instead of void.
+                            anySealedTypesExhaustive(branches) -> {
```
tjpalmer: This function is where the work happens.

interp/src/commonMain/kotlin/lang/temper/interp/IfTransform.kt
```diff
@@ -217,25 +235,47 @@ internal object IfTransform : ControlFlowTransform("if") {
...
-                if (controlFlow != null && !hasFinalElse) {
-                    controlFlow = ControlFlow.StmtBlock(
-                        controlFlow.pos,
-                        listOf(
-                            controlFlow,
-                            ControlFlow.Stmt(
-                                macroCursor.referenceToVoid(controlFlow.pos.rightEdge),
-                            ),
-                        ),
-                    )
```
tjpalmer, line 234: Here's the same removal from #430, but I didn't include the comment explaining the removal, given the extra context above. But based on tests, I believe Bobby is right that the other void (or now alternatively panic) is enough for our remaining handling.

interp/src/commonMain/kotlin/lang/temper/interp/IfTransform.kt
```diff
@@ -217,25 +235,47 @@ internal object IfTransform : ControlFlowTransform("if") {
...
+/**
+ * Check for simple sealed exhaustiveness. This requires that a single name be
+ * checked against each sealed subtype of some sealed supertype. If so, default
+ * to panic instead of void, which allows for type inference without an else.
+ *
+ * Rely on checks elsewhere for valid downcasts. And if the above holds and we
+ * could downcast, the check would be exhaustive in any case.
```
tjpalmer, line 270: Comment above explains the strategy.

interp/src/commonMain/kotlin/lang/temper/interp/IfTransform.kt
```diff
@@ -217,25 +235,47 @@ internal object IfTransform : ControlFlowTransform("if") {
...
+    var checkedName: TemperName? = null
+    branches@ for (branch in branches) {
+        val condition = branch.first?.target as? CallTree ?: continue@branches
+        condition.childOrNull(0)?.valueContained?.stateVector == BuiltinFuns.isFn || continue@branches
+        val nextCheckedName = condition.childOrNull(1)?.nameContained ?: continue@branches
+        when (checkedName) {
+            null -> checkedName = nextCheckedName
+            else if checkedName != nextCheckedName -> break@branches
+            else -> {}
+        }
```
tjpalmer, line 285: This requires that all checks be for the same *named* value. If we want to be more flexible, we could track all checks for all named values, but `when` blocks should all be for the same, anyway.

And yes, this needs to be a named value, since that's much easier to check. Maybe our `when` handling already invents a named value when needed? I don't remember for sure.

interp/src/commonMain/kotlin/lang/temper/interp/IfTransform.kt
```diff
@@ -217,25 +235,47 @@ internal object IfTransform : ControlFlowTransform("if") {
...
+        val subtype = condition.childOrNull(2)?.valueContained?.stateVector as? ReifiedType ?: continue@branches
+        val subdef = subtype.type2.definition
+        supertypes@ for (supertype in subdef.superTypes) {
+            val supershape = supertype.definition as? TypeShape
+            val sealedSubs = supershape?.sealedSubTypes ?: continue@supertypes
+            if (subdef in sealedSubs) {
+                val foundSealedSubs = allFoundSealedSubs.getOrPut(supershape) { mutableSetOf() }
+                foundSealedSubs.add(subdef)
+                if (foundSealedSubs.size == sealedSubs.size) {
```
tjpalmer, line 294: In a set, we can check just the sizes instead of the contents.

mikesamuel:
I like the way this is done at the post-desugaring `if` step.

I wonder what we should do with a case like this:

```mermaid
classDiagram
    High <|-- Mid1
    High <|-- Mid2
    Mid1 <|-- Low1A
    Mid1 <|-- Low1B
    Mid2 <|-- Low2A
    Mid2 <|-- Low2B

```

By pure analysis of the following desugared `when`, it covers all Mid1 but does not cover Mid2.

```ts
sealed interface High {}

sealed interface Mid1 {}
sealed interface Mid2 {}

class Low1A extends Mid1 {}
class Low1B extends Mid1 {}
class Low2A extends Mid2 {}
class Low2B extends Mid2 {}

let f(x: High): String {
  when (x) {
    is Low1A -> "A";
    is Low1B -> "B";
  }
}
```

Could we detect that and issue an informative error message after we have type info?

mikesamuel (approved):

mikesamuel:
Approving. I think we can add some testcases and discuss ways to represent in the IR the expectation about when the inserted panic is unreachable, and how to undo that when the context really is void or issue a warning when it's not and explain the mismatched assumption.

tjpalmer:
> Could we detect that and issue an informative error message after we have type info?

Actually, looks like we could go simpler with just:

```ts
interface High {}

sealed interface Mid1 extends High {}
class Low1A extends Mid1 {}
class Low1B extends Mid1 {}

class Low2 extends High {}
```

Anyway, good catch. I guess I was just being genius. I'll check out some of the things you've recommended and make appropriate changes before merging this.

tjpalmer:
And just to demo what type checks are and aren't allowed:

```ts
$ interface A {}; sealed interface B extends A {}; class C extends B {}; class D extends A {}
interactive#0: D__3
$ let d: A = new D();
interactive#1: void
$ d is C
interactive#2: false
$ d is D
interactive#3: true
$ new D() is C
1: new D() is C
   ┗━━━━━━━━━━┛
[interactive#4:1+0-12]@G: Runtime type check from D__3 to C__2 can never succeed
interactive#4: false
```

I just hadn't been thinking thoroughly when I originally coded this.

tjpalmer:
Closed in favor of #452, which reuses some code from this branch.

PR #432, opened 2026-06-04
Use decorators for `@imu` and `@partialImu`

tjpalmer:
- Fix disambiguate stage handling of decorators and metadata on type formals
- Conditionally add stay nodes to type formals
- Remove Imu and PartialImu types
- No changes yet to the checked rules, and changes still are needed such as:
  - Preventing hiding mutability
  - Not sure if more still is needed on treating effectively imu things as imu for cases like `<@imu T>`

PR #440, opened 2026-06-18
Syntax support for decoratable type formals

mikesamuel:
Type formals, like class's and interface's type shapes can have type metadata. PR #432 uses this to allow restricting type parameters to Imu types.

In the below, we need to allow application of arbitrarily many decorators of decorators:

    @W @X class C<@Y @Z T extends D, U> extends E, F {}

Unfortunately, the commas mean different things there. The first comma ends the `extends clause` so that another type formal, U, can be declared separately from T.  The second comma separates two distinct bounds that are extended.

Ideally programmers would just use `&` to separate different extended types, but there's a lot of muscle memory working against that.

This commit reworks *CanNest* to receive an innermost bracket so it can forbid the comma eating version of extends inside angle brackets but not outside.

It also tweaks *Grammer* to allow for decoration on type formals and changes the early representation of complex type arguments (including those with upper bounds) to mirror the representation of complex value arguments and consolidates some code around decorators and extends-like operator parsing.

Refactorings:

- `extends` is no longer a builtin, it's consistently represented as symbol metadata.
- `@in` and `@out` are now the explicit way to mention variance and the docs for those point out that they're deprecated.
- share code that handles type formal declarations between *DisambiguateTypeMacro* (for type formals on classes and interfaces) and *formalizeTypeArgs* (for type formals on functions).

Bugs fixed:

- a very odd bug in the definition of `/` (or) in our parser combinators
- *TypeDisambiguateMacro* was collecting some error messages but not doing anything with them.
- precedence fixup so *ExtendsComma* agrees with the other extends-like operators


astbuild/src/commonMain/kotlin/lang/temper/astbuild/Combinator.kt
```diff
@@ -347,8 +347,8 @@ internal data class Or private constructor(val options: List<Combinator>) : Comb
             }
             return when (flatList.size) {
                 0 -> empty
-                1 -> options[0]
-                else -> Or(options)
+                1 -> flatList[0]
+                else -> Or(flatList)
```
mikesamuel, line 351: Unsure how any of this worked before since the `/` combinator notation calls into this.

I ran into this when trying to fold with *Or.empty* as the zero for matching `extends` like operators in Grammar.kt.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -377,29 +381,7 @@ val grammar = ProductionNames.run {
...
-                                (
-                                    Operator.Paren y `(` y Expr y
-                                        "(" y TopLevelNoGarbageNoComment y ")" y `)` y `)`
-                                    )
-                            )
-                        ),
-                )
-                )
-            )
+        DecoratedLet / decorated(TopLevelNoGarbageNoComment)
```
mikesamuel, line 384: Pulled all the decorator handling stuff out into one helper.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -443,42 +432,16 @@ val grammar = ProductionNames.run {
...
-     * Captures low precedence operators that may follow a parenthesized argument list.
+     * Captures low-precedence operators that may follow a parenthesized argument list.
      *
      * - `: ReturnType` desugars to `\outType`, `ReturnType`.
      * - `extends SuperType` and `implements SuperType* desugars to `\super`, `SuperType`.
      */
     CalleeAndArgs `：＝` (
         // The "return type" desugars to a declaration of an output parameter
         // named "return"
-        (
```
mikesamuel, line 454: Also pulled all the extends-like handling out into a helper.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -1021,16 +975,19 @@ val grammar = ProductionNames.run {
...
-        Operator.Leaf y (
-            callTree(
-                `(` y
-                    ("in".rename(ParsedName("@in")) / "out".rename(ParsedName("@out"))) y
-                    Id y `)`,
-            )
+        // Output complex type arguments like complex value arguments,
+        // a block with metadata that is converted early on into a declaration.
+        (
+            allExtendsLike y blockTree(
```
mikesamuel, line 984: Here's the main change.  Now, the extends handling all uses the metadata form instead of generating calls to the extends function.

This allows simplifying some of the disambiguate stage code and mirrors what we do with complex value arguments.

astbuild/src/commonTest/kotlin/lang/temper/astbuild/BuildTreeTest.kt
```diff
@@ -5038,10 +5064,41 @@ class BuildTreeTest {
...
+        wantJson = """
+            |[ "Call", [
+            |    [ "RightName", "let" ],
+            |    [ "Value", "\\word: Symbol" ], [ "LeftName", "f" ],
+            |
+            |    [ "Value", "\\typeArg: Symbol" ],
+            |    [ "Call", [
+            |        [ "RightName", "@" ],
+            |        [ "RightName", "imu" ],
+            |        [ "Block", [
```
mikesamuel, line 5079: Here's an example of the block form that mirrors the complex value argument form.

cst/src/commonMain/kotlin/lang/temper/cst/CanNest.kt
```diff
@@ -103,11 +136,11 @@ fun canNest(
             },
         )
 
-        // `extends` with commas is not allowed inside angle brackets or comma lists.
+        // `extends` with commas is not allowed inside angle brackets, even decorated, or in comma lists.
         child.operator in extendsLikeCommaOperators &&
             (
                 parent.operator == Operator.Comma ||
-                    (parent.operator == Operator.Angle && parent.childCount >= 2)
+                    (bracket?.operator == Operator.Angle && bracket.childCount >= 2)
```
mikesamuel, line 143: See the PR description for why this is necessary.  In short, wrapping a decoration around means we can't have enough context.

docs/for-users/temper-docs/docs/reference/builtins.md
```diff
@@ -2041,66 +2041,6 @@ TODO: Document how Temper enum types relate to types per backend.
 
 <!-- /snippet: builtin/enum -->
 
-<!-- snippet: builtin/extends : # `extends` keyword -->
-
-<a name="builtin&#45;extends" class="snippet-anchor-name"></a>
-
-### *SubType* `extends` *SuperType*
```
mikesamuel, line 2048: Extends is no longer a builtin so this moved to the docs for `interface`.

frontend/src/commonTest/kotlin/lang/temper/frontend/AssertModuleAtStage.kt
```diff
@@ -111,6 +113,7 @@ fun assertModuleAtStage(
     moduleResultNeeded: Boolean = false,
     loc: ModuleName? = null,
     stagingFlags: Set<BuiltinName> = emptySet(),
+    stackTracesForErrors: Boolean = false,
```
mikesamuel, line 116: Dumping a trace can help figure out what to rework.

lexer/src/commonMain/kotlin/lang/temper/lexer/Operator.kt
```diff
@@ -258,11 +258,6 @@ enum class Operator(
     DollarCurly(24, $$"${", Prefix, closer = true, minArity = 0, maxArity = 1),
     UnicodeRun(24, "\\u{", Prefix, closer = true, minArity = 0, maxArity = Int.MAX_VALUE),
 
-    /**
-     * Tag is a prefix operator, so that use of brackets outside of infix position gets us
-     * something vaguely like JSX tag expressions.
-     */
-    Tag(24, "<", Prefix, closer = true, minArity = 1, maxArity = Int.MAX_VALUE),
```
mikesamuel, line 265: I was playing around way back with React style tags.  Got rid of it.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -561,10 +524,10 @@ val grammar = ProductionNames.run {
         )
 
     /**
-     * Relates a match case, e.g. a pattern, to a consequence of matching that pattern.
+     * Relates a match case, e.g., a pattern, to a consequence of matching that pattern.
      */
     MatchBranch `：＝` (
-        // Put else first to special case it before general expressions. //
+        // Put else first to special-case it before general expressions.
```
tjpalmer, line 530: I probably had the trailing `//` from some time ago. I sometime put that on lines to make spots while working. And sometimes I fail to remember or notice them when making prs or merging.

astbuild/src/commonMain/kotlin/lang/temper/astbuild/Grammar.kt
```diff
@@ -2006,7 +2001,7 @@ internal abstract class DomainSpecificLanguage {
      * ```
      *
      * Whether a `{` is followed by properties or statements is made based on the preceding
-     * non-comment & non-space token.
+     * non-comment, non-space token.
```
tjpalmer, line 2004: `&` vs `,` in action, in English.

astbuild/src/commonTest/kotlin/lang/temper/astbuild/BuildTreeTest.kt
```diff
@@ -5066,27 +5124,68 @@ class BuildTreeTest {
...
+            |        [ "RightName", "@" ],
+            |        [ "RightName", "in" ],
+            |        [ "Call", [
+            |            [ "RightName", "@" ],
+            |            [ "RightName", "on" ],
+            |            [ "Block", [
+            |                [ "Value", "\\typeArg: Symbol" ],
+            |                [ "RightName", "K" ],
+            |                [ "Value", "\\super: Symbol" ],
+            |                [ "RightName", "MapKey" ],
```
tjpalmer, line 5172: Based on the tldr, I expecting something more equivalent to `@extends(MapKey) K`, but this is fine, too.

parser/src/commonTest/kotlin/lang/temper/parser/ParseTest.kt
```diff
@@ -2960,16 +2825,16 @@ class ParseTest {
 
     @Test
     fun adjacentWords() = assertParseTree(
-        input = "public class Foo extends Bar",
+        input = "export class Foo extends Bar",
```
tjpalmer, line 2828: Good catch. I make this mistake a lot, independent of who made this one (which I haven't looked up).

tjpalmer (approved):
I think we should merge this before #432 because this one is more fundamental.

PR #441, opened 2026-06-19
Use qnames as connected keys

tjpalmer:
- Use plain `@connected` now without keys
- Still require `@connected` for treating as connected
- Minimize backend implementation code changes
- Fix unrelated rust `zeroize` version problem while working this code
- Change the exposed name (although not yet the file name) of implicits to core, per discussion on discord


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppSupportNetwork.kt
```diff
@@ -118,147 +118,147 @@ internal object CppSupportNetwork : SupportNetwork {
         connectedKey: String,
         genre: Genre,
     ): SupportCode? = when (connectedKey) {
-        "::getConsole" -> Like.core("get_console")
```
tjpalmer: Lots of this is effectively search-and-replace. But the details sometimes get subtle.

be-cppv/src/commonMain/resources/lang/temper/be/cppv/temper-core/int.hpp
```diff
@@ -110,7 +110,7 @@ Expected<int32_t> to_int32(int64_t i) {
     i < std::numeric_limits<int32_t>::min() ||
     i > std::numeric_limits<int32_t>::max()
   ) {
-    return Unexpected("Int64::toInt32");
+    return Unexpected("core.type Int64.toInt32()");
```
tjpalmer, line 113: Also from search and replace. These changes aren't technically needed but still seemed work doing.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpSupportNetwork.kt
```diff
@@ -817,14 +819,15 @@ internal val listify = StaticCall(
...
-private val listBuilderAdd = StaticCall("ListBuilder::add", StandardNames.temperCoreListedAdd)
-private val listBuilderAddAll = StaticCall("ListBuilder::addAll", StandardNames.temperCoreListedAddAll)
-private val listBuilderRemoveLast = StaticCall("ListBuilder::removeLast", StandardNames.temperCoreListedRemoveLast)
-private val listBuilderReverse = StaticCall("ListBuilder::reverse", StandardNames.temperCoreListedReverse)
-private val listBuilderSort = StaticCall("ListBuilder::sort", StandardNames.temperCoreListedSort)
-private val listBuilderSplice = StaticCall("ListBuilder::splice", StandardNames.temperCoreListedSplice)
+private val listBuilderAdd = StaticCall("core.type ListBuilder.add()", StandardNames.temperCoreListedAdd)
+private val listBuilderAddAll = StaticCall("core.type ListBuilder.addAll()", StandardNames.temperCoreListedAddAll)
+private val listBuilderRemoveLast =
+    StaticCall("core.type ListBuilder.removeLast()", StandardNames.temperCoreListedRemoveLast)
```
tjpalmer, line 825: Sometimes need extra line wrapping because of the longer names.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpSupportNetwork.kt
```diff
@@ -835,8 +838,10 @@ private object ListedFilter : CSharpInlineSupportCode("Listed::filter") {
     }
 }
 
-private val listForEach = StaticCall("List::forEach", StandardNames.temperCoreListedForEach)
-private object ListedGet : CSharpInlineSupportCode(listedTypes.map { "$it::get" } + listOf("Mapped::get")) {
+private val listForEach = StaticCall("core.type List.forEach()", StandardNames.temperCoreListedForEach)
+private object ListedGet : CSharpInlineSupportCode(
+    listedTypes.map { "core.type $it.get()" } + listOf("core.type Mapped.get()"),
+) {
```
tjpalmer, line 844: Some backends need map-name logic updates.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpSupportNetwork.kt
```diff
@@ -1328,42 +1340,51 @@ private object IsNull : CSharpInlineSupportCode("isNull", BuiltinOperatorId.IsNu
...
+private val dateGetDay = PropertyAccess("std/temporal.type Date.day", "Day")
 private val dateGetDayOfWeek = StaticCall(
-    "Date::getDayOfWeek",
+    "std/temporal.type Date.get dayOfWeek()",
     StandardNames.temperStdTemporalTemporalSupportIsoWeekdayNum,
 )
-private val dateGetMonth = PropertyAccess("Date::getMonth", "Month")
-private val dateGetYear = PropertyAccess("Date::getYear", "Year")
+private val dateGetMonth = PropertyAccess("std/temporal.type Date.month", "Month")
+private val dateGetYear = PropertyAccess("std/temporal.type Date.year", "Year")
```
tjpalmer, line 1373: The missing `get` is interesting for these cases, but we'll see more on this later.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpTranslator.kt
```diff
@@ -1535,7 +1520,7 @@ internal class CSharpTranslator(
         val definition = nominalType.definition
 
         // TODO: Shouldn't we have a TmpL.TypeName here?
-        val connectedKey = definition.metadata[connectedSymbol, TString]
+        val connectedKey = definition.connectedKey
```
tjpalmer, line 1537: I defined convenience `connectedKey` accessors in several places. These all check if `@connected` is set and then return the qname if any.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaSupportNetwork.kt
```diff
@@ -1365,12 +1371,12 @@ val JavaLang.mappedToListBuilderWith by receiver { separateCode(temperMappedToLi
...
 val JavaLang.mapBuilderRemove by receiver { separateCode(temperMapBuilderRemove) }
 val JavaLang.mapBuilderSet by receiver {
-    inlineSupport("MapBuilder::set", arity = 3, needsSelf = true) { pos, args ->
+    inlineSupport("core.type MapBuilder.set()", arity = 3, needsSelf = true) { pos, args ->
         args[0].method("put", args[1], args[2], pos = pos)
     }
 }
 val JavaLang.mapBuilderConstructor by receiver {
-    inlineSupport("MapBuilder::constructor", arity = 0) { pos, _ ->
+    inlineSupport("core.type MapBuilder.constructor()", arity = 0) { pos, _ ->
```
tjpalmer, line 1379: I think these names are mostly ignored, and the names in the map below are what's used. In be-csharp and be-rust, I used these names to make the map. But that's not the case in all backends right now.

be-java/src/commonMain/resources/lang/temper/be/java/temper-core/src/main/java/temper/core/Core.java
```diff
@@ -358,7 +358,7 @@ public static long float64ToInt64(double n) {
     }
 
     /**
-     * Implements connected method {@code Float64::toString}.
+     * Implements connected method {@code core.type Float64.toString()}.
```
tjpalmer, line 361: Again, not needed, but seemed nice. Not sure any checks against mistakes in these.

be-js/src/commonMain/kotlin/lang/temper/be/js/JsSupportNetwork.kt
```diff
@@ -1508,15 +1509,20 @@ private val ignoreIdiomExpander: Inliner =
...
-            } else {
-                segment.asciiTitleCase()
-            }
-        }
-        .joinToString(""),
+            when (i) {
+                0 -> segment.asciiUnTitleCase()
+                else -> segment.asciiTitleCase()
+            }.trimEnd('(', ')') // Exclude parens.
+        }.joinToString(""),
```
tjpalmer, line 1525: I had to make similar changes in be-lua and be-tmpl. I mostly tried to minimize difference from previous calculated names, without customizing the logic too much for things that were previously inconsistent.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/net.js
```diff
@@ -5,7 +5,7 @@
  * @param {string | null} bodyMimeType
  * @return {Promise<Response>}
  */
-export function stdNetSend(url, method, bodyContent, bodyMimeType) {
+export function sendRequest(url, method, bodyContent, bodyMimeType) {
```
tjpalmer, line 8: This one lost some context in the automatic naming, but it seems unlikely to cause collisions later.

And maybe a future PR could make backend implementation naming more robust and incorporate more qname info, even though I tried to minimize changes here.

Clearly any user-space connections need to be based on thorough name information. But I suspect we should generate expected names from more structured processing like already do for translation in most cases, rather than basing those on munged qnames. And maybe once we have that standardize for user-space for each backend, maybe we could come back and change core/std handling to rely on the standardized logic and naming. But basing on qname strings was easier for now in this PR.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/regex.js
```diff
@@ -112,7 +112,7 @@ export const regexCompiledSplit = (_, compiled, text) => {
  * @param {string} formatted
  * @returns {RegExp}
  */
-export const regexCompileFormatted = (_, formatted) => {
+export const regexFormatterRegexCompileFormatted = (_, formatted) => {
```
tjpalmer, line 115: This one got longer in backends with name munging. Otherwise, I'd have had to special-case this everywhere.

be-js/src/commonTest/js/runtime-support/connecteds-supported-test.mjs
```diff
@@ -73,8 +73,8 @@ const skipped = new Set([
     "float64Sqrt",
     "float64Tan",
     "float64Tanh",
-    "::getConsole",
-    "ignore",
+    "core.getConsole()",
+    "core.ignore()",
```
tjpalmer: Some of these are definitely scoped more clearly than before.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
```diff
@@ -1833,19 +1840,19 @@ do
...
     return date.month
 end
 
-function temper.date_getday(date)
+function temper.date_day(date)
     return date.day
 end
 
-function temper.date_getdayofweek(date)
+function temper.date_dayofweek(date)
```
tjpalmer, line 1848: I either needed to drop `get` in some cases or add it in others, because we weren't consistent in our old keys.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
```diff
@@ -152,10 +152,11 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
                     append("regex = { version = \"=1.12.2\", optional = true }\n")
                     append("time = { version = \"=0.3.41\", optional = true }\n")
                     append("ureq = { version = \"=3.1.2\", optional = true }\n")
+                    append("zeroize = { version = \"=1.8.2\", optional = true }\n")
```
tjpalmer, line 155: This is an incidental fix for something that freshly broke on us. Crate `ureq` depends on `rustls` which currently depends on `zeroize` 1.8, but without locking down to 1.8. So cargo was trying to get 1.9, which depends on rust 2024 edition, which isn't supported until rust 1.85.

So this explicit dependency gets us an old version of zeroize that still works. We might just need to update our minimum rust someday, though.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -1114,3 +1025,8 @@ fun toSigBestEffort(descriptor: Descriptor?) = when (descriptor) {
     is Signature2 -> descriptor
     is Type2 -> withType(descriptor, fn = { _, sig, _ -> sig }, fallback = { null })
 }
+
+fun Map<Symbol, Value<*>>?.connectedKey(): String? = when {
+    this != null && connectedSymbol in this -> this[qNameSymbol]?.let { TString.unpackOrNull(it) }
+    else -> null
+}
```
tjpalmer, line 1271: Most of my helpers were on things like MemberShape or such, where context helps more than just some map. But I wanted this map case for convenience also.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLTranslator.kt
```diff
@@ -1401,20 +1397,14 @@ class TmpLTranslator internal constructor(
...
-                                    }
+                        val connectedKey = decl?.parts?.connectedKey
+                        if (connectedKey != null) {
+                            val sig = constructor.descriptor
+                                ?.let { constructorSig ->
+                                    factorySignatureFromConstructorSignature(constructorSig)
                                 }
+                            if (sig != null) {
+                                return@findConnectedKey connectedKey to sig
                             }
```
tjpalmer, line 1408: I actually cut down lines in several places, even if overall like count is up (in part due to lots of line wrapping).

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLTranslator.kt
```diff
@@ -3624,13 +3605,13 @@ private fun genericComparisonHackaround(
...
-            BuiltinOperatorId.EqGeneric -> return "StringIndexOption::compareTo::eq"
-            BuiltinOperatorId.NeGeneric -> return "StringIndexOption::compareTo::ne"
-            BuiltinOperatorId.CmpGeneric -> return "StringIndexOption::compareTo"
+            BuiltinOperatorId.LtGeneric -> return "core.type StringIndexOption.compareTo()::lt"
+            BuiltinOperatorId.LeGeneric -> return "core.type StringIndexOption.compareTo()::le"
+            BuiltinOperatorId.GeGeneric -> return "core.type StringIndexOption.compareTo()::ge"
+            BuiltinOperatorId.GtGeneric -> return "core.type StringIndexOption.compareTo()::gt"
+            BuiltinOperatorId.EqGeneric -> return "core.type StringIndexOption.compareTo()::eq"
+            BuiltinOperatorId.NeGeneric -> return "core.type StringIndexOption.compareTo()::ne"
+            BuiltinOperatorId.CmpGeneric -> return "core.type StringIndexOption.compareTo()"
```
tjpalmer, line 3614: I wasn't sure what to do with the `::...` pseudo-keys here, so I left them somewhat as before. Gets the job done for now.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TranslateDotHelper.kt
```diff
@@ -473,54 +470,28 @@ internal object TranslateDotHelper {
...
-    } else {
-        null
-    }
-}
-
 private fun propertyShapeFor(methodShape: MethodShape): PropertyShape? =
     methodShape.enclosingType.properties.firstOrNull { it.symbol == methodShape.symbol }
 
-private const val GETTER_AFFIX = "::get"
-private const val SETTER_AFFIX = "::set"
```
tjpalmer, line 522: Ok, this is an awkward thing that I just chose a simple solution for at this time. In `std/temporal`, we have `year`, `month`, and `day` properties on `Date` that aren't explicitly defined as getters. And qname assignment currently doesn't apply to the generated getters/setters, and they don't get seen here. So there's still no clear way to connect both getters and setters for such things at this time.

A future PR could improve the frontend handling of such things, or maybe we could allow `@connected` only on explicit getters and setters, not simple property declarations. But for the scope of this PR, we no longer can manually name things `Date::getYear`, we also can't check for such manual naming here, either, so I also no longer filter above.

This works for now, but for user-space we'll need to choose a better solution in a future PR, as discussed in my previous paragraph.

be/src/commonTest/kotlin/lang/temper/be/tmpl/TestBackend.kt
```diff
@@ -177,19 +178,23 @@ internal open class TestSupportNetwork(
...
             else -> TestSupportCode(baseName = baseName, signature = null)
         }
     }
 
+    private fun mungeQName(qname: String): String = qname
+        .split(".", "::", " ")
+        // Skip module name and type qualifier.
+        .subListToEnd(1)
+        .filter { it != "type" }
+        .joinToString("") { it.trimEnd('(', ')').asciiTitleCase() }
```
tjpalmer, line 196: Vaguely similar to js and lua handling. I chose to leave get/set in the naming here, because seems like tmpl backend ought to be more explicit.

be/src/commonTest/kotlin/lang/temper/be/tmpl/TmpLBackendTest.kt
```diff
@@ -830,9 +830,9 @@ class TmpLBackendTest {
             |           content:
             |               ```
             |               //// work//isEmptyWrapper/ => isEmptyWrapper.tmpl
-            |               let StringIsEmpty#0 = builtins.StringIsEmpty;
+            |               let StringGetIsEmpty#0 = builtins.StringGetIsEmpty;
             |               let return__0(@QName("test-library/isEmptyWrapper.(s)") s__1: String): Boolean {
-            |                 return StringIsEmpty#0(s__1);
+            |                 return StringGetIsEmpty#0(s__1);
```
tjpalmer, line 835: So we get the extra `Get` in here. I chose to exclude this in js and lua.

frontend/src/commonMain/kotlin/lang/temper/frontend/implicits/ImplicitsModule.kt
```diff
@@ -132,7 +132,7 @@ object ImplicitsModule {
             )
             module.advance()
             if (DEBUG) {
-                console.group("Implicits module at ${module.stageCompleted}") {
+                console.group("Core module at ${module.stageCompleted}") {
```
tjpalmer, line 135: Not strictly needed, but trying to prevent the implicits name from leaking into messages.

frontend/src/commonMain/kotlin/lang/temper/frontend/implicits/ListFns.kt
```diff
@@ -25,50 +25,50 @@ import kotlin.math.max
...
+    object Get : SigFnBuilder("core.type List.get()") {
         override fun invoke(args: ActualValues, cb: InterpreterCallback, interpMode: InterpMode): PartialResult {
             val ls = TList.unpackContent(args[0])
             val i = TInt.unpackOrFail(args, 1, cb, interpMode) { return@invoke it }
-            if (i < 0 || i > Int.MAX_VALUE) {
+            if (i < 0) {
                 throw Panic()
             }
-            return ls.getOrNull(i.toInt()) ?: Fail
+            return ls.getOrNull(i) ?: Fail
```
tjpalmer, line 41: Applied some cleanup recommended by ij. This is straggling from when we went Int32.

frontend/src/commonMain/kotlin/lang/temper/frontend/syntax/AttachQNameMetadata.kt
```diff
@@ -39,13 +38,14 @@ import lang.temper.value.staticPropertySymbol
...
 import lang.temper.value.typeFormalSymbol
+import lang.temper.value.varSymbol
 
 /**
  * Attaches [qNameSymbol] to [QName] metadata to [DeclTree]s.
  */
 internal fun attachQNameMetadata(module: Module, root: BlockTree) {
     val (libraryName, relPath) = when (val loc = module.loc) {
-        is ImplicitsCodeLocation -> implicitsLibraryName to FilePath.emptyPath
+        is ImplicitsCodeLocation -> coreLibraryName to FilePath.emptyPath
```
tjpalmer, line 48: I changed only some of the var naming from implicits to core in this pr.

frontend/src/commonMain/kotlin/lang/temper/frontend/syntax/AttachQNameMetadata.kt
```diff
@@ -100,6 +101,9 @@ internal fun attachQNameMetadata(module: Module, root: BlockTree) {
                     qNameBuilder.part(parsedName, kind)
                     val qName = qNameBuilder.toQName()
                     qNameToDecls.putMultiList(qName, t)
+                    if (initial is FunTree && varSymbol !in metadata) {
+                        declToFn[t] = initial
+                    }
```
tjpalmer, line 106: Track non-var function values for putting qnames on along with their decls.

frontend/src/commonMain/kotlin/lang/temper/frontend/syntax/AttachQNameMetadata.kt
```diff
@@ -121,7 +125,7 @@ internal fun attachQNameMetadata(module: Module, root: BlockTree) {
             if (parts != null) {
                 val metadata = parts.metadataSymbolMultimap
                 val typeDefined = metadata[typeDefinedSymbol, TType]
-                val definition = ((typeDefined as? ReifiedType)?.type as? NominalType)?.definition
+                val definition = (typeDefined?.type as? NominalType)?.definition
```
tjpalmer, line 128: Recommended by ij due to improved kotlin inference.

frontend/src/commonMain/kotlin/lang/temper/frontend/syntax/AttachQNameMetadata.kt
```diff
@@ -161,17 +165,24 @@ internal fun attachQNameMetadata(module: Module, root: BlockTree) {
...
+                        V(pos, nameTextValue)
+                    }
+                }
             }
         }
     }
 }
 
-val implicitsLibraryName = DashedIdentifier("implicits")
+val coreLibraryName = DashedIdentifier(ImplicitsCodeLocation.diagnostic)
```
tjpalmer, line 185: Centralize naming more.

frontend/src/commonMain/resources/implicits/Implicits.temper
```diff
@@ -118,12 +118,12 @@ export class Null extends Equatable, Imu {
  * //!outputs  "Empty is not null: true"
  * ```
  */
-@connected("Empty")
+@connected
```
tjpalmer, line 112: Nice to see the keys gone.

frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
```diff
@@ -496,6 +496,8 @@ class DefineStageTest {
                                             ["Value", "\"test-code.type C.constructor().return=\": String"],
                                           ]
                                         ],
+                                        ["Value", "\\QName: Symbol"],
+                                        ["Value", "\"test-code.type C.constructor()\": String"],
```
tjpalmer, line 500: Got qname on most fun trees now.

frontend/src/commonTest/kotlin/lang/temper/frontend/SyntaxMacroStageTest.kt
```diff
@@ -285,6 +287,7 @@ class SyntaxMacroStageTest {
         input = """
         let f = fn (x) {};
         let g = fn g(y) {};
+        var h = fn (z) {}; // *var* fn should get neither symbol nor qname
```
tjpalmer, line 290: Added for testing the `var` case.

frontend/src/commonTest/kotlin/lang/temper/frontend/SyntaxMacroStageTest.kt
```diff
@@ -350,6 +357,35 @@ class SyntaxMacroStageTest {
...
+                            [ "LeftName", "z__0" ],
+                            [ "Value", "\\word: Symbol" ],
+                            [ "Value", "\\z: Symbol" ],
+                            [ "Value", "\\QName: Symbol" ],
+                            [ "Value", "\"test-code.h().(z)\": String" ],
+                          ]
+                        ],
+                        [ "Value", "\\returnedFrom: Symbol" ],
+                        [ "Value", "true: Boolean" ],
+                        [ "Block", [
```
tjpalmer, line 375: No qname before the block here because var, even though qname does go on the param still just above. Maybe we should remove that in a future pr?

fundamentals/src/commonMain/kotlin/lang/temper/env/Exporter.kt
```diff
@@ -39,6 +42,13 @@ data class Export(
     override val reifiedType: Value<*>?
         get() = null // Could we get one from typeInferences?
 
+    val connectedKey: String?
+        get() = when {
+            connectedSymbol in declarationMetadata ->
+                declarationMetadata[qNameSymbol]?.lastOrNull()?.let { TString.unpackOrNull(it) }
+            else -> null
+        }
```
tjpalmer, line 50: One of a few similar helpers.

fundamentals/src/commonMain/kotlin/lang/temper/value/FnParts.kt
```diff
@@ -20,7 +20,7 @@ class FnParts internal constructor(
     // Since the actual type of the varArg from the caller's perspective doesn't exist as a decl
     // we need a more advanced type here
     val restFormal: RestFormal?,
-    val connected: String?,
+    val connected: Boolean,
```
tjpalmer, line 23: I considered dropping this and just requiring going to metadata, but I didn't do that here.

fundamentals/src/commonMain/kotlin/lang/temper/value/FnParts.kt
```diff
@@ -130,7 +134,7 @@ internal fun decomposeFun(tree: FunTree, metadataMultimap: LiveMetadataMap): FnP
         typeFormals = typeFormals,
         formals = formals,
         restFormal = typeOfList(restFormal.firstOrNull()),
-        connected = connected,
+        connected = connectedSymbol in metadataMultimap,
```
tjpalmer, line 137: Here's the bool value.

fundamentals/src/commonMain/kotlin/lang/temper/value/PseudoCode.kt
```diff
@@ -359,6 +359,7 @@ internal class PseudoTreeBuilder(
                     filterBySymbol = { symbol ->
                         when (symbol) {
                             in typeMemberMetadataSymbols -> detail.showTypeMemberMetadata
+                            qNameSymbol -> detail.showQNames
```
tjpalmer, line 362: Need to check this for fun trees now, too.

interp/src/commonMain/kotlin/lang/temper/interp/ConnectedDecorator.kt
```diff
@@ -1,52 +1,27 @@
...
-                        values = listOf(TString, r),
-                    ),
-                    TProblem,
-                )
-            }
-        }
-    }
+) {
+    void
 }
```
tjpalmer, line 25: Just void now. If we want the option to specify backend names here in the future, we'll have to come back to this.

interp/src/commonMain/kotlin/lang/temper/interp/Interpreter.kt
```diff
@@ -184,6 +184,11 @@ class Interpreter(
     private var goingOutOfStyle = stage == Stage.Run
     private val isProcessingImplicits = nameMaker.namingContext.isImplicits
 
+    /** Helps centralize tracking access to connecteds. */
+    fun connection(qname: String?): ((Signature2) -> Value<*>)? {
+        return connecteds[qname]
+    }
```
tjpalmer, line 190: This made dev and debug easier, even though it's boring. It was more elaborate in the middle of dev.

interp/src/commonMain/kotlin/lang/temper/interp/Interpreter.kt
```diff
@@ -1829,7 +1834,7 @@ class Interpreter(
             restInputsType = restType?.type2,
             typeFormals = typeFormals.toList(),
         )
-        val connected = connecteds[parts?.connected]?.let { it(signature) }
+        val connected = connection(parts?.connectedKey)?.let { it(signature) }
```
tjpalmer, line 1837: Use the centralized method now.

interp/src/commonMain/kotlin/lang/temper/interp/Interpreter.kt
```diff
@@ -2469,7 +2474,9 @@ class Interpreter(
         override val isProcessingImplicits: Boolean
             get() = document.isImplicits
 
-        override fun connection(connectedKey: String): ((Signature2) -> Value<*>)? = connecteds[connectedKey]
+        override fun connection(qname: String): ((Signature2) -> Value<*>)? {
+            return this@Interpreter.connection(qname)
```
tjpalmer, line 2478: Here, too.

name/src/commonMain/kotlin/lang/temper/name/ModuleLocation.kt
```diff
@@ -40,20 +40,20 @@ sealed interface ModuleLocation : CodeLocation, Comparable<ModuleLocation>, Toke
...
         is ImplicitsCodeLocation -> 0
         is ModuleName -> -1
     }
 
-    override fun toString() = wordImplicits.text
+    override fun toString() = wordCore.text
 }
 
-private val wordImplicits = OutputToken("implicits", OutputTokenType.Word)
+private val wordCore = OutputToken("core", OutputTokenType.Word)
```
tjpalmer, line 56: Per discussion, expose name `core` instead of `implicits`, especially for all the qname exposure in this pr.

be-js/src/commonTest/js/runtime-support/connecteds-supported-test.mjs
```diff
@@ -117,6 +117,9 @@ const skipped = new Set([
     "stringIndex",
     "valueResult",
     "valueResultConstructor",
+    // Other custom things.
+    "dequeIsEmpty",
+    "promiseBuilderPromise",
```
tjpalmer, line 122: Actually, looks like I include `Get` for js, but some of the manually mapped ones don't have that.

be-js/src/commonTest/js/runtime-support/connecteds-supported-test.mjs
```diff
@@ -130,11 +133,18 @@ const skipped = new Set([
...
+                // Duplicate logic from JsSupportNetwork's connectedKeyToExportedName.
+                return connectedMethod
+                    .split(/\.|::/)
+                    .slice(1)
+                    // .filter((x) => x != "type")
+                    .map((x) => x.split(" ").at(-1))
+                    .map((x) => x.replace(/[()]+$/, ""))
+                    .map((x, i) => (i ? x[0].toUpperCase() : x[0].toLowerCase()) + x.slice(1))
+                    .join("");
+            });
```
tjpalmer, line 147: I still want to delete this test, and I still haven't.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppSupportNetwork.kt
mikesamuel, line 1: Already, this is just more readable.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaNames.kt
```diff
@@ -384,9 +382,7 @@ class JavaNames private constructor(
         if (typeDef is TypeFormal) {
             return typeFormal(typeDef.name).toQualName()
         }
-        val connectedKey = TString.unpackOrNull(
-            typeDef.metadata[connectedSymbol]?.firstOrNull(),
-        )
+        val connectedKey = typeDef.connectedKey
```
mikesamuel, line 385: I see TypeDefinition now has a getter to simplify the metadata lookup: qname when connected is present.

be-js/src/commonTest/js/runtime-support/deque-test.mjs
```diff
@@ -1,7 +1,7 @@
 import {
   dequeConstructor,
   dequeAdd,
-  dequeIsEmpty,
+  dequeGetIsEmpty,
```
mikesamuel, line 4: I like avoiding ambiguity between the getter and its property.

be-test-helpers/src/commonMain/kotlin/lang/temper/be/TranslatorTests.kt
```diff
@@ -797,7 +797,7 @@ abstract class TranslatorTests(
                             TmpL.ReturnStatement(
                                 p0,
                                 (
-                                    supportNetwork.translateConnectedReference(p0, "empty", Genre.Library)
+                                    supportNetwork.translateConnectedReference(p0, "core.empty()", Genre.Library)
```
mikesamuel, line 800: Ok. core.empty() is the empty generator step result.

be/src/commonMain/kotlin/lang/temper/be/tmpl/SupportNetwork.kt
```diff
@@ -101,7 +101,7 @@ interface SupportNetwork {
     ): SupportCode? = when (fn) {
         is NamedBuiltinFun -> getSupportCode(pos, fn, genre)
         is CoverFunction -> getSupportCode(pos, fn, genre)
-        is LongLivedUserFunction -> (fn.stayLeaf.incoming?.source as? FunTree)?.parts?.connected?.let { key ->
+        is LongLivedUserFunction -> (fn.stayLeaf.incoming?.source as? FunTree)?.parts?.connectedKey?.let { key ->
```
mikesamuel, line 104: Ok, so this checks is connected and then gets the qname.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLTranslator.kt
mikesamuel: The impending demise of cover functions will get rid of generic equality so hopefully we'll not need these hacks to narrow equality checks in TmpL.  We'll still probably need some way to say "translate this comparison, but what you're really doing is `<` or else we're asking each backend to do some kind of simplification of `!(a < b)` -> `a >= b` kind of stuff.

Then we'll get a form of generic equality back with `provides`.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TranslateDotHelper.kt
mikesamuel: Can we approach this from a backwards-compat point of view?

If a library evolves in two ways with a connected class in two ways:

```ts
// version 0.1
class Complex {
  public real: Float64;
  public imaginary: Float64;

  public get magnitude(): Float64 { (real * real + imaginary * imaginary).sqrt() }
  public get theta: Float64 { /* something something atan2 */ }

  public static fromCoords(real: Float64, imaginary: Float64): Complex { ... }
  public static fromPolar(magnitude: Float64, theta: Float64): Complex { ... }
}
```

If a language wanted to connect this type to one that used either internal representation, could it do that?

If the library changed it's representation so that real and imaginary weren't backed, but were computed from backed properties (magnitude, theta), would that be backwards compatible with connections?

Should assuming connection of a property implies connection of its setter and getter?  Would it help if that were explicit?
Are we missing getter MethodShapes for backed properties pre-TmpL?

tjpalmer: > If the library changed it's representation so that real and imaginary weren't backed, but were computed from backed properties (magnitude, theta), would that be backwards compatible with connections?

If the type is connected as well as the methods, then the internals can be whatever the connector wants. And connecting actual property storage doesn't make any sense to me.

> Should assuming connection of a property implies connection of its setter and getter?

What we had before this PR was you could get one or the other, but not both, depending on its connected key naming. If we want to allow connected full properties, it should likely apply to both its getter and setter (if both exist), and that should involve better frontend processing than we do today. Generated getters/setters would both need qnames and would need metadata copied to both. That all felt out of scope for this PR, so I mostly just have a placeholder of getter support only, which works for the currently needed case.

> Would it help if that were explicit? Are we missing getter MethodShapes for backed properties pre-TmpL?

We could also forbid non-explicit getters/setters for connected things. Which would require more hoop jumping for the Date implementation in Temper.

tjpalmer: > Should assuming connection of a property implies connection of its setter and getter? Would it help if that were explicit?

Turns out this matters sooner than later because of be-cpp. Unlike the other backends, it both wants to connect the getters and keep the properties. So I'll be adjusting things a bit.

mikesamuel (approved):

mikesamuel:
Approving this and related changes.  I think we can get the improvements in and refine with follow-ons later.

I'll keep working on fixing gaps in QName coverage and getter/setter representation in frontend and tmpl in the meantime which should just incrementally improve/simplify things.

PR #443, opened 2026-06-24
Avoid hiding mut classes under `partialImu` supers

tjpalmer:
- Maybe finish imu checking for now
- Test more that we can't hide mut when upcasting interfaces
- Add checks against upcast hiding mut for classes also, which allowing cases where no backed property applies
- Flag type-formal-less partialImu interfaces now too
- Don't worry type-formal-less partialImu if not directly marked

### DIFF
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
@@ -45,7 +45,7 @@ enum class ImuMessage(
     ),
     ExpectedImuType("Expected imu type but got %s"),
     DeepPartialImuTypeDoesNotImplySuperTypeImu(
-        "PartialImu interface %s's type parameter <%s> would not be imu" +
+        "PartialImu type %s's type parameter <%s> would not be imu" +
             " when cast to its effectively Imu super-type %s because %s is not imu",
     ),
     DeepPartialImuTypeCanNeverBeImu(
@@ -119,16 +119,16 @@ class ImuChecker(
             getSuperTypeTree(superType).hasSymbol(partialImuSymbol)
         }
 
-        val presumedImu = buildSet {
-            typeShape.formals.mapNotNullTo(this) { formal ->
-                when {
-                    // All partialImu supers need to reference this formal to be able to presume imu.
-                    // Otherwise, even for classes, we could upcast and hide mutability.
-                    partialImuSupers.all { it.references(formal) } -> MkType2(formal).get()
-                    else -> null
-                }
+        // Split formals into those referenced by all partialImuSupers and those that aren't.
+        // For those that aren't, we could track all non-referencers, but one is good enough.
+        val (presumedImu, novelFormals) = typeShape.formals
+            .map { formal -> MkType2(formal).get() to partialImuSupers.find { !it.references(formal) } }
+            .partition { (_, partialImuSuperWithoutFormal) -> partialImuSuperWithoutFormal == null }
+            .let { (presumedImu, novelFormals) ->
+                buildSet { presumedImu.mapTo(this) { it.first } } to
+                    @Suppress("UNCHECKED_CAST")
+                    (novelFormals.toMap() as Map<TypeParamRef, Type2>)
             }
-        }
         var passes = true
         if (typeShape.formals.isEmpty() && typeShape.metadata.containsKey(partialImuSymbol)) {
             // Types with no type parameters have no reason to be directly tagged partialImu.
@@ -146,6 +146,7 @@ class ImuChecker(
                 typeName,
                 partialImuSymbol,
                 presumedImu = presumedImu,
+                novelFormals = novelFormals,
             )
         ) {
             passes = false
@@ -279,6 +280,8 @@ class ImuChecker(
         typeName: TemperName,
         claimed: Symbol,
         presumedImu: Set<Type2> = emptySet(),
+        /** Formals that could be presumed imu except they're absent in partialImu supertypes. */
+        novelFormals: Map<TypeParamRef, Type2> = emptyMap(),
     ): Boolean {
         var passes = true
         val contravariantTypeParameters = contravariantTypeParametersFor(typeShape)
@@ -293,6 +296,7 @@ class ImuChecker(
                             contravariantTypeParameters,
                             claimed,
                             presumedImu = presumedImu,
+                            novelFormals = novelFormals,
                         )
                     ) {
                         passes = false
@@ -309,6 +313,8 @@ class ImuChecker(
         contravariantTypeParameters: Set<TypeDefinition>,
         claimed: Symbol,
         presumedImu: Set<Type2> = emptySet(),
+        /** Formals that could be presumed imu except they're absent in partialImu supertypes. */
+        novelFormals: Map<TypeParamRef, Type2> = emptyMap(),
     ): Boolean {
         var passes = true
         val propertyName = ParsedName(property.symbol.text)
@@ -330,11 +336,23 @@ class ImuChecker(
                         ?: property.declarationPos
                     logSink.log(ImuMessage.ExpectedImuType, typePos, listOf(nonImuPart))
                 }
-                logSink.log(
-                    ImuMessage.ImuClassPropertyIsNotImu,
-                    property.declarationPos,
-                    listOf(typeName, claimed.text, propertyName, staticType),
-                )
+                when (val hiddenMutSuper = novelFormals[nonImuPart]) {
+                    null -> logSink.log(
+                        ImuMessage.ImuClassPropertyIsNotImu,
+                        property.declarationPos,
+                        listOf(typeName, claimed.text, propertyName, staticType),
+                    )
+                    else -> logSink.log(
+                        ImuMessage.DeepPartialImuTypeDoesNotImplySuperTypeImu,
+                        property.declarationPos,
+                        listOf(
+                            typeName,
+                            nonImuPart.definition.diagnosticTypeName,
+                            hiddenMutSuper,
+                            nonImuPart,
+                        ),
+                    )
+                }
             } else {
                 val contravariantParamUsed = mentionedInType(staticType, contravariantTypeParameters)
                 if (contravariantParamUsed != null) {
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
@@ -130,8 +130,8 @@ class ImuChecker(
             }
         }
         var passes = true
-        if (typeShape.formals.isEmpty() && typeShape.abstractness == Abstractness.Concrete) {
-            // Class with no type parameters has no reason to be PartialImu.
+        if (typeShape.formals.isEmpty() && typeShape.metadata.containsKey(partialImuSymbol)) {
+            // Types with no type parameters have no reason to be directly tagged partialImu.
             passes = false
             logSink.log(
                 ImuMessage.PartialImuTypeHasNoTypeParameters,
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
@@ -19,6 +19,7 @@ import lang.temper.type.WellKnownTypes
 import lang.temper.type2.MkType2
 import lang.temper.type2.SuperTypeTree2
 import lang.temper.type2.Type2
+import lang.temper.type2.TypeParamRef
 import lang.temper.type2.hackMapOldStyleToNew
 import lang.temper.value.DeclTree
 import lang.temper.value.Tree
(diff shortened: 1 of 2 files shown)


frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
```diff
@@ -44,7 +45,7 @@ enum class ImuMessage(
     ),
     ExpectedImuType("Expected imu type but got %s"),
     DeepPartialImuTypeDoesNotImplySuperTypeImu(
-        "PartialImu interface %s's type parameter <%s> would not be imu" +
+        "PartialImu type %s's type parameter <%s> would not be imu" +
             " when cast to its effectively Imu super-type %s because %s is not imu",
```
tjpalmer, line 49: This applies to classes now sometimes also.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
```diff
@@ -111,14 +112,26 @@ class ImuChecker(
...
-        val presumedImu = buildSet {
-            typeShape.formals.mapTo(this) {
-                MkType2(it).get()
-            }
+        // Find the shallowest super-types that are partial imu.
+        val partialImuSupers = typeShape.superTypes.map { superType ->
+            hackMapOldStyleToNew(superType)
+        }.filter { superType ->
+            getSuperTypeTree(superType).hasSymbol(partialImuSymbol)
         }
```
tjpalmer, line 120: Moved up from a lower branch.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
```diff
@@ -111,14 +112,26 @@ class ImuChecker(
...
+        // Split formals into those referenced by all partialImuSupers and those that aren't.
+        // For those that aren't, we could track all non-referencers, but one is good enough.
+        val (presumedImu, novelFormals) = typeShape.formals
+            .map { formal -> MkType2(formal).get() to partialImuSupers.find { !it.references(formal) } }
+            .partition { (_, partialImuSuperWithoutFormal) -> partialImuSuperWithoutFormal == null }
+            .let { (presumedImu, novelFormals) ->
+                buildSet { presumedImu.mapTo(this) { it.first } } to
+                    @Suppress("UNCHECKED_CAST")
+                    (novelFormals.toMap() as Map<TypeParamRef, Type2>)
+            }
```
tjpalmer, line 131: This part mostly applies to classes, not interfaces. As commented above, the gist here is that we can't presume a type formal is imu if it's not also referenced in *all* partialImuSupers. Because of the upcast problem.

Meanwhile, I don't have a proof that being mentioned is good enough, but I think it is. That is, each partialImuSuper gets to presume its type formals are imu, which would require than any type args in the bindings for those formals would also be imu. And if that interface itself has upcast problems, it will be caught other checks already in place before this pr.

Also as mentioned in the comment above, for each novel formal, we want to know some partialImuSuper that fails to it. This is useful for error messaging.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
```diff
@@ -133,26 +146,22 @@ class ImuChecker(
...
+                novelFormals = novelFormals,
             )
         ) {
             passes = false
         }
 
         when (typeShape.abstractness) {
+            // If we require the same for classes as for interfaces, we can't distinguish
+            // backed properties vs computed-only things.
             Abstractness.Concrete -> {}
```
tjpalmer, line 158: I earlier tried just making both classes and interfaces go into the checks below, but for classes, we don't need to be quite as strict. If the novel formal isn't used in a backed property in a class, we're ok. I added a test case for that. So this comment references this matter, and it's why I just trimmed presumedImu for the check above.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
```diff
@@ -322,11 +336,23 @@ class ImuChecker(
...
+                        ImuMessage.DeepPartialImuTypeDoesNotImplySuperTypeImu,
+                        property.declarationPos,
+                        listOf(
+                            typeName,
+                            nonImuPart.definition.diagnosticTypeName,
+                            hiddenMutSuper,
+                            nonImuPart,
+                        ),
+                    )
+                }
```
tjpalmer, line 355: Specializing the error message when it's an upcast problem. And this is why we want to know some partialImuSuper that fails to reference the type formal that could otherwise be presumed imu.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
```diff
@@ -111,14 +112,26 @@ class ImuChecker(
...
+            .let { (presumedImu, novelFormals) ->
+                buildSet { presumedImu.mapTo(this) { it.first } } to
+                    @Suppress("UNCHECKED_CAST")
+                    (novelFormals.toMap() as Map<TypeParamRef, Type2>)
+            }
         var passes = true
-        if (presumedImu.isEmpty() && typeShape.abstractness == Abstractness.Concrete) {
-            // Class with no type parameters has no reason to be PartialImu.
+        if (typeShape.formals.isEmpty() && typeShape.metadata.containsKey(partialImuSymbol)) {
+            // Types with no type parameters have no reason to be directly tagged partialImu.
```
tjpalmer, line 134: This is both broader and narrower than before. It now applies to interfaces as well as classes. But it only applies to directly marked partialImu types. This allows subtyping a partialImu type that had a type param while filling binding it to some fixed imu type. I added a test case for this.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -123,6 +123,10 @@ class ImuCheckerTest {
                 |}
                 |-> Class BadImuContravariant claims imu but property p: List<T__13> uses contravariant type T__13!
                 |
+                |BadPartialImuInterfaceNoTypeArgs:
+                |interface BadPartialImuInterfaceNoTypeArgs {}
+                |-> Type BadPartialImuInterfaceNoTypeArgs claims partialImu but has no type parameters!
```
tjpalmer, line 128: Already had this interface here, but I'm now flagging it as uselessly partialImu.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -157,7 +161,26 @@ class ImuCheckerTest {
...
-                |-> PartialImu interface BadPartialImuViaUpcast's type parameter <T> would not be imu when cast to its effectively Imu super-type PartialImuInterfaceOneArg__20<List<U__46>> because T__45 is not imu!
+                |-> PartialImu type BadPartialImuViaUpcast's type parameter <T> would not be imu when cast to its effectively Imu super-type PartialImuInterfaceOneArg__20<List<U__46>> because T__45 is not imu!
+                |
+                |BadPartialImuViaUpcast2:
+                |interface BadPartialImuViaUpcast2<T, U> extends PartialImuInterfaceOneArg<U> {}
+                |-> PartialImu type BadPartialImuViaUpcast2's type parameter <T> would not be imu when cast to its effectively Imu super-type PartialImuInterfaceOneArg__20<U__49> because T__48 is not imu!
+                |
+                |BadPartialImuViaUpcast3:
+                |interface BadPartialImuViaUpcast3<T> extends BadPartialImuInterfaceNoTypeArgs {}
+                |-> PartialImu type BadPartialImuViaUpcast3's type parameter <T> would not be imu when cast to its effectively Imu super-type BadPartialImuInterfaceNoTypeArgs__16 because T__51 is not imu!
```
tjpalmer, line 172: These three new bad interfaces are just using previously existing checks. But I wanted the extra cases so I could feel more confident the checks were happening.

And notes that BadPartialImuInterfaceNoTypeArgs gets the same checks as any other partialImu interface, even though it's separately flagged as uselessly partialImu.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -157,7 +161,26 @@ class ImuCheckerTest {
...
+                |
+                |BadPartialImuClassViaUpcast:
+                |class BadPartialImuClassViaUpcast<T, U>(
+                |  public prop: T,
+                |  public prop2: U,
+                |) extends PartialImuInterfaceOneArg<List<U>> {}
+                |-> PartialImu type BadPartialImuClassViaUpcast's type parameter <T> would not be imu when cast to its effectively Imu super-type PartialImuInterfaceOneArg__20<List<U__57>> because T__56 is not imu!
```
tjpalmer, line 179: Mostly the same as the original bad upcast interface test case but now as a class.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -157,7 +161,26 @@ class ImuCheckerTest {
...
+                |
+                |BadPartialImuClassViaUpcast2:
+                |class BadPartialImuClassViaUpcast2<T>(public prop: T) extends BadPartialImuInterfaceNoTypeArgs {}
+                |-> PartialImu type BadPartialImuClassViaUpcast2's type parameter <T> would not be imu when cast to its effectively Imu super-type BadPartialImuInterfaceNoTypeArgs__16 because T__59 is not imu!
```
tjpalmer, line 183: And a case with the no args partialImu super.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -206,7 +229,7 @@ class ImuCheckerTest {
             |  public p: List<T>;
             |}
             |
-            |@partialImu interface PartialImuInterfaceNoTypeArgs {}
+            |@partialImu interface BadPartialImuInterfaceNoTypeArgs {}
```
tjpalmer, line 232: This one is checked as bad now.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -235,6 +258,22 @@ class ImuCheckerTest {
             |}
             |
             |interface BadPartialImuViaUpcast<T, U> extends PartialImuInterfaceOneArg<List<U>> {}
+            |interface BadPartialImuViaUpcast2<T, U> extends PartialImuInterfaceOneArg<U> {}
+            |interface BadPartialImuViaUpcast3<T> extends BadPartialImuInterfaceNoTypeArgs {}
+            |interface PartialImuViaUpcast4<@imu T> extends BadPartialImuInterfaceNoTypeArgs {}
```
tjpalmer, line 263: Also checking previous logic, this one is ok because T is imu.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -235,6 +258,22 @@ class ImuCheckerTest {
...
+            |@imu class ImuClassThatWouldBeOk(public prop: BadPartialImuInterfaceNoTypeArgs) {}
+            |class BadPartialImuClassViaUpcast<T, U>(
+            |  public prop: T,
+            |  public prop2: U,
+            |) extends PartialImuInterfaceOneArg<List<U>> {}
+            |class BadPartialImuClassViaUpcast2<T>(public prop: T) extends BadPartialImuInterfaceNoTypeArgs {}
+            |class PartialImuClassViaUpcast<@imu T>(public prop: T) extends BadPartialImuInterfaceNoTypeArgs {}
```
tjpalmer, line 270: Ok because T is imu.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -235,6 +258,22 @@ class ImuCheckerTest {
...
+            |class PartialImuClassViaUpcast2<T> extends BadPartialImuInterfaceNoTypeArgs {
+            |  public hi(thing: T): T { thing }
+            |}
+            |class PartialImuClassViaUpcast3<T> extends ImuDeepInterface {
+            |  public hi(thing: T): T { thing }
+            |}
```
tjpalmer, line 276: One partialImu and one imu super test cases. Both are ok because T isn't in a backed property. For upcast checks on interfaces rather than classes, these would be bad anyway, as shown in other test cases above.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/ImuCheckerTest.kt
```diff
@@ -246,6 +285,7 @@ class ImuCheckerTest {
             |  public p: List<Neither>;
             |}
             |interface BadPartialImuInterfaceHidesParam<T> extends PartialImuInterfaceOneArg<ListBuilder<T>> {}
+            |interface TechnicallyPartialImuInterfaceNoTypeArgs extends PartialImuInterfaceOneArg<String> {}
```
tjpalmer, line 288: And the example of inheriting from a partialImu but having no type formals of its own, which I'm calling ok because it's not directly marked partialImu.

mikesamuel:
Is the problem that we didn't have checks or test coverage before this one?

```
                |BadPartialImuViaUpcast2:
                |interface BadPartialImuViaUpcast2<T, U> extends PartialImuInterfaceOneArg<U> {}
                |-> PartialImu type BadPartialImuViaUpcast2's type parameter <T> would not be imu when cast to its effectively Imu super-type PartialImuInterfaceOneArg__20<U__49> because T__48 is not imu!
```

Does this check from https://hackmd.io/@temper/S1Mn7F7gMx catch that?

Screenshot 2026-06-25 at 4 37 13 pm

tjpalmer:
> Is the problem that we didn't have checks or test coverage before this one?

I think I commented (in pr self review) that you already had good checks on the BadPartialImuViaUpcast interface cases. I just added a few extras to explore a bit more. I didn't carefully vet your existing logic vs what's in the hackmd page, but your existing logic handled everything I threw at it except for classes.

And that's why I added new checks in this pr for classes. Because those *weren't* being checked against upcasts.

mikesamuel (approved):

PR #445, opened 2026-06-26
Avoid connecting properties for now

tjpalmer:


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppSupportNetwork.kt
```diff
@@ -190,9 +190,9 @@ internal object CppSupportNetwork : SupportNetwork {
             put("core.empty()", Like.core("empty"))
             put("core.ignore()", Like.ignoring(theLastArg))
             put("core.type Boolean.toString()", Like.core("Boolean::toString"))
-            put("std/temporal.type Date.day", Like.core("Date::getDay"))
-            put("std/temporal.type Date.month", Like.core("Date::getMonth"))
-            put("std/temporal.type Date.year", Like.core("Date::getYear"))
+            put("std/temporal.type Date.get day()", Like.core("Date::getDay"))
+            put("std/temporal.type Date.get month()", Like.core("Date::getMonth"))
+            put("std/temporal.type Date.get year()", Like.core("Date::getYear"))
```
tjpalmer, line 195: Clearer intent this way.

be-cpp/src/commonMain/resources/lang/temper/be/cpp/core/date.hpp
```diff
@@ -15,31 +15,31 @@ namespace temper {
 
             template<class D>
             int32_t getYear(const std::shared_ptr<D>& d) {
-                return d->year;
+                return d->_year;
             }
```
tjpalmer, line 19: The actual fields have uglier names now.

And be-cpp currently makes public fields. We shouldn't do that. But even if we keep this organization the same, we could `friend` this helper code here for access.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TranslateDotHelper.kt
```diff
@@ -470,7 +470,7 @@ internal object TranslateDotHelper {
 
 internal fun connectedKeyForMember(member: MemberShape): String? = when (member) {
     is MethodShape -> connectedKeyForMethod(member)
-    is PropertyShape -> member.connectedKey
+    is PropertyShape -> null // connect only through accessors methods for now
```
tjpalmer, line 473: Frontend things should go along with this, but that's TODO in this PR. This tmpl restriction at least limits backends from connecting properties themselves until we work more details out in the future.

frontend/src/commonMain/kotlin/lang/temper/frontend/implicits/DateTodayFn.kt
```diff
@@ -39,9 +39,9 @@ class DateTodayFn(val sig: Signature2) : CallableValue, Stayless {
                 val propertyName = propertyShape.name as ModularName
                 val propertyParsedName = (propertyName as ResolvedParsedName).baseName
                 val value = when (propertyParsedName.nameText) {
-                    "year" -> Value(today.year, TInt)
-                    "month" -> Value(today.monthNumber, TInt)
-                    "day" -> Value(today.dayOfMonth, TInt)
+                    "_year" -> Value(today.year, TInt)
+                    "_month" -> Value(today.monthNumber, TInt)
+                    "_day" -> Value(today.dayOfMonth, TInt)
```
tjpalmer, line 44: Interpreter also has to use the private names.

frontend/src/commonMain/resources/std/temporal/temporal.temper.md
```diff
@@ -90,22 +90,25 @@ Here's just enough of a Date type to get us started.
...
+      private _month: Int;
+
       /**
        * The day of the month in [1, 31]
        * additionally constrained by the length of [month].
        */
       @connected
-      public day: Int;
+      public get day(): Int { _day }
+      private _day: Int;
```
tjpalmer, line 111: Now all getters with private still-imu fields. Key is that the connection is specifically on the getters now.

tjpalmer:
Went with #446 instead because even though this one might be formally better, I like the fewer changes for now over there, and after frontend improvements (such as in #451?), some of this changes in this pr might not be needed.

PR #446, opened 2026-06-26
Imply connected getters vs setters

tjpalmer:


be/src/commonMain/kotlin/lang/temper/be/tmpl/TranslateDotHelper.kt
```diff
@@ -470,7 +469,10 @@ internal object TranslateDotHelper {
 
 internal fun connectedKeyForMember(member: MemberShape): String? = when (member) {
     is MethodShape -> connectedKeyForMethod(member)
-    is PropertyShape -> member.connectedKey
+    is PropertyShape -> when {
+        member.hasSetter || member.getter != null -> null // any hint of accessors means no connection
+        else -> member.connectedKey
+    }
```
tjpalmer, line 475: Maybe should also just always say `null` here for now.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TranslateDotHelper.kt
```diff
@@ -483,8 +485,18 @@ private fun connectedKeyForMethod(methodShape: MethodShape): String? {
...
+                propertyShape.hasSetter -> null
+                else -> propertyShape.connectedKey // no reliable `hasGetter`
+            }
+        }
+        MethodKind.Setter -> propertyShapeFor(methodShape)?.let { propertyShape ->
+            when {
+                propertyShape.hasSetter && propertyShape.getter == null -> propertyShape.connectedKey
+                else -> null
+            }
+        }
```
tjpalmer, line 499: A future PR could push qnames and connected tag to generated getter/setters, and we clean this up to just working on those.

PR #447, opened 2026-06-26
frontend: internal representation of method calls skips separate bind step

mikesamuel:
Previously, method calls were represented using separate bind and call steps.
There's something elegant to treating complex calls like the below as
multiple distinct steps.

    subject.method<TypeActual>(arg0, arg1)
    ┣━━━━━┛      ┃           ┃           ┃
    ┃SUBJECT     ┃           ┃           ┃
    ┣━━━━━━━━━━━━┛           ┃           ┃
    ┃  BOUND METHOD          ┃           ┃
    ┣━━━━━━━━━━━━━━━━━━━━━━━━┛           ┃
    ┃    BOUND AND PARAMETERIZED METHOD  ┃
    ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
           COMPLETE METHOD CALL

It was convenient to type each part separately.

- The subject has type `MyClass<Bar>`
- Imagine, the method, pre-binding to a subject, has type
  `<U>(MyClass<T>) -> (U, U) -> Foo<T, U>`,
  so when bound to `MyClass<Bar>` the *T* binds to *Bar* and the overall type
  for the bound method has type `<U>(U, U) -> Foo<Bar, U>`
- Since there are explicit type parameters, the bound and parameterized
  method has type `(TypeActual, TypeActual) -> Foo<Bar, TypeActual>`
- Then the method call has type `Foo<Bar, TypeActual>`.

That only works if you can represent calls to overloaded methods as
one type, an intersection (`&`) type of function types.  Note also
that the type of the bound method is a function type that returns a
function type.  There's currying going on.  Code reasoning about this
has to repeatedly pack and unpack intersection types and curry and
uncurry type info.

Translating intersection types is hard; most typed languages
(TypeScript being an exception) do not allow user's to explicitly
specify intersection types, though many use them under the hood.

Also, reasoning about whole calls is more common than reasoning about the
parts above (again, outside of the Typer), so many parts of the frontend and
the TmpL translator were complicated by the need to look through calls
so that the could reassemble full argument lists (subject and args)
across two calls.

Ongoing changes to the internal representation of types and to type
solving made it easier to sequester overload handling to a small
portion of the *Typer* and *TypeSolver*, obviating the need for
intersection types, and the need for union types (previously used for
nullable type representation and unions of success and failure types
in result types).

This commit reworks *DotHelper*; instead of *BindMemberAccessor*
specifying bind semantics, now *CallMemberAccessor* specifies full
call semantics.

There is a lot of cleanup and reworking of *Typer* internally to
type the new representation, and cleanup of existing code that
produces method calls to produce the newer, flatter calls.

A lot of this reworking of the typer prefigures work that will
be necessary to use *Type2* and *Signature* throughout instead
of the old *StaticType* representations.

Signed-off-by: Mike Samuel <mikesamuel@gmail.com>


be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLTranslator.kt
```diff
@@ -1763,15 +1762,6 @@ class TmpLTranslator internal constructor(
             translateDotHelperCall(
                 tree,
                 callee = effectiveCallee,
-                outerCallTree = null,
```
mikesamuel, line 1756: There is no outer call tree.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TranslateDotHelper.kt
```diff
@@ -191,21 +186,9 @@ internal object TranslateDotHelper {
...
         }
 
-        val declaredCalleeType: Signature2?
-        val actualCalleeType: Signature2?
-        if (outerCallTree == null) {
-            declaredCalleeType = hackTryStaticTypeToSig(callTree.typeInferences?.variant)
-            actualCalleeType = hackTryStaticTypeToSig(callTree.childOrNull(0)?.typeInferences?.type)
-        } else if (dotHelper.memberAccessor is BindMemberAccessor) {
-            val innerCallee = callTree.childOrNull(0)
+        val declaredCalleeType: Signature2? = firstMember?.descriptor
```
mikesamuel, line 185: Still want to store the chosen MemberShape with DotHelper post typing.  That's a PR for a different time though.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TranslateDotHelper.kt
mikesamuel, line 1: Some indenting changed, so recommend viewing in `-w` mode.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TranslateDotHelper.kt
```diff
@@ -191,21 +186,9 @@ internal object TranslateDotHelper {
...
+        val actualCalleeType: Signature2? = hackTryStaticTypeToSig(callTree.childOrNull(0)?.typeInferences?.type)
 
-            actualCalleeType = innerCallee?.typeInferences?.type?.let {
-                hackTryStaticTypeToSig(BindMethodTypeHelper.uncurry(it))
```
mikesamuel, line 199: No need to uncurry anything anymore.

builtin/src/commonMain/kotlin/lang/temper/builtin/RegexLiteralMacro.kt
```diff
@@ -147,11 +147,9 @@ internal object RegexLiteralMacro : BuiltinMacro(regexLiteralBuiltinName.builtin
         // Replace the macro call with regex constructor calls.
         val callPos = macroEnv.pos
         macroEnv.replaceMacroCallWith {
-            Call {
```
mikesamuel, line 150: Lots of this kind of change to code that produces method calls: the outer call wrapper goes away in favour of just the inner call which adopts the extra arguments.

Before `(Call (Call do_bind_foo subject) arg)`
After `(Call do_bind_foo subject arg)`

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -341,7 +344,7 @@ internal class TypeChecker(
         val rightType = rightTree.typeInferences?.type
         // TODO: We probably want to enforce that we have either a left or a right type from the
         // checker, but baby steps.
-        checkSubType(t.pos, leftType, rightType)
+        checkSubType(t, leftType, rightType)
```
mikesamuel, line 347: Makes printf debugging easier.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/BindMethodTypeHelper.kt
mikesamuel, line 1: Did currying and uncurrying but no longer needed.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/FilterOutMaskedMembers.kt
```diff
@@ -21,3 +36,186 @@ internal fun filterOutMaskedMembers(
     }
     membersGrouped.removeMatching { it.value.isEmpty() }
 }
+
+/**
```
mikesamuel, line 40: This comment pretty much explains it, but consolidating some paths in Typer that computed function types for different kinds of calls flushed out some bugs that were load-bearing before.

This is redoing some of the analysis you did in auto-inserting super calls.

There are also more tests in TyperTest now that show this in action.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/InlineToRepairUnrealizedGoals.kt
```diff
@@ -652,15 +653,13 @@ private class InlineToRepairUnrealizedGoals(
 
         // Look for a callee with the form (Call (Call DotHelper(ExternalBind) subject))
         val (dotHelper, thisArg) = run findDotHelper@{
```
mikesamuel, line 655: Here's an example of analysis that used to have to look through multiple method calls that no longer does.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/InlineToRepairUnrealizedGoals.kt
```diff
@@ -196,7 +197,7 @@ private class InlineToRepairUnrealizedGoals(
                 FunArg(
                     // edge indices include callee, but there's also a subject that needs to be curried in,
                     // so we subtract one for the callee and add one back.
-                    argIndex = edgeIndex,
+                    argIndex = edgeIndex - parent.firstArgumentIndex,
```
mikesamuel, line 200: Account for the fact that non-subject args are now part of the same call that has the subject.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/SimplifyDotHelper.kt
```diff
@@ -58,7 +54,7 @@ internal fun simplifyDotHelper(
     var lastNonExtensionResolution: VariantResolution? = null
     var lastResolution: VariantResolution? = null
     for (variant in variants.reversed()) {
-        if (variant.first == variantMatch || variant.first == variantMatchRefined) {
+        if (variant.first equivalent variantMatch || variant.first equivalent variantMatchRefined) {
```
mikesamuel, line 57: I started getting function types here that had argument name info compared to functions without.  That problem will go away though when Signature2's are the thing we're dealing with directly.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -947,35 +946,23 @@ internal class Typer(
...
             }
 
-        val effectiveCalleeType = effectiveCallee?.let { ti.decisionType(it) }?.let { inferredType ->
-            if (isNew) {
-                inferredType // Probably just Type.  Handled below.
-            } else if (inferredType is NominalType && functionalInterfaceSymbol in inferredType.definition.metadata) {
-                inferredType
-            } else { // Extract function types
-                distributeCalleeTypeThroughIntersection(
-                    coerceCalleeTypeToIntersectionOfFunctionTypes(inferredType).also { coerced ->
```
mikesamuel, line 957: More currying and uncurrying adjustments that are no longer necessary.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -813,7 +814,7 @@ internal class Typer(
     }
 
     private fun typeRegularCall(tree: CallTree) {
-        fixupCalleeType(tree)
+        val typedVariants = fixupCalleeType(tree)
```
mikesamuel, line 817: Getting information on variants directly from the horse's mouth.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/Typer.kt
```diff
@@ -2460,51 +2463,93 @@ internal class Typer(
...
+            ?: return TypedVariants(InvalidType, listOf(BecauseMalformedSpecialCall(t.pos, t)))
+        val member = DotMember(memberName)
         val (type, reasons) =
-            typeForGets(t.pos, typeArg.pos, receiverType, memberName, emptyList(), callable)
-        return type to reasons
+            typeForGets(t.pos, typeArg.pos, receiverType, member, emptyList(), callable)
+        return TypedVariants(type, reasons)
+    }
+
+    private fun filterStaticExtensionsByReceiverType(
```
mikesamuel, line 2493: Factored out of typeForGets

frontend/src/commonMain/kotlin/lang/temper/frontend/MaybeAdjustDotHelper.kt
```diff
@@ -190,10 +209,16 @@ internal fun maybeAdjustDotHelper(
...
+                        Call(t.pos) {
+                            V(calleePos, Value(newDotHelper))
+                            dotParts.forEach {
+                                Replant(freeTree(it))
+                            }
+                        }
+                        args.forEach {
                             Replant(freeTree(it))
                         }
                     }
```
mikesamuel, line 224: If we're turning a direct method call into a read of a property with a functional type and then a call of that property, then we actually need to introduce a second call wrapper instead of reusing the second call involved in the bind-then-invoke style.

frontend/src/commonTest/kotlin/lang/temper/frontend/CleanupTemporariesTest.kt
```diff
@@ -122,7 +122,7 @@ class CleanupTemporariesTest {
                 |    t4#0 = t3#0;
                 |    toLogOrNotToLog__0 = t4#0;
                 |    if (toLogOrNotToLog__0) {
-                |      do_bind_log(console#0)("k")
+                |      do_call_log(console#0, "k")
```
mikesamuel, line 125: Lots of this kind of cleanup.  I used the following perl script to automate some of it but a lot of it was by hand.

```perl
use strict;

my $go = 1;

while (<>) {
    if (s/(do_i?)bind_(\w+\([^()]*(?:\(\))?)\)\(\)/$1call_$2)/) {
        print "$.: $_" unless $go;
    } elsif (s/(do_i?)bind_(\w+\([^()]*(?:\(\))?)\)\(/$1call_$2, /) {
        print "$.: $_" unless $go;
    } elsif (!$go && m/do_i?bind/) {
        print "$.: FAIL: $_";
    }
    print if $go;
}

```

frontend/src/commonMain/kotlin/lang/temper/frontend/GeneratorHelperFns.kt
```diff
@@ -212,20 +210,12 @@ object GeneratorStepperFn : CallableValue, StaylessMacroValue {
...
                             interpMode,
                         )
-                        boundMethod.and { boundMethodValue ->
-                            (TFunction.unpackOrNull(boundMethodValue) as? CallableValue)
-                                ?.invoke(ActualValues.Empty, cb, interpMode)
-                                ?: cb.fail(
-                                    MessageTemplate.ExpectedValueOfType,
-                                    values = listOf(TFunction, boundMethodValue),
-                                )
-                        }
```
mikesamuel, line 228: No longer need to get a bound function and then invoke it.  Just invoke the method.

fundamentals/src/commonMain/kotlin/lang/temper/type/DotHelper.kt
mikesamuel, line 1: This is where the main IR change happens.

fundamentals/src/commonMain/kotlin/lang/temper/type/DotHelper.kt
```diff
@@ -124,16 +119,16 @@ class DotHelper(
         }
         val args = macroEnv.args
         val sizeWanted = when (memberAccessor) {
```
mikesamuel, line 121: Since the call operator can now have zero or more args from inside the parentheses in addition to the one subject argument that bind had, the arity check differs.

fundamentals/src/commonMain/kotlin/lang/temper/type/DotHelper.kt
```diff
@@ -261,36 +256,17 @@ class DotHelper(
...
-                        } else {
-                            // TODO: handle call of functions from getter as in
-                            //     obj.f()
-                            // where (obj.f) is a use of a getter than gets a function.
-                            Fail
-                        }
-                    }
-                    else -> TODO("Call of function stored in property")
+                    is MethodShape if (method.methodKind == MethodKind.Normal) ->
+                        dispatchCallTo(method)
```
mikesamuel, line 263: Method dispatch just reuses the same machinery as getter and setter dispatch; they're all methods at the end of the day.

fundamentals/src/commonMain/kotlin/lang/temper/type/MemberShape.kt
```diff
@@ -359,13 +361,19 @@ sealed class ExternalMemberAccessor(
...
+
+/** Reading a property internally. */
 object InternalGet : InternalMemberAccessor("iget"), GetMemberAccessor
 object InternalSet : InternalMemberAccessor("iset"), SetMemberAccessor
-object InternalBind : InternalMemberAccessor("ibind"), BindMemberAccessor
+object InternalCall : InternalMemberAccessor("icall"), CallMemberAccessor
 object ExternalGet : ExternalMemberAccessor("get"), GetMemberAccessor
 object ExternalSet : ExternalMemberAccessor("set"), SetMemberAccessor
-object ExternalBind : ExternalMemberAccessor("bind"), BindMemberAccessor
+object ExternalCall : ExternalMemberAccessor("call"), CallMemberAccessor
```
mikesamuel, line 387: Renamed these by hand instead of using refactoring which broke all existing references and forced me to look at and adjust code that used to deal with bind-style.

fundamentals/src/commonMain/kotlin/lang/temper/type/Types.kt
```diff
@@ -29,6 +29,7 @@ import lang.temper.value.Stayless
 enum class TypeOpPrecedence {
     Or, // Binds loosest
     And,
+    Fn,
```
mikesamuel, line 32: Ran into a minor bug in debug output where, in a nullable function type, the `?` was glomming onto the end of a function-type string representation.

Added tests to StaticTypeTest.

fundamentals/src/commonMain/kotlin/lang/temper/type/Types.kt
```diff
@@ -886,6 +891,27 @@ val StaticType.isBubbly: Boolean get() = when (this) {
     else -> false
 }
 
+val StaticType.nullity: Nullity get() = when (this) {
```
mikesamuel, line 894: Mirrors the nullity member on Type2.

fundamentals/src/commonMain/kotlin/lang/temper/value/PseudoCode.kt
mikesamuel, line 1: Reworked the code that turns DotHelper calls back into `subject.method(...)` syntax when resugar options are on.

I consolidated some explicit and implicit type arg stuff too.

Added tests to PseudoCodeTest.

fundamentals/src/commonTest/kotlin/lang/temper/type/StaticTypeTest.kt
```diff
@@ -27,6 +29,26 @@ class StaticTypeTest {
...
             toStringViaTokenSink { BubbleType.renderTo(it) },
         )
+        val strToVoid = MkType.fn(listOf(), listOf(WellKnownTypes.stringType), null, WellKnownTypes.voidType)
+        assertStringsEqual(
+            "fn (String): Void",
+            toStringViaTokenSink { strToVoid.renderTo(it) },
+        )
+        val strToVoidOrNull = MkType.nullable(strToVoid)
+        assertStringsEqual(
+            "(fn (String): Void)?",
```
mikesamuel, line 39: This was coming out as `fn (String): Void?` which is confusing.

interp/src/commonTest/kotlin/lang/temper/interp/InterpreterTest.kt
```diff
@@ -776,34 +776,6 @@ class InterpreterTest {
         }
     }
 
-    private fun assertResult(
```
mikesamuel, line 779: Moved this down.  I prefer to have all the test helpers either at the top or the bottom.

mikesamuel:
Some breakage in C# and Rust backends which I think are related to tweaks to the declared type derivation in translateDotHelper.

I think the C# side exposes a legit issue, see #448.

The Rust side I think is a problem with this PR.

Going to differentially check declared/actual type with printf debugging in some of the Rust functional tests before and after this change.  Maybe will try to boil it down to a RustBackendTest case first.

mikesamuel:
The Rust failure is in this code from json.temper.md:

```ts
    let decodeHexUnsigned(
      sourceText: String, start: StringIndex, limit: StringIndex
    ): Int {
      var n = 0;
      var i = start;
      while (i.compareTo(limit) < 0) {
        let cp = sourceText[i];
        let digit = if (char'0' <= cp && cp <= char'0') {
          cp - char'0'
        } else if (char'A' <= cp && cp <= char'F') {
          cp - char'A' + 10
        } else if (char'a' <= cp && cp <= char'f') {
          cp - char'a' + 10
        } else {
          return -1;
        }
        n = (n * 16) + digit;
        i = sourceText.next(i);
      }
      n
    }
```

The problem happens in `while (i.compareTo(limit) < 0) {`

My guess is that the problem comes from `i` being types as *StringIndex* but the declared type has the `this` parameter as a `StringIndexOption`.

mikesamuel:
Before these changes:

```
For .compareTo @ test/impl.temper:6+9-27
  declaredCalleeType=(this : StringIndexOption, StringIndexOption) -> Int32
  actualCalleeType=(StringIndex, StringIndexOption) -> Int32
For .get @ test/impl.temper:7+13-26
  declaredCalleeType=(this : String, StringIndex) -> Int32
  actualCalleeType=(String, StringIndex) -> Int32
For .next @ test/impl.temper:18+8-26
  declaredCalleeType=(this : String, StringIndex) -> StringIndex
  actualCalleeType=(String, StringIndex) -> StringIndex
```

After:

```
For .compareTo @ test/impl.temper:6+9-27
  declaredCalleeType=(this : StringIndexOption, StringIndexOption) -> Int32
  actualCalleeType=(StringIndexOption, StringIndexOption) -> Int32
For .get @ test/impl.temper:7+13-26
  declaredCalleeType=(this : String, StringIndex) -> Int32
  actualCalleeType=(String, StringIndex) -> Int32
For .next @ test/impl.temper:18+8-26
  declaredCalleeType=(this : String, StringIndex) -> StringIndex
  actualCalleeType=(String, StringIndex) -> StringIndex
```

Strangely, it's the `this` type for the actual callee type that has changed.

mikesamuel:
Merged in main which fixes C# failure.

be-csharp/src/commonTest/kotlin/lang/temper/be/csharp/CSharpBackendTest.kt
```diff
@@ -476,7 +476,7 @@ class CSharpBackendTest {
                 |static TestGlobal()
                 |{
                 |    builder__0 = new C::OrderedDictionary<string, string>();
-                |    Map1 = C::Mapped.ToMap(builder__0);
+                |    Map1 = C::Mapped.ToMap(C::Mapped.AsReadOnly(builder__0));
```
mikesamuel, line 479: Not sure whether this extra cast here a problem?

docs/for-users/temper-docs/docs/reference/builtins.md
```diff
@@ -2098,7 +2098,7 @@ Unlike in some other languages, curly brackets (`{...}`) are **required** around
 ```temper
 for (var i = 0; i < 3; ++i)
   console.log(i.toString()) // Curly brackets missing
-// ❌ Expected function type, but got Invalid!, No type for subject of .toString!, Expected function type, but got Invalid!
+// ❌ Expected function type, but got Function!, No type for subject of .toString!
```
tjpalmer, line 2101: > Expected function type, but got Function!, No type for subject of .toString!

That "but got Function" is a bit unfortunate, but we can refine later. That second message is helpful. And only two messages instead of three is also nice.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/FilterOutMaskedMembers.kt
tjpalmer: > This is redoing some of the analysis you did in auto-inserting super calls.

Maybe we can clean out some of the later logic at some point.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/FilterOutMaskedMembers.kt
```diff
@@ -21,3 +36,186 @@ internal fun filterOutMaskedMembers(
...
+ *
+ *      interface I<X> { f(): X }
+ *      interface J extends I<String> {}    // J.f has return type String
+ *      interface K { f(): String { "Yep" } }
+ *
+ *      class C extends J, K {}
+ *
+ * Given these two alternative signatures, `fn (K): String` and `fn (I<String>): String`
+ * the type solver has no way to pick between the two leading to an unresolvable overload
+ * decision.
```
tjpalmer, line 52: Or if we choose to error more in the frontend, that would also let us clean back out some from the backends. Since we don't promise semantics when there are frontend errors.

But I guess we probably don't want to error unless we have a manual way to override super selection. Or we could just say "too bad, use composition instead of inheritance here".

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/FilterOutMaskedMembers.kt
```diff
@@ -21,3 +36,186 @@ internal fun filterOutMaskedMembers(
...
+ *
+ * Using those two facts, we can contextualize the descriptors of the f methods to
+ *
+ * - `fn (I<String>): String` from I<K>
+ * - `fn (K): String` from K
+ *
+ * Then we can compare the signatures for equality (ignoring the `this` arg types),
+ * and since they're equivalent, drop any but the shallowest tier ones.
+ */
+internal fun filterOutDeeperMembers(
```
tjpalmer, line 88: I can't tell if inheritance filtering need is increased by this PR or if you just happened to notice it as interesting. And maybe you already explained that elsewhere. I'm ok either way. Just hard to grok everything in review.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/TyperTest.kt
```diff
@@ -1896,6 +1933,112 @@ class TyperTest {
...
+        //
+        // A and C both implement a() but since C is reachable first,
+        // its implementation unambiguously wins.
+        """
+            |    interface A { a(): String { "A wins!" } }
+            |    interface B extends A {}
+            |    interface C { a(): String { "C wins!" } }
+            |    class D extends B & C {}
+            |    new D().a()
+            |/// ┗━━━━━━━━━┛ : String
```
tjpalmer, line 1953: How can we tell which wins from this test? Or did that involve manual inspection? Either way, I'm not worried. Just wondering if I'm missing something.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/TyperTest.kt
```diff
@@ -1896,6 +1933,112 @@ class TyperTest {
...
+            |    interface C { a<T>(x: T): T { x } }
+            |    class D extends B & C {}
+            |    new D().a("hi")
+            |/// ┗━━━━━━━━━━━━━┛ : String
+        """.trimMargin(),
+        skipImplicits = true,
+        // We resolved types just fine, but parameterized methods in interfaces are still suss.
+        wantErrors = listOf(
+            "1+20: Illegal type parameter T. Overridable methods don't allow generics!",
+            "3+20: Illegal type parameter T. Overridable methods don't allow generics!",
```
tjpalmer, line 2009: I wouldn't worry too much about nice handling of type formals that are already illegal.

tjpalmer (approved):
Looks like lots of the logic got simpler, which is nice. Overall, I prefer the new way.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/TyperTest.kt
mikesamuel: It doesn't.  I just copied your style of test.

I'd like to get to the point where we're storing member shapes with dot helpers post typing.

Then, maybe I could add a way to point at a dot and use the comment style to point out which member is stored as a way of checking `@overload @operator @extension` and super-type resolution.

PR #448, opened 2026-06-26
be-csharp: Mapped.toMap was too lenient when subject type was Mapped.

mikesamuel:
The **type Map** functional test checks that `MapBuilder.toMap()` copies defensively.

This adds code to check that `Mapped.toMap` also does.

be-csharp's implementation of those allowed for two subject signatures and optimistically assumed that something typed as IReadOnlyMap was not mutable, but it can be an alias to a reference to a MapBuilder.

### DIFF
--- a/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
+++ b/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
@@ -120,12 +120,12 @@ namespace TemperLang.Core
 
         public static int UShr(this int a, int b)
         {
-            return (int) ((uint) a >> (b & 0x1f));
+            return (int)((uint)a >> (b & 0x1f));
         }
 
         public static long UShr(this long a, int b)
         {
-            return (long) ((ulong) a >> (b & 0x3f));
+            return (long)((ulong)a >> (b & 0x3f));
         }
 
         private const string DIGITS = "0123456789abcdefghijklmnopqrstuvwxyz";
@@ -138,15 +138,17 @@ namespace TemperLang.Core
             if (n < 0)
             {
                 signed = true;
-                un = ((uint) ~n) + 1;
-            } else {
-                un = (uint) n;
+                un = ((uint)~n) + 1;
             }
-            uint uRadix = (uint) radix;
+            else
+            {
+                un = (uint)n;
+            }
+            uint uRadix = (uint)radix;
             // do...while means for 0 we get "0"
             do
             {
-                int digitValue = (int) (un % uRadix);
+                int digitValue = (int)(un % uRadix);
                 un = un / uRadix;
                 sb.Append(DIGITS[digitValue]);
             }
@@ -172,15 +174,17 @@ namespace TemperLang.Core
             if (n < 0)
             {
                 signed = true;
-                un = ((ulong) ~n) + 1;
-            } else {
-                un = (ulong) n;
+                un = ((ulong)~n) + 1;
             }
-            uint uRadix = (uint) radix;
+            else
+            {
+                un = (ulong)n;
+            }
+            uint uRadix = (uint)radix;
             // do...while means for 0 we get "0"
             do
             {
-                int digitValue = (int) (un % uRadix);
+                int digitValue = (int)(un % uRadix);
                 un = un / uRadix;
                 sb.Append(DIGITS[digitValue]);
             }
--- a/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Listed.cs
+++ b/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Listed.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Collections.ObjectModel;
 using System.Linq;
 
@@ -227,11 +228,10 @@ namespace TemperLang.Core
         }
 
         /// <summary>
-        /// Always ensures wrapping in ReadOnlyCollection. If collection is
-        /// already an IList, doesn't create a new one.
+        /// Always ensures converting to an immutable list. If collection is
+        /// already an ImmutableList, doesn't create a new one.
         /// </summary>
         public static IReadOnlyList<T> ToReadOnlyList<T>(this IEnumerable<T> collection) =>
-            collection as ReadOnlyCollection<T>
-            ?? new ReadOnlyCollection<T>(collection as IList<T> ?? new List<T>(collection));
+            collection.ToImmutableList();
     }
 }
--- a/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/OrderedDictionary.cs
+++ b/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/OrderedDictionary.cs
@@ -13,12 +13,24 @@ namespace TemperLang.Core
         /// might also be held via a reference typed as Temper Map<...>.
         public static IReadOnlyDictionary<TKey, TValue> ToMap<TKey, TValue>(
             this IReadOnlyDictionary<TKey, TValue> dictionary
-        ) => dictionary.ToImmutableDictionary();
+        ) => (dictionary is ReadOnlyDictionaryWrapper<TKey, TValue>)
+            ? (ReadOnlyDictionaryWrapper<TKey, TValue>)dictionary
+            : (dictionary is ImmutableDictionary<TKey, TValue>)
+            ? (ImmutableDictionary<TKey, TValue>)dictionary
+            : new ReadOnlyDictionaryWrapper<TKey, TValue>(
+                new OrderedDictionary<TKey, TValue>(dictionary.ToList())
+            );
 
         /// ToMap needs to make a copy if input might be mutated.
         public static IReadOnlyDictionary<TKey, TValue> ToMap<TKey, TValue>(
             this IDictionary<TKey, TValue> dictionary
-        ) => dictionary.ToImmutableDictionary();
+        ) => (dictionary is ReadOnlyDictionaryWrapper<TKey, TValue>)
+            ? (ReadOnlyDictionaryWrapper<TKey, TValue>)dictionary
+            : (dictionary is ImmutableDictionary<TKey, TValue>)
+            ? (ImmutableDictionary<TKey, TValue>)dictionary
+            : new ReadOnlyDictionaryWrapper<TKey, TValue>(
+                new OrderedDictionary<TKey, TValue>(dictionary.ToList())
+            );
 
         /// Duplicated for IDictionary below.
         #region Static Methods for IReadOnlyDictionary
--- a/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/OrderedDictionary.cs
+++ b/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/OrderedDictionary.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 
 namespace TemperLang.Core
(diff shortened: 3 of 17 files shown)


be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/OrderedDictionary.cs
```diff
@@ -7,17 +7,25 @@ namespace TemperLang.Core
...
     {
-        /// ToMap does not need to make a copy, if both are immutable
+        /// ToMap needs to make a copy if input might be aliased via a mutable API
+        /// because IReadOnlyDictionary is the same type as Temper Mapped<...>, so
+        /// might also be held via a reference typed as Temper Map<...>.
         public static IReadOnlyDictionary<TKey, TValue> ToMap<TKey, TValue>(
             this IReadOnlyDictionary<TKey, TValue> dictionary
-        ) => dictionary;
+        ) => (dictionary is ReadOnlyDictionaryWrapper<TKey, TValue>)
+            ? dictionary as ReadOnlyDictionaryWrapper<TKey, TValue>
```
mikesamuel: Maybe should generalize this test to ImmutableDictionary | ImmutableSortedDictionary | FrozenDictionary (.NET 8+).

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/OrderedDictionary.cs
```diff
@@ -7,17 +7,25 @@ namespace TemperLang.Core
...
         public static IReadOnlyDictionary<TKey, TValue> ToMap<TKey, TValue>(
             this IDictionary<TKey, TValue> dictionary
-        ) => new ReadOnlyDictionaryWrapper<TKey, TValue>(
-            new OrderedDictionary<TKey, TValue>(dictionary.ToList())
-        );
+        ) => (dictionary is ReadOnlyDictionaryWrapper<TKey, TValue>)
+            ? dictionary as ReadOnlyDictionaryWrapper<TKey, TValue>
+            : new ReadOnlyDictionaryWrapper<TKey, TValue>(
+                new OrderedDictionary<TKey, TValue>(dictionary.ToList())
+            );
```
mikesamuel: Maybe these two could be collapsed into one.  It seems that IReadOnlyDictionary is disjoint from IDictionary.

be-csharp/src/commonTest/kotlin/lang/temper/be/csharp/CSharpBackendTest.kt
```diff
@@ -455,6 +455,35 @@ class CSharpBackendTest {
...
+            csharp = """
+                |internal static G::IDictionary<string, string> builder__0;
+                |public static G::IReadOnlyDictionary<string, string> Map1;
+                |internal static G::IReadOnlyDictionary<string, string> mapped__0;
+                |public static G::IReadOnlyDictionary<string, string> Map2;
+                |static TestGlobal()
+                |{
+                |    builder__0 = new C::OrderedDictionary<string, string>();
+                |    Map1 = C::Mapped.ToMap(builder__0);
+                |    mapped__0 = C::Mapped.AsReadOnly(builder__0);
```
mikesamuel, line 480: This is just a cast.

be-csharp/src/commonTest/kotlin/lang/temper/be/csharp/CSharpBackendTest.kt
```diff
@@ -455,6 +455,35 @@ class CSharpBackendTest {
...
+                |    Map2 = C::Mapped.ToMap(mapped__0);
```
mikesamuel, line 481: Which caused this ToMap to go to the first variant above which was just identity.

mikesamuel:
AI use disclosure: asked Claude code for info about cheap RTTI checks for immutable maps in C# to see if there are ways to better interop with dictionaries from host program code.

Q:

> Does C#'s standard library have a type for an immutable map that can be tested via `is`?  So given a map, I can do a cheap runtime type check to decide whether to defensively copy it or not?

A:

> - Yes — `System.Collections.Immutable.ImmutableDictionary<TKey, TValue>` is exactly built for this. It's a sealed reference type, so is is a plain type check (no boxing, no interface dispatch overhead worth worrying about):
> - ...
> - There's also `ImmutableSortedDictionary<TKey, TValue>` if you need ordering, and the same pattern applies (is `ImmutableSortedDictionary<TKey, TValue>`).
> - .NET 8+ adds FrozenDictionary<TKey, TValue> (System.Collections.Frozen) — immutable and optimized for fast lookups, but expensive to build, so it's meant for long-lived, read-heavy data rather than as a general defensive-copy target. You can test for it the same way (is FrozenDictionary<TKey, TValue>) if your API might also receive those.
>
> So the idiom is solid: check is `ImmutableDictionary<TKey, TValue>` (and optionally is `FrozenDictionary<TKey, TValue>)`, and only fall back to ToImmutableDictionary() when neither matches.

We do work to preserve key iteration order, so switching to ToImmutableDictionary might not work for us even if it meets our supported version requirements.

mikesamuel:
Realized I forgot to check this functional test on other backends.  Possible this functional test exposes more problems.  Will switch to draft if CI fails.

mikesamuel:
~~TODO: Look into ListBuilder -> Listed -> .toList() as well.~~

Added test code to the list/operations functional test.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
```diff
@@ -120,12 +120,12 @@ public static long DivSafe(this long a, long b)
 
         public static int UShr(this int a, int b)
         {
-            return (int) ((uint) a >> (b & 0x1f));
+            return (int)((uint)a >> (b & 0x1f));
```
mikesamuel, line 123: Ran `dotnet format` in the parent directory.

mikesamuel:
[System.Collections.Immutable](https://www.nuget.org/packages/system.collections.immutable/) is available for:

- .NET 8.0
- .NET Standard 2.0
- .NET Framework 4.6.2

(Screenshot table below)

[C# Backend Supported Versions](https://temperlang.dev/reference/target-languages/#c-backend)

> Targets [.NET](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) 6.0, although [.NET Framework](https://dotnet.microsoft.com/en-us/download/dotnet-framework) 4.8 might be supported in the future.

Afaict, .NET 4.6.2 is compatible with supporting 4.8 in the future, so I focused on checking whether it works with .NET 6.0.

IIuc, this part of *TemperLang.Core.csproj* already addresses that.

https://github.com/temperlang/temper/blob/e8eae0bac13949d684f077109313aca8e52cf409/be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/TemperLang.Core.csproj#L4

----

Screenshot 2026-06-30 at 2 06 27 pm

----

I additionally checked that the `ToImmutable{List,Dictionary}` work with dotnetfiddle.

<details><summary>Screenshot 2026-06-30 at 2 21 35 pm</summary>

```csharp
using System;
using System.Collections.Generic;
using System.Collections.Immutable;
		
public class Program
{
	public static void Main()
	{
		List<string> mutList = new List<string>();
        Dictionary<string, string> mutMap = new Dictionary<string, string>();
        mutList.Add("Before");
        mutMap["Before"] = "Before";

        IReadOnlyList<string> defensiveListCopy = mutList.ToImmutableList();
		IReadOnlyDictionary<string, string> defensiveMapCopy = mutMap.ToImmutableDictionary();

        mutList.Add("Modified after copy");
        mutMap["After"] = "Modified after copy";

        Console.WriteLine($"defensiveListCopy={string.Join(", ", defensiveListCopy)}");
		Console.WriteLine($"defensiveMapCopy={string.Join(", ", defensiveMapCopy.Keys)}");
	}
}
```

</details>

mikesamuel:
Ugh.  ToImmutableDictionary is nice but does not preserve key enumeration order.

mikesamuel:
Fixed

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Listed.cs
```diff
@@ -227,11 +228,10 @@ internal static IComparer<T> ToComparer<T>(this Func<T, T, int> compare)
...
         /// <summary>
-        /// Always ensures wrapping in ReadOnlyCollection. If collection is
-        /// already an IList, doesn't create a new one.
+        /// Always ensures converting to an immutable list. If collection is
+        /// already an ImmutableList, doesn't create a new one.
         /// </summary>
         public static IReadOnlyList<T> ToReadOnlyList<T>(this IEnumerable<T> collection) =>
-            collection as ReadOnlyCollection<T>
-            ?? new ReadOnlyCollection<T>(collection as IList<T> ?? new List<T>(collection));
+            collection.ToImmutableList();
```
tjpalmer, line 235: [Looks like this does avoid copies](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableList.cs#L70) when the type is already an ImmutableList.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/OrderedDictionary.cs
tjpalmer: > It seems that IReadOnlyDictionary is disjoint from IDictionary.

And I missed [a comment that apparently this still *isn't* happening](https://github.com/dotnet/runtime/issues/31001#issuecomment-3999709613) this year. And it might never.

Meanwhile, is your `(dictionary is ReadOnlyDictionaryWrapper<TKey, TValue>) ? (ReadOnlyDictionaryWrapper<TKey, TValue>)dictionary` portion above safe? It seems somewhat similar to the old (my own?) logic in `toList()` that you had to move away from.

functional-test-suite/src/commonMain/resources/types/list/operations/operations.temper.md
```diff
@@ -411,3 +411,30 @@ C\# had a problem with these, trying to make overloads in local variables.
...
+
+    do {
+      let aListBuilder = new ListBuilder<String>();
+      aListBuilder.add("first");
+
+      let builtFirst = aListBuilder.toList();
+
+      aListBuilder.add("second");
+
+      let builtSecond = (aListBuilder as Listed<String>).toList();
```
tjpalmer, line 429: I suppose extra cautious could be an intermediate variable, but I suppose things were failing before relevant backend changes?

functional-test-suite/src/commonMain/resources/types/map/map.temper.md
```diff
@@ -220,17 +220,29 @@ other key = other value
...
 ```log
 --- builder to map (same as above) ---
 key = replaced value
 other key = other value
 added key = added value
+--- builder to map via mapped (same as above) ---
+key = replaced value
+other key = other value
+added key = added value
 ```
```
tjpalmer, line 246: Overall, it's good to have these tests.

tjpalmer (approved):

tjpalmer:
> so I focused on checking whether it works with .NET 6.0.

You mention it needing 8.0 earlier in your comment, but we verified that it does work on 6.0? And we could possibly up min dotnet to some non-EOL sometime if we want.

And maybe the further we go, the less 4.8 matters also? Not sure. (But I also see that you said that it's supposed to work there.)

functional-test-suite/src/commonMain/resources/types/list/operations/operations.temper.md
mikesamuel: Yep.  Each functional test change exposed existing backend failures.

mikesamuel:
> > so I focused on checking whether it works with .NET 6.0.
> 
> You mention it needing 8.0 earlier in your comment, but we verified that it does work on 6.0? And we could possibly up min dotnet to some non-EOL sometime if we want.
> 
> And maybe the further we go, the less 4.8 matters also? Not sure. (But I also see that you said that it's supposed to work there.)

Yeah.  I think it's good to track versions, but don't want to get in the habit of "version X.Y.Z has something that'd be useful for this PR, let's change our stated support."  That kind of thing should be intentional and discussed by a group larger than one PR author and its reviewers.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/OrderedDictionary.cs
mikesamuel: My understanding is that this indicates that a defensive copy happened already so the underlying map is no longer shared.

The difference from the changes to toList is that toList doesn't need to jump through hoops to preserve enumeration order.

PR #449, opened 2026-06-27
Rust chain builder

### LINKED ISSUE #367: Change Rust builders
We currently make builders like this:

```rs
#[derive(Clone, Default)]
pub struct HiOptions {
    pub i: Option<i32>
}

#[derive(Clone)]
pub struct HiBuilder<T: ..., U: ...> {
    pub t: Option<T>, pub u: U
}

impl<T: ..., U: ...> HiBuilder<T, U> {
    pub fn build(self) -> Hi<T, U> {
        self.build_with(std::default::Default::default())
    }
    pub fn build_with(self, options: HiOptions) -> Hi<T, U> {
        Hi::new(self.t, self.u, options.i)
    }
}
```

But we can easily make naming conflicts with the name `...Builder`, even though that's idiomatic. On the other hand, the builder style above *isn't* the most common. I was a bit creative making structs with pub fields. Something like this would be more common:

```rs
#[derive(Clone, Default)]
pub struct HiBuilder<T: ..., U: ...> {
    t: Option<Option<T>>, u: Option<U>, i: Option<i32>
}

impl<T: ..., U: ...> HiBuilder<T, U> {
    pub fn t(mut self, t: Option<T>) -> Self {
        self.t = Some(t);
        self
    }
    pub fn u(mut self, u: U) -> Self {
        self.u = Some(u);
        self
    }
    pub fn i(mut self, i: i32) -> Self {
        self.i = Some(i);
        self
    }
    pub fn build(self) -> Hi<T, U> {
        Hi::new(self.t.unwrap(), self.u.unwrap(), self.i)
    }
}
```

If we coalesce Builder and Options like that (where the abstraction here also helps #271), we can adjust that further to something like this:

```rs
pub impl<T: ..., U: ...> Hi<T, U> {
    pub fn builder() -> HiBuilder<T, U> {
        builders::HiBuilder::default()
    }

    ...
}

pub mod builders {
    #[derive(Clone, Default)]
    pub struct HiBuilder<T: ..., U: ...> {
        t: Option<Option<T>>, u: Option<U>, i: Option<i32>
    }

    impl<T: ..., U: ...> HiBuilder<T, U> {
        pub fn t(mut self, t: Option<T>) -> Self {
            self.t = Some(t);
            self
        }
        pub fn u(mut self, u: U) -> Self {
            self.u = Some(u);
            self
        }
        pub fn i(mut self, i: i32) -> Self {
            self.i = Some(i);
            self
        }
        pub fn build(self) -> Hi<T, U> {
            super::Hi::new(self.t.unwrap(), self.u.unwrap(), self.i)
        }
    }
}
```

Then you access through the static method, such as `Hi::<A, B>::builder().t(...).u(...).build()` or some such. This might be more convenient and idiomatic while reducing risk of collision with Temper naming conventions. But without type state pattern such as in the `bon` crate, which has its own pain, you can easily get runtime errors.

So maybe try this:

```rs
pub impl<T: ..., U: ...> Hi<T, U> {
    ...
}

pub mod builders {
    #[derive(Clone, Default)]
    pub struct Hi<T: ..., U: ...> {
        pub t: Option<T>,
        pub u: U,
    }

    impl<T: ..., U: ...> Hi<T, U> {
        pub fn build(self) -> super::Hi<T, U> {
            self.options().build()
        }
        pub fn options(self) -> HiOptions<T, U> {
            HiOptions::new(self)
        }
    }

    #[derive(Clone)]
    pub struct HiOptions<T: ..., U: ...> {
        hi: Hi,
        i: Option<i32>,
    }

    impl<T: ..., U: ...> HiOptions<T, U> {
        pub new(hi: Hi) -> Self {
            Self { hi, i: None }
        }
        pub fn i(mut self, i: i32) -> Self {
            self.i = Some(i);
            self
        }
        pub fn build(self) -> Hi<T, U> {
            super::Hi::new(self.hi.t, self.hi.u, self.i)
        }
    }
}
```

tjpalmer:
- Fix #367.
- Put builders in `builders` submodule to isolate the generated name better, even though name collisions could still happen so far (such as for a manually named `builders` submodule).
- Fix some issues with generics.
- Use method chaining for optionals as better future proofing.
- I still haven't made a `Whatever::builder()` convenience as recommended in #367, but the most important changes are here, and we can get back to that later if we want. It's another place for potential name collision, though.
- I also considered going with `...::builders::Whatever` instead of `...::builders::WhateverBuilder`, but I figured some people might want to `use` the `builders` directly into scope, and that would be a bit simpler without the namespace collision of saying just `Whatever`.

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
@@ -152,10 +152,11 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
                     append("regex = { version = \"=1.12.2\", optional = true }\n")
                     append("time = { version = \"=0.3.41\", optional = true }\n")
                     append("ureq = { version = \"=3.1.2\", optional = true }\n")
+                    append("zeroize = { version = \"=1.8.2\", optional = true }\n")
                     // Below aren't dependencies section anymore, but eh.
                     append("\n")
                     append("[features]\n")
-                    append("net = [\"ureq\"]\n")
+                    append("net = [\"ureq\", \"zeroize\"]\n")
                     // Implied: append("regex = [\"regex\"]\n")
                     append("temporal = [\"time\"]\n")
                 }
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
@@ -377,13 +377,28 @@ internal fun Rust.FunctionParamOption.toId(): Rust.Id {
     }
 }
 
-internal fun Rust.GenericParam.toArg(): Rust.Id {
+internal fun Iterable<Rust.GenericParam>.except(unwanteds: Iterable<Rust.GenericParam>): List<Rust.GenericParam> {
+    val unwantedIdTexts = unwanteds.mapTo(mutableSetOf()) { it.lastIdText() }
+    return buildList {
+        for (generic in this@except) {
+            if (generic.lastIdText() !in unwantedIdTexts) {
+                add(generic.deepCopy())
+            }
+        }
+    }
+}
+
+internal fun Rust.GenericParam.lastId(): Rust.Id {
     return when (this) {
         is Rust.Id -> this
         is Rust.TypeParam -> id
         is Rust.PathSegments -> segments.last() as Rust.Id
-    }.deepCopy()
+    }
 }
+
+internal fun Rust.GenericParam.lastIdText(): String = lastId().outName.outputNameText
+
+internal fun Rust.GenericParam.toArg(): Rust.Id = lastId().deepCopy()
 
 internal fun Rust.Path.makeTypeRef(generics: List<Rust.GenericParam>): Rust.Type = when {
     generics.isEmpty() -> this
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -96,9 +96,10 @@ class RustTranslator(
         DescriptorsForDeclarations.Key(RustBackend.Factory),
     )?.nameToDescriptor ?: mapOf()
     private var closureCount = 0
+    private val builderItems = mutableListOf<Rust.Item>()
     private val decls = mutableMapOf<ResolvedName, DeclInfo>()
-    private var insideMutableType = false
     private val failVars = mutableSetOf<ResolvedName>()
+    private var insideMutableType = false
     private val functionContextStack = mutableListOf<FunctionContext>()
     private val logSink = LogSink.devNull // TODO what?
     private val loopLabels = mutableListOf<Rust.Id?>()
@@ -179,6 +180,15 @@ class RustTranslator(
                     // Init and declare our own things.
                     add(init)
                     addAll(moduleItems)
+                    // Builders.
+                    if (builderItems.isNotEmpty()) {
+                        builderItems.add(Rust.Use(pos, "super::*".toId(pos)).toItem())
+                        Rust.Module(
+                            pos,
+                            id = "builders".toId(pos), // TODO Uniqueness?
+                            block = Rust.Block(pos, statements = builderItems),
+                        ).toItem().also { add(it) }
+                    }
                     // Tests.
                     if (testItems.isNotEmpty()) {
                         testItems.add(Rust.Use(pos, "super::*".toId(pos)).toItem())
@@ -358,6 +368,37 @@ class RustTranslator(
         }.let { (requireds, optionals) ->
             requireds.map { it.second } to optionals.map { it.second }
         }
+        fun paramsToStructAddedChain(): List<Rust.GenericParam> {
+            val id = builderId.deepCopy() // TODO Stop the deepCopy later.
+            val params = rustParams
+            val attrs = listOf(buildDerive(pos, listOf("Clone", "Default")))
+            val usedTypes = params.findAllSimpleTypeNames()
+            val filteredGenerics = generics.filter { genericReused ->
+                val generic = genericReused.deepCopy() // TODO Stop the deepCopy later.
+                val genericId = when (generic) {
+                    is Rust.Id -> generic
+                    is Rust.TypeParam -> generic.id
+                    else -> error(generic)
+                }.outName.outputNameText
+                genericId in usedTypes
+            }.deepCopy()
+            Rust.Struct(
+                pos = pos,
+                id = id,
+                generics = filteredGenerics,
+                fields = paramPairs.map { (tmpl, paramReused) ->
+                    val param = paramReused.deepCopy() // TODO Stop the deepCopy later.
+                    param as Rust.FunctionParam
+                    val type = when {
+                        tmpl.optional -> param.type
+                        else -> param.type?.deepCopy()?.option()
+                    }
+                    Rust.StructField(param.pos, pub = pub, id = param.pattern as Rust.Id, type = type)
+                },
+            ).also { builderItems.add(it.toItem(attrs = attrs, pub = pub)) }
+            return filteredGenerics
+        }
+        paramsToStructAddedChain()
         fun paramsToStructAdded(
             id: Rust.Id,
             params: List<Rust.FunctionParamOption>,
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -414,8 +414,9 @@ class RustTranslator(
             else -> {
                 // Manage optionals independently so they can default more easily.
                 val attrs = listOf(buildDerive(pos, listOf("Clone")))
+                val selfishParam = Rust.FunctionParam(pos, selfish.deepCopy(), builderId.makeTypeRef(filteredGenerics))
                 val selfishPlusOptionals = buildList {
-                    add(Rust.FunctionParam(pos, selfish.deepCopy(), builderId.makeTypeRef(filteredGenerics)))
+                    add(selfishParam)
                     addAll(optionals)
                 }
                 val optionalGenerics = paramsToStructAdded(optionsId, selfishPlusOptionals, attrs = attrs)
@@ -425,6 +426,26 @@ class RustTranslator(
                     trait = null,
                     type = optionsId.makeTypeRef(optionalGenerics),
                     items = buildList {
+                        Rust.Function(
+                            pos,
+                            id = "new".toId(pos),
+                            params = listOf(selfishParam.deepCopy()),
+                            returnType = "Self".toKeyId(pos),
+                            block = Rust.Block(
+                                pos,
+                                result = Rust.StructExpr(
+                                    pos,
+                                    id = "Self".toKeyId(pos),
+                                    members = buildList {
+                                        add(Rust.StructExprField(pos, selfish.deepCopy(), null))
+                                        for (param in optionals) {
+                                            param is Rust.FunctionParam
+                                            add(Rust.StructExprField(pos, param.toId(), expr = "None".toId(pos)))
+                                        }
+                                    },
+                                ),
+                            ),
+                        ).also { add(it.toItem(pub = pub.deepCopy())) }
                         for (param in optionals) {
                             param as Rust.FunctionParam
                             // Undo optional for setter method params.
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -334,7 +334,7 @@ class RustTranslator(
         returnType: Type2,
     ) {
         // TODO Share this exclusion logic with Java & Js.
-        // If only `this` plus up to 1 more, don't bother with builder. TODO Instead checked named/optional?
+        // If only `this` plus up to 1 more, don't bother with builder.
         fn.parameters.parameters.count { it.name != fn.parameters.thisName } <= 1 && return
         // And for now, skip those with rest parameters. TODO Extract to list value?
         fn.parameters.restParameter != null && return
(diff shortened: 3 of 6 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
```diff
@@ -152,10 +152,11 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
                     append("regex = { version = \"=1.12.2\", optional = true }\n")
                     append("time = { version = \"=0.3.41\", optional = true }\n")
                     append("ureq = { version = \"=3.1.2\", optional = true }\n")
+                    append("zeroize = { version = \"=1.8.2\", optional = true }\n")
                     // Below aren't dependencies section anymore, but eh.
                     append("\n")
                     append("[features]\n")
-                    append("net = [\"ureq\"]\n")
+                    append("net = [\"ureq\", \"zeroize\"]\n")
```
tjpalmer, line 159: I've also fixed this on a different pr that's not yet merged.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -179,6 +180,15 @@ class RustTranslator(
...
                     addAll(moduleItems)
+                    // Builders.
+                    if (builderItems.isNotEmpty()) {
+                        builderItems.add(Rust.Use(pos, "super::*".toId(pos)).toItem())
+                        Rust.Module(
+                            pos,
+                            id = "builders".toId(pos), // TODO Uniqueness?
+                            block = Rust.Block(pos, statements = builderItems),
+                        ).toItem().also { add(it) }
+                    }
```
tjpalmer, line 191: Put builders in a separate submodule to reduce risk of name collision. Still need some way to work around when it there is, though.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -324,7 +334,7 @@ class RustTranslator(
         returnType: Type2,
     ) {
         // TODO Share this exclusion logic with Java & Js.
-        // If only `this` plus up to 1 more, don't bother with builder. TODO Instead checked named/optional?
+        // If only `this` plus up to 1 more, don't bother with builder.
```
tjpalmer, line 337: We don't do named anymore except for property bags.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -343,7 +353,7 @@ class RustTranslator(
         val builderId = "${targetId.outName.outputNameText}Builder".toId(targetId.pos)
         val optionsId = "${targetId.outName.outputNameText}Options".toId(targetId.pos)
         // Figure out if we have requireds and/or optionals.
-        // We need to separate these in Rust because requireds often can't Default.
+        // We need to separate these in Rust because requireds often can't default.
```
tjpalmer, line 356: I'm not making anything Default now, anyway, but the general principle still applies.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -358,10 +368,13 @@ class RustTranslator(
         }.let { (requireds, optionals) ->
             requireds.map { it.second } to optionals.map { it.second }
         }
+        val self = "self".toKeyId(pos)
+        val selfish = "selfish".toKeyId(pos)
```
tjpalmer, line 372: Probably could predefine `"Self"` also, but didn't get around to it. There's also name collision risk for "selfish" here, but I've banked on unlikely for now.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -358,10 +368,13 @@ class RustTranslator(
...
         fun paramsToStructAdded(
             id: Rust.Id,
             params: List<Rust.FunctionParamOption>,
             attrs: List<Rust.AttrOuter> = listOf(),
+            fieldPub: Rust.VisibilityPub? = null,
```
tjpalmer, line 377: Options fields aren't public now.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -380,88 +393,102 @@ class RustTranslator(
...
-                                returnType = rustReturnType,
-                                block = Rust.Block(pos, result = callNew(optionals)),
-                            ).also { add(it.toItem(pub = pub)) }
-                        },
-                    ).also { moduleItems.add(it.toItem()) }
+                val attrs = listOf(buildDerive(pos, listOf("Clone")))
+                val selfishParam = Rust.FunctionParam(pos, selfish.deepCopy(), builderId.makeTypeRef(filteredGenerics))
+                val selfishPlusOptionals = buildList {
+                    add(selfishParam)
+                    addAll(optionals)
```
tjpalmer, line 413: Easier to see backend test cases for the effects of all this.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1242,6 +1233,18 @@ class RustBackendTest {
...
+                |    pub struct CBuilder {
+                |        pub x: std::sync::Arc<String>, pub y: std::sync::Arc<String>
+                |    }
+                |    impl CBuilder {
+                |        pub fn build(self) -> C {
+                |            C::new(self.x, self.y)
+                |        }
+                |    }
+                |    use super::*;
+                |}
```
tjpalmer, line 1247: Here's a `builders` submodule. In this case, both fields are required, so there's just a struct for named init. And it's pretty much the exact same generated code as before in this case except for being in the submodule.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1827,6 +1821,18 @@ class RustBackendTest {
...
+                |    pub struct HaBuilder {
+                |        pub i: i32, pub j: i32
+                |    }
+                |    impl HaBuilder {
+                |        pub fn build(self) -> Ha<T> {
+                |            Ha::new(self.i, self.j)
+                |        }
+                |    }
+                |    use super::*;
+                |}
```
tjpalmer, line 1835: This case was/is showing off needless generics, but it was wrong before, and we don't actually test this case in funtests. Not easy to test right now beyond the backend test. Maybe once we have userspace `@connected`, we can more easily add some backend code to test things like builders and other things we care about.

For now, I've manually tested that this compiles in Rust.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1897,6 +1887,39 @@ class RustBackendTest {
...
                 |temper_core::impl_any_value_trait!(Hi<T, U>, [] where U: std::cmp::Eq + std::hash::Hash);
+                |pub mod builders {
+                |    #[derive(Clone)]
+                |    pub struct HiBuilder<T: Clone + std::marker::Send + std::marker::Sync + 'static, U: std::cmp::Eq + std::hash::Hash + Clone + std::marker::Send + std::marker::Sync + 'static> {
+                |        pub t: Option<T>, pub u: U
+                |    }
+                |    #[derive(Clone)]
+                |    pub struct HiOptions<T: Clone + std::marker::Send + std::marker::Sync + 'static, U: std::cmp::Eq + std::hash::Hash + Clone + std::marker::Send + std::marker::Sync + 'static> {
+                |        selfish: HiBuilder<T, U>, i: Option<i32>
+                |    }
```
tjpalmer, line 1898: Options are different now. My "selfish" thing stores required params, and optionals are additions Option fields here. And the options struct has all the type params of both now.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1897,6 +1887,39 @@ class RustBackendTest {
...
+                |                selfish, i: None
+                |            }
+                |        }
+                |        pub fn i(mut self, i: i32) -> Self {
+                |            self.i = Some(i);
+                |            self
+                |        }
+                |        pub fn build(self) -> Hi<T, U> {
+                |            Hi::new(self.selfish.t, self.selfish.u, self.i)
+                |        }
```
tjpalmer, line 1911: There's now chainer methods for setting optionals. No need to statically check that all are set, because they're optional. And the `new` above means we hide how many fields are here, so new optionals added in the future don't lead to compiler errors.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -470,19 +491,65 @@ class RustTranslator(
...
+        Rust.Impl(
+            pos,
+            generics = requiredGenerics.deepCopy(),
+            trait = null,
+            type = builderId.makeTypeRef(requiredGenerics),
+            items = buildList {
+                Rust.Function(
+                    pos,
+                    id = "build".toId(pos),
+                    generics = generics.except(requiredGenerics),
```
tjpalmer, line 521: The methods need the expected return value generics beyond what's already on the builder type itself.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -377,14 +377,29 @@ internal fun Rust.FunctionParamOption.toId(): Rust.Id {
...
+internal fun Iterable<Rust.GenericParam>.except(unwanteds: Iterable<Rust.GenericParam>): List<Rust.GenericParam> {
+    val unwantedIdTexts = unwanteds.mapTo(mutableSetOf()) { it.lastIdText() }
+    return buildList {
+        for (generic in this@except) {
+            if (generic.lastIdText() !in unwantedIdTexts) {
+                add(generic.deepCopy())
+            }
+        }
+    }
+}
```
tjpalmer, line 389: Used repeatedly now in builder logic.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -377,14 +377,29 @@ internal fun Rust.FunctionParamOption.toId(): Rust.Id {
...
+internal fun Rust.GenericParam.lastId(): Rust.Id {
     return when (this) {
         is Rust.Id -> this
         is Rust.TypeParam -> id
         is Rust.PathSegments -> segments.last() as Rust.Id
-    }.deepCopy()
+    }
 }
 
+internal fun Rust.GenericParam.lastIdText(): String = lastId().outName.outputNameText
```
tjpalmer, line 399: Used in new helper above and also serves as factored logic that was very similar and previously in the builder logic in the translator.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -1897,6 +1887,39 @@ class RustBackendTest {
...
+                |    }
+                |    impl<T: Clone + std::marker::Send + std::marker::Sync + 'static, U: std::cmp::Eq + std::hash::Hash + Clone + std::marker::Send + std::marker::Sync + 'static> HiBuilder<T, U> {
+                |        pub fn build(self) -> Hi<T, U> {
+                |            self.options().build()
+                |        }
+                |        pub fn options(self) -> HiOptions<T, U> {
+                |            HiOptions::new(self)
+                |        }
```
tjpalmer, line 1919: This allows for `...::builders::HiBuilder { ... }.options().i(...).build()` or just `.build()` directly on the builder if no options are wanted.

The main alternative to direct struct fields for compile-time-checked builder pattern in rust is the type state pattern such as used by the popular `bon` crate, but the macros there need to generate a lot more than I generate here.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -2181,6 +2195,41 @@ class RustBackendTest {
...
+            |    impl Vec2Builder {
+            |        pub fn build(self) -> Vec2 {
+            |            self.options().build()
+            |        }
+            |        pub fn options(self) -> Vec2Options {
+            |            Vec2Options::new(self)
+            |        }
+            |    }
+            |    use super::*;
+            |}
```
tjpalmer, line 2232: And this example is optionals-only, and (`...::builders::`) `Vec2Builder {}.build()` or `Vec2Builder {}.options().x(1.5).build()` or variations work, which I think still isn't horrible, even if the primary builder is empty.

functional-test-suite/src/commonMain/resources/classes/object-literals/object-literals.temper.md
```diff
@@ -10,9 +10,9 @@ We infer the constructor by matching up property names.
 
 Define local types, too. We'll import other types syntactically further down.
 
-    class Person(
+    export class Person( // export for exploring generated builders
       public name: String,
-      public age: Int,
+      public age: Int = -100, // mostly just for exploring be-rust codegen
     ) extends Stringable {
```
tjpalmer, line 16: This ensures that the behind-the-scenes builder has both required and optional params. Most in funtests that I found were all optional or all required. So this also ends up at least testing that rust code compiles. No generics here, though.

functional-test-suite/src/commonMain/resources/functions/named-args/named-args.temper.md
```diff
@@ -12,7 +12,7 @@ but named values only work for constructing class instances.
 So make an abusive class that has a constructor only for side effects, so we
 can see what happens.
 
-    class Greet {
+    export class Greet { // export for exploring generated builders
```
tjpalmer, line 15: Across backends, I typically only generate builders for exported types, since temper-built code itself does need the convenience.

mikesamuel (approved):

PR #450, opened 2026-06-29
Special-case upcasting to Listed

tjpalmer:
- Implied upcasting to List was handled else, but now handle explicit also
- Reorganize casting code a bit
- Also fix an unrelated transitive dependency issue
- Nothing here for Mapped, so that might be needed in the future also

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
@@ -152,10 +152,11 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
                     append("regex = { version = \"=1.12.2\", optional = true }\n")
                     append("time = { version = \"=0.3.41\", optional = true }\n")
                     append("ureq = { version = \"=3.1.2\", optional = true }\n")
+                    append("zeroize = { version = \"=1.8.2\", optional = true }\n")
                     // Below aren't dependencies section anymore, but eh.
                     append("\n")
                     append("[features]\n")
-                    append("net = [\"ureq\"]\n")
+                    append("net = [\"ureq\", \"zeroize\"]\n")
                     // Implied: append("regex = [\"regex\"]\n")
                     append("temporal = [\"time\"]\n")
                 }
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -1036,27 +1036,29 @@ class RustTranslator(
         pos: Position,
         found: Description,
         wanted: Description,
-    ): Rust.Path = when {
-        found.definition() == WellKnownTypes.noStringIndexTypeDefinition &&
-            wanted.definition() == WellKnownTypes.stringIndexOptionTypeDefinition
-        -> "temper_core".toKeyId(pos).extendWith(listOf("string", "cast_none_as_index_option"))
-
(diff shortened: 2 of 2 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
```diff
@@ -152,10 +152,11 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
                     append("regex = { version = \"=1.12.2\", optional = true }\n")
                     append("time = { version = \"=0.3.41\", optional = true }\n")
                     append("ureq = { version = \"=3.1.2\", optional = true }\n")
+                    append("zeroize = { version = \"=1.8.2\", optional = true }\n")
                     // Below aren't dependencies section anymore, but eh.
                     append("\n")
                     append("[features]\n")
-                    append("net = [\"ureq\"]\n")
+                    append("net = [\"ureq\", \"zeroize\"]\n")
```
tjpalmer, line 159: Transitive dependency upgrade issue, so peg the version to a compatible one. This is also fixed in #441 and #449, but those aren't merged yet.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1036,27 +1036,29 @@ class RustTranslator(
...
-
-        found.definition() == WellKnownTypes.stringIndexOptionTypeDefinition &&
-            wanted.definition() == WellKnownTypes.noStringIndexTypeDefinition
-        -> "temper_core".toKeyId(pos).extendWith(listOf("string", "cast_as_no_index"))
-
-        else -> {
-            val typeSegment = Rust.GenericArgs(pos, listOf(translateType(wanted.type!!, pos)))
-            // TODO Some way to make fewer intermediate lists?
-            "temper_core".toKeyId(pos).extendWith("cast").extendWith(listOf(typeSegment))
+    ): Rust.Path = when (found.definition()) {
```
tjpalmer, line 1039: Gave a `when` subject because ij wanted me to. I reorganized further while at it.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1036,27 +1036,29 @@ class RustTranslator(
...
+        WellKnownTypes.listTypeDefinition, WellKnownTypes.listBuilderTypeDefinition -> when (wanted.definition()) {
+            WellKnownTypes.listedTypeDefinition -> TO_LISTED_TO_LISTED_NAME.toId(pos)
+            else -> null
         }
```
tjpalmer, line 1043: This case on Listed is what's actually new here. The `AsAnyValue`-based casting doesn't work for specialized Listed and its subtypes.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1825,6 +1827,12 @@ class RustTranslator(
             }
         }.let { result ->
             when {
+                // Upcasting to listed doesn't need checked at all for any valid casts.
+                // Any invalid casts are reported as errors by frontend validation.
+                // But we still need to wrapOk because outer layers with less context still
+                // see the claimed tmpl type as being bubbly.
+                wanted.definition() == WellKnownTypes.listedTypeDefinition -> result.wrapOk()
+                // For others, trust standard type expectations.
```
tjpalmer, line 1835: Slight additional tweak. We get from frontend and tmpl that these casts could throw, so we need accommodate some. The helper `to_listed` method knows that it succeeds every time so isn't typed for potential errors.

PR #451, opened 2026-06-30
generate getters and setters consistently with qnames

mikesamuel:
Trying to see if generating getters and setters (including pureVirtual ones for interfaces) might simplify some things.

Still need to maybe drop pureVirtual bodies to make declaring abstract easier and make it easier for backends that only want getters and setters when they are semantically different from a direct read of a backed property to skip such setters and getters.

### DIFF
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/define/DefineStage.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/define/DefineStage.kt
@@ -11,6 +11,7 @@ import lang.temper.frontend.flipDeclaredNames
 import lang.temper.frontend.interpretiveDanceStage
 import lang.temper.frontend.json.mixinJsonInterop
 import lang.temper.frontend.syntax.DotOperationDesugarer
+import lang.temper.frontend.syntax.fillInMissingQnameMetadata
 import lang.temper.log.Debug
 import lang.temper.log.FailLog
 import lang.temper.log.LogSink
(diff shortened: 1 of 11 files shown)


frontend/src/commonMain/kotlin/lang/temper/frontend/define/DefineStage.kt
```diff
@@ -98,6 +99,14 @@ internal class DefineStage(
                             .snapshot(configKey, AstSnapshotKey, root)
                     }
 
+                    Debug.Frontend.DefineStage.FillInMissingQNames(configKey)
```
mikesamuel, line 102: Just make sure there're qnames on anything that was added since.

frontend/src/commonTest/kotlin/lang/temper/frontend/DefineStageTest.kt
```diff
@@ -2680,7 +2739,7 @@ class DefineStageTest {
             |      T__0 = type (T__0);
             |      I__0<T__0> extends AnyValue;
             |      @property(\x) @visibility(\public) @QName("test-code.type I.x") @stay @fromType(I__0<T__0>) let x__3: T__0;
-            |      @property(\y) @visibility(\public) @QName("test-code.type I.y") @stay @fromType(I__0<T__0>) let y__1;
+            |      @property(\y) @visibility(\public) @stay @fromType(I__0<T__0>) @QName("test-code.type I.y") let y__1: Int32;
```
mikesamuel, line 2742: Now when there's a setter, we infer the type of the abstract property properly.

tjpalmer:
Overall looking good. And I can't tell from a glance, but for interfaces, I think it would be good to generate the exact same internal representation for either of these:

```ts
interface Hi {
  message: String;
  var name: String;
}
```

```ts
interface Hi {
  get message(): String;
  get name(): String;
  set name(name: String);
}
```

Maybe we should prefer one to the other, but from an interface perspective, I think these should mean the same thing.

These would be different for classes, though.

PR #452, opened 2026-06-30
Panic (correctly now?) on missing else for exhaustive sealed when

### LINKED ISSUE #427: when expression with is-Type arms infers Void instead of arm type

tjpalmer:
- Fix #427
- Improved version of #439 that knows if the static type is narrow enough
- If there's a simple chain checking every variant of a sealed type, and if the variable is typed to that sealed type, insert panic instead void
- Don't insert post-chain void on missing `else` because that's harder now, but that affects positions and also makes semicolon more vital
- Also make `truthiness` smarter for simplify control flow, but not really taking advantage of that here

### DIFF
--- a/be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
+++ b/be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
@@ -133,7 +133,7 @@ class PyBackendTest {
             |var m: MapBuilder<String, Int>? = null;
             |for (let e of [1, 2, 3]) {
             |  if (e == 2) {
-            |    m = new MapBuilder<String, Int>()
+            |    m = new MapBuilder<String, Int>();
             |  }
             |}
             |if (m is MapBuilder<String, Int>) {
--- a/cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
+++ b/cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
@@ -636,9 +636,9 @@ class ReplTest {
         repl.processLine("let hi(i: Int): Int { when (i) { 1 -> 1 } }")
         assertPending(
             """
-                |1: t hi(i: Int): Int { when (i) { 1 -> 1 } }
-                |                       ┗━━━━━━━━━━━━━━━━━┛
-                |[interactive#0:1+22-41]@G: Cannot assign to Int32 from Void
+                |1: when (i) { 1 -> 1 } }
+                |                      ⇧
+                |[interactive#0:1+41]@G: Cannot assign to Int32 from Void
                 |interactive#0: void
                 |
             """.trimMargin(),
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
@@ -162,7 +162,11 @@ internal class TypeStage(
         }
 
         // With type info, we can finalize `else` values.
-        replaceVoidishPanics(root)
+        Debug.Frontend.TypeStage.ReplaceVoidishPanics(configKey)
+            .benchmarkIf(BENCHMARK, "ReplaceVoidishPanics") {
+                replaceVoidishPanics(root)
+            }
+        Debug.Frontend.TypeStage.AfterReplaceVoidishPanics.snapshot(configKey, AstSnapshotKey, root)
 
         if (genre != Genre.Documentation) {
             Debug.Frontend.TypeStage.UseBeforeInit(configKey)
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
@@ -1,6 +1,8 @@
 package lang.temper.frontend.typestage
 
 import lang.temper.ast.TreeVisit
+import lang.temper.ast.VisitCue
+import lang.temper.builtin.BuiltinFuns
 import lang.temper.common.Console
 import lang.temper.common.benchmarkIf
 import lang.temper.common.calledFor
@@ -26,11 +28,19 @@ import lang.temper.log.snapshot
 import lang.temper.name.ResolvedName
 import lang.temper.stage.Stage
 import lang.temper.type.InvalidType
+import lang.temper.type.NominalType
 import lang.temper.type.StaticType
+import lang.temper.type.WellKnownTypes
+import lang.temper.value.BasicTypeInferences
 import lang.temper.value.BlockTree
+import lang.temper.value.CallTree
 import lang.temper.value.TBoolean
 import lang.temper.value.Tree
+import lang.temper.value.VoidishPanicFn
+import lang.temper.value.functionContained
+import lang.temper.value.staticTypeContained
 import lang.temper.value.toLispy
+import lang.temper.value.void
 import lang.temper.value.InterpreterCallback.NullInterpreterCallback as nullCallback
 
 private const val BENCHMARK = true
@@ -152,6 +162,9 @@ internal class TypeStage(
             dumpMissingTypeInfo(root, "After typer", console)
         }
 
+        // With type info, we can finalize `else` values.
+        replaceVoidishPanics(root)
+
         if (genre != Genre.Documentation) {
             Debug.Frontend.TypeStage.UseBeforeInit(configKey)
                 .benchmarkIf(BENCHMARK, "UseBeforeInit") {
@@ -251,4 +264,44 @@ private fun dumpMissingTypeInfo(ast: Tree, description: String, console: Console
             }
         }
     }
+}
+
+/**
+ * Replace voidish panics with either void or panic.
+ * TODO Flow typing would be more general than this.
+ */
+private fun replaceVoidishPanics(root: BlockTree) {
+    val expectedTreeSize = VoidishPanicFn.sigs.first().requiredInputTypes.size + 1
+    TreeVisit.startingAt(root).forEach tree@{ tree ->
+        tree is CallTree || return@tree VisitCue.Continue
+        tree.size == expectedTreeSize || return@tree VisitCue.Continue
+        tree.child(0).functionContained == VoidishPanicFn || return@tree VisitCue.Continue
+        // From here down, we know we need either void or standard panic.
+        val incoming = tree.incoming
+        incoming?.replace {
+            run exhaustive@{
+                // Check fail bailing with null goes to void below.
+                val foundType = tree.child(1).typeInferences?.type ?: return@exhaustive null
+                val expectedType = tree.child(2).staticTypeContained ?: return@exhaustive null
+                foundType is NominalType && expectedType is NominalType || return@exhaustive null
+                // We don't support subtyping the same interface under different type actuals,
+                // so simple subtyping is fine here. Any broken type actuals will cause static
+                // type errors elsewhere.
+                isSimpleSubtype(foundType, expectedType) || return@exhaustive null
+                // Passed all checks, so panic it is.
+                Call(tree.typeInferences) { V(BuiltinFuns.vPanic) }
+            } ?: run {
+                // Or bail here to void on check fails.
+                V(void, WellKnownTypes.voidType)
+            }
+        }
+        // Either way, these calls are effectively leaves, so don't bother with kids.
+        VisitCue.SkipOne
+    }.visitPreOrder() // Ok because we never recurse target kids.
+}
+
+/** Just checks for same or subtype, ignoring type actuals. */
+private fun isSimpleSubtype(sub: NominalType, sup: NominalType): Boolean {
+    sub.definition == sup.definition && return true
+    return sub.definition.superTypes.any { isSimpleSubtype(it, sup) }
 }
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
@@ -288,7 +288,7 @@ private fun replaceVoidishPanics(root: BlockTree) {
                 // type errors elsewhere.
                 isSimpleSubtype(foundType, expectedType) || return@exhaustive null
                 // Passed all checks, so panic it is.
-                Call(tree.typeInferences) { V(BuiltinFuns.vPanic) }
+                Call(BuiltinFuns.vPanic, tree.typeInferences) {}
             } ?: run {
                 // Or bail here to void on check fails.
                 V(void, WellKnownTypes.voidType)
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
@@ -31,7 +31,6 @@ import lang.temper.type.InvalidType
 import lang.temper.type.NominalType
 import lang.temper.type.StaticType
 import lang.temper.type.WellKnownTypes
-import lang.temper.value.BasicTypeInferences
 import lang.temper.value.BlockTree
 import lang.temper.value.CallTree
 import lang.temper.value.TBoolean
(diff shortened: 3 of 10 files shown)


be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -133,7 +133,7 @@ class PyBackendTest {
             |var m: MapBuilder<String, Int>? = null;
             |for (let e of [1, 2, 3]) {
             |  if (e == 2) {
-            |    m = new MapBuilder<String, Int>()
+            |    m = new MapBuilder<String, Int>();
             |  }
```
tjpalmer, line 137: Without the semicolon, there was complaint about the `void` in the secret `else` block. Also some new statements below. The change is from no longer throwing a `void` after the `if`/`else if` chain as a whole. More on that later.

But it would be nice at least to improve error messaging for things like missing semicolon here.

cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
```diff
@@ -636,9 +636,9 @@ class ReplTest {
         repl.processLine("let hi(i: Int): Int { when (i) { 1 -> 1 } }")
         assertPending(
             """
-                |1: t hi(i: Int): Int { when (i) { 1 -> 1 } }
-                |                       ┗━━━━━━━━━━━━━━━━━┛
-                |[interactive#0:1+22-41]@G: Cannot assign to Int32 from Void
+                |1: when (i) { 1 -> 1 } }
+                |                      ⇧
+                |[interactive#0:1+41]@G: Cannot assign to Int32 from Void
```
tjpalmer, line 641: This also is from no longer putting the `void` after the `if` chain. The secret else block got and still gets the trailing position. But the trailing void got a larger range. If I put the whole range on the added `else { void }`, then other tests break that expect the trailing pos. I went with trailing pos between the two, so this test had to change.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
```diff
@@ -152,6 +161,9 @@ internal class TypeStage(
             dumpMissingTypeInfo(root, "After typer", console)
         }
 
+        // With type info, we can finalize `else` values.
+        replaceVoidishPanics(root)
```
tjpalmer: Helper just added to this file below because it's small.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
```diff
@@ -252,3 +264,43 @@ private fun dumpMissingTypeInfo(ast: Tree, description: String, console: Console
...
+/**
+ * Replace voidish panics with either void or panic.
+ * TODO Flow typing would be more general than this.
+ */
+private fun replaceVoidishPanics(root: BlockTree) {
+    val expectedTreeSize = VoidishPanicFn.sigs.first().requiredInputTypes.size + 1
+    TreeVisit.startingAt(root).forEach tree@{ tree ->
+        tree is CallTree || return@tree VisitCue.Continue
+        tree.size == expectedTreeSize || return@tree VisitCue.Continue
+        tree.child(0).functionContained == VoidishPanicFn || return@tree VisitCue.Continue
```
tjpalmer, line 281: I more and more like this guard chain style of coding. I hope others don't find it obtuse. To me, it reads as a chain of conditions.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
```diff
@@ -252,3 +264,43 @@ private fun dumpMissingTypeInfo(ast: Tree, description: String, console: Console
...
+        incoming?.replace {
+            run exhaustive@{
+                // Check fail bailing with null goes to void below.
+                val foundType = tree.child(1).typeInferences?.type ?: return@exhaustive null
+                val expectedType = tree.child(2).staticTypeContained ?: return@exhaustive null
+                foundType is NominalType && expectedType is NominalType || return@exhaustive null
+                // We don't support subtyping the same interface under different type actuals,
+                // so simple subtyping is fine here. Any broken type actuals will cause static
+                // type errors elsewhere.
+                isSimpleSubtype(foundType, expectedType) || return@exhaustive null
```
tjpalmer, line 293: This also is a very simple function below. We really don't need all the extra logic from TypeContext or such, if I've thought through it correctly, as mentioned in the comment.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -76,6 +76,92 @@ class GenerateCodeStageTest {
...
+            |export sealed interface Shape extends Geometric {}
+            |export class Circle() extends Shape {}
+            |export class Square() extends Shape {}
+            |export let describeGeometric(g: Geometric): String {
+            |  when (g) {
+            |    is Circle -> "circle";
+            |    is Square -> "square";
+            |    // defaults to void here because it starts above the sealed type
+            |  }
+            |}
```
tjpalmer, line 95: Too broad, so we get void and errors.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -76,6 +76,92 @@ class GenerateCodeStageTest {
...
+            |export let describeShape(s: Shape): String {
+            |  when (s) {
+            |    is Circle -> "circle";
+            |    is Square -> "square";
+            |    // defaults to panic here because those are exhaustive for Shape
+            |  }
+            |}
```
tjpalmer, line 102: Sufficiently narrow, so we get panic and no errors.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -76,6 +76,92 @@ class GenerateCodeStageTest {
...
+            |      constructor__2 = (@stay fn constructor(@impliedThis(Square) this__2: Square) /* return__2 */: Void {
+            |          return__2 = void
+            |      });
+            |      `test//`.describeGeometric = (@stay fn describeGeometric(g__0 /* aka g */: Geometric) /* return__3 */: String {
+            |          if (is(g__0, Circle)) {
+            |            return__3 = "circle"
+            |          } else if (is(g__0, Square)) {
+            |            return__3 = "square"
+            |          } else {
+            |            return__3 = void
```
tjpalmer, line 141: Here's the void.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -76,6 +76,92 @@ class GenerateCodeStageTest {
...
+            |          }
+            |      });
+            |      `test//`.describeShape = (@stay fn describeShape(s__0 /* aka s */: Shape) /* return__4 */: String {
+            |          if (is(s__0, Circle)) {
+            |            return__4 = "circle"
+            |          } else if (is(s__0, Square)) {
+            |            return__4 = "square"
+            |          } else {
+            |            return__4 = panic ⋖ String ⋗()
```
tjpalmer, line 150: Here's the panic.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -76,6 +76,92 @@ class GenerateCodeStageTest {
...
+            |          }
+            |      })
+            |
+            |      ```
+            |  },
+            |  errors: [
+            |    "Cannot assign to String from Void!",
+            |    "Expected subtype of String, but got Void!",
+            |    "Void expressions cannot be used as values!",
```
tjpalmer, line 159: These from the void case. Before fixing things, I got this trio twice, once for each of the functions above. Feel free to review commit history if interested.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -2853,9 +2939,7 @@ class GenerateCodeStageTest {
             |          return__0 = do_get_end(s__0)
             |      });
             |      f__0 = (@stay fn f(s__1 /* aka s */: String) /* return__1 */: Boolean {
-            |          let t#0;
-            |          t#0 = (fn g)(s__1);
-            |          return__1 = is(t#0, NoStringIndex)
+            |          return__1 = is((fn g)(s__1), NoStringIndex)
```
tjpalmer, line 2942: This is from the more thorough `truthiness`, but seems to be the only case we get of improvement from that. This is generate code stage, and I think the later simplify rounds get most cases.

fundamentals/src/commonMain/kotlin/lang/temper/value/NullaryNeverFn.kt
```diff
@@ -44,6 +44,25 @@ object BubbleFn : NullaryNeverFn {
...
+
+    override val callMayFailPerSe: Boolean get() = false
+
+    override fun invoke(args: ActualValues, cb: InterpreterCallback, interpMode: InterpMode) =
+        throw Panic()
+
+    override val sigs = nullaryNeverReturnsSigs(
+        requiredInputTypes = listOf(WKT.anyValueType2, WKT.typeType2),
+    ) { it }
+}
```
tjpalmer, line 64: Put in a reified type for the expected type so I wouldn't have to mess with the expected type formals of nullary functions. But I didn't try it in other forms, so maybe it would have been fine. In any case, this works, too.

fundamentals/src/commonMain/kotlin/lang/temper/value/SimplifyControlFlow.kt
```diff
@@ -179,8 +179,21 @@ fun simplifyControlFlow(
...
+        tree is CallTree && tree.size == 2 || return null
+        val value = tree.child(1).valueContained(TBoolean) ?: return null
+        // Single boolean operand, so dig more at the function.
+        val fn = tree.child(0).functionContained ?: return null
+        fn is CoverFunction && fn.covered.all { covered ->
+            covered is NamedBuiltinFun && covered.builtinOperatorId == BuiltinOperatorId.BooleanNegation
+        } || return null
+        // Yep, got it.
+        return !value
+    }
```
tjpalmer, line 196: I didn't add a type stage test for this, but it gets a fair amount of if/else chain simplification done at type stage with this, because we have lots of `if (!false)` instead of `if (true)` at that point.

I was thinking that I might try some very simple post-typer flow typing, and the simplified chain from this made it look easier to do. I ended up not doing anything with it, but if we theoretically want function macro stage to see simpler code, this helps.

If we want tests of before and after type stage, let me know. Or if we want to remove this adjustment from this pr, also let me know.

interp/src/commonMain/kotlin/lang/temper/interp/IfTransform.kt
```diff
@@ -188,11 +197,24 @@ internal object IfTransform : ControlFlowTransform("if") {
...
+                                        Call {
+                                            V(BuiltinFuns.vVoidishPanic)
+                                            Rn(checkedName)
+                                            V(Value(expectedType))
+                                        }
+                                    }
+                                }.edge(0)
+                                macroCursor.referenceTo(panicCall)
+                            }
+                        }.let { ControlFlow.Stmt(it) }
```
tjpalmer, line 217: This is fairly close to the logic in #439 but refines some. Especially with the "voidish panic" for later cleanup.

fundamentals/src/commonMain/kotlin/lang/temper/value/SimplifyControlFlow.kt
tjpalmer: For reference, here's a manual capture of just after simplify 2 in type stage both before the above change:

```ts
`test//`.describeShape = (@stay fn describeShape(s__13 /* aka s */: Shape) /* return__7 */: String {
    var t#23 ⦂ Boolean, t#24 ⦂ Boolean, t#25 ⦂ String;
    fn__14: do {
      if (!false) {
        t#23 = is(s__13, Circle)
      } else {
        t#23 = false
      };
      if (t#23) {
        t#25 = "circle"
      } else {
        if (!false) {
          t#24 = is(s__13, Square)
        } else {
          t#24 = false
        };
        if (t#24) {
          t#25 = "square"
        } else {
          t#25 = panic ⋖ String ⋗()
        }
      };
      return__7 = t#25
    }
})
```

And after:

```ts
`test//`.describeShape = (@stay fn describeShape(s__13 /* aka s */: Shape) /* return__7 */: String {
    var t#23 ⦂ Boolean, t#24 ⦂ Boolean, t#25 ⦂ String;
    fn__14: do {
      t#23 = is(s__13, Circle);
      if (t#23) {
        t#25 = "circle"
      } else {
        t#24 = is(s__13, Square);
        if (t#24) {
          t#25 = "square"
        } else {
          t#25 = panic ⋖ String ⋗()
        }
      };
      return__7 = t#25
    }
})
```

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
mikesamuel: Yeah.  Not ideal.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
mikesamuel: We've got other micro passes in the benchmark and snapshot pattern.  Could you add an entry to logger-names.json, run `gradle kcodegen:updateGeneratedCode`, and use that pattern here?

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
mikesamuel: It's an idiom I've adapted to.

I'm also used to `break if condition;` and `die unless condition;` style from Perl, so having a one-line idiom for this kind of thing seems reasonable to me.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
```diff
@@ -252,3 +264,43 @@ private fun dumpMissingTypeInfo(ast: Tree, description: String, console: Console
...
+        // Either way, these calls are effectively leaves, so don't bother with kids.
+        VisitCue.SkipOne
+    }.visitPreOrder() // Ok because we never recurse target kids.
+}
+
+/** Just checks for same or subtype, ignoring type actuals. */
+private fun isSimpleSubtype(sub: NominalType, sup: NominalType): Boolean {
+    sub.definition == sup.definition && return true
+    return sub.definition.superTypes.any { isSimpleSubtype(it, sup) }
+}
```
mikesamuel, line 310: Ok, so given a sub from `<T extends Foo>` and sup = Foo, this'll work.

mikesamuel (approved):

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
tjpalmer: > We've got other micro passes in the benchmark and snapshot pattern. Could you add an entry to logger-names.json, run `gradle kcodegen:updateGeneratedCode`, and use that pattern here?

I'll add it. I haven't checked, but I really hope our snapshotter is null for the cli-facing compiler.

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/TypeStage.kt
tjpalmer: > Ok, so given a sub from `<T extends Foo>` and sup = Foo, this'll work.

I've manually checked this now with `export let describeShape<T extends Shape>(s: T): String` instead of `export let describeShape(s: Shape): String`, and it still claims compatibility and provides the panic instead of void, because of how TypeFormal.supertypes works, which I think is correct. But we also get errors like `"Unrelated types cannot be targeted with is or as runtime type checks: \u003c[Circle]\u003e from T__0!"`, which I think also is correct.

I think I won't push the adjusted test, though.

PR #454, opened 2026-07-01
Avoid saying `temper` in temper-built c++

tjpalmer:
- Libraries built by temper should just be libraries in whatever target language
- Stop putting `temper::` namespacing around user libraries
  - If we don't yet have c++ namespace config in config.temper.md, we can add that in the future
- Also say `global_init_...` instead of `temper_init_...` to init user libraries

### DIFF
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
@@ -41,7 +41,7 @@ import lang.temper.name.LanguageLabel
  * ## Initialization
  *
  * All module-level variable initializations and init blocks are deferred to an explicit
- * `temper_init_<module>()` function, called from `main()`, to avoid the Static
+ * `global_init_<module>()` function, called from `main()`, to avoid the Static
  * Initialization Order Fiasco (SIOF). Cross-module dependency init functions are
  * called first within each module's init function.
  *
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
@@ -123,7 +123,7 @@ class CppBackend private constructor(
             allTestInfos.addAll(translator.testInfos)
             translator.moduleInitFuncName?.let { name ->
                 val libNs = safeCppNamespace(cppNames.library(cppLibraryName).text)
-                allInitFuncs.add("temper::${libNs}::$name")
+                allInitFuncs.add("$libNs::$name")
                 // Track include path for this module's header so main.cpp can see all init decls
                 val loc = mod.codeLocation.codeLocation
                 allInitIncludes.add(translator.cpp.includePathForModule(loc))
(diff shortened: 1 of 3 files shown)


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
```diff
@@ -41,7 +41,7 @@ import lang.temper.name.LanguageLabel
  * ## Initialization
  *
  * All module-level variable initializations and init blocks are deferred to an explicit
- * `temper_init_<module>()` function, called from `main()`, to avoid the Static
+ * `global_init_<module>()` function, called from `main()`, to avoid the Static
```
tjpalmer, line 44: We also still need subnamespaces for submodules of libraries. Currently, they all go in the same namespace. I checked by looking at the output of object-literals funtest, but the convention above already shows the expectation of a shared namespace.

Once we split namespaces, maybe this would be just `global_init` or `module_init` in each module's own namespace.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
```diff
@@ -140,7 +140,7 @@ class CppBackend private constructor(
         )
 
         // Compute the C++ namespace for the std library's Test type
-        val testNs = "temper::${safeCppNamespace(cppNames.library("std").text)}"
+        val testNs = safeCppNamespace(cppNames.library("std").text)
```
tjpalmer, line 143: I also looked at the output here, and the namespace we make for std is `temper_std`. So this was saying `temper::temper_std` and now just says `temper_std`. Perhaps we could say just `temper::std` in the future, but this PR is just a first pass on these things.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -26,38 +26,34 @@ class CppBackendTest {
...
             cpp = """
                 |#include <my-test-library/something.hpp>
-                |namespace temper {
-                |  namespace my_test_library {
-                |    static std::shared_ptr<temper::core::Console::Type> console_0;
-                |    static int32_t x;
-                |    void greet(std::string name) {
-                |      temper::core::Console::log(console_0, "Hi:");
-                |      temper::core::Console::log(console_0, name);
+                |namespace my_test_library {
```
tjpalmer, line 29: This is now a namespace I'd be happy with users of my temper-built library using.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -131,29 +123,25 @@ class CppBackendTest {
...
-                |    void temper_init_something() {
-                |      static bool initialized = false;
-                |      if(initialized) {
-                |        return;
-                |      }
-                |      initialized = true;
-                |      x = 42;
+                |namespace my_test_library {
+                |  int32_t x;
+                |  void global_init_something() {
```
tjpalmer, line 128: Because the backend test places this under a submodule called something. In the future, it would be nice to see `my_test_library::something::global_init()` or some such.

mikesamuel:
Agree with the goal.  We shouldn't be "Sent from my iPhone"-ing translations of code that isn't related to the Temper project itself.

The implementation looks fine to me but @ShawSumma would know better.

mikesamuel (approved):

PR #455, opened 2026-07-01
Fix be-rust for call instead of bind

tjpalmer:
- Rely on fixed StringIndexOption compare typing
- Expect more inlining in a backend test

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
@@ -24,9 +24,7 @@ import lang.temper.name.OutName
 import lang.temper.name.ParsedName
 import lang.temper.name.name
 import lang.temper.type.WellKnownTypes
-import lang.temper.type2.DefinedNonNullType
 import lang.temper.type2.DefinedType
-import lang.temper.type2.MkType2
 import lang.temper.type2.Signature2
 import lang.temper.type2.Type2
 import lang.temper.type2.withType
@@ -678,19 +676,8 @@ private object CmpGeneric : MethodCall(
         returnType: Type2,
         translator: RustTranslator,
     ): Rust.Expr {
-        // Work around core.type StringIndexOption.compareTo() typing if needed.
-        val self = arguments.getOrNull(0)
-        val effectiveArgs = when ((self?.type as? DefinedNonNullType)?.definition) {
-            WellKnownTypes.stringIndexTypeDefinition -> buildList {
-                // Left should be an option, not a string index, but we don't get the function type that way.
-                val optionType = MkType2(WellKnownTypes.stringIndexOptionTypeDefinition).get()
-                add(TypedArg((self.expr as Rust.Expr).wrapSome(), optionType))
-                addAll(arguments.subListToEnd(1))
-            }
-            else -> arguments
-        }
-        // And cast result as i32 for be-rust int expectations.
-        return super.inlineToTree(pos, effectiveArgs, returnType, translator).infix(RustOperator.As, "i32".toId(pos))
+        // Cast result as i32 for temper Int type expectations.
+        return super.inlineToTree(pos, arguments, returnType, translator).infix(RustOperator.As, "i32".toId(pos))
     }
 }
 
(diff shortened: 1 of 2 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -678,19 +676,8 @@ private object CmpGeneric : MethodCall(
...
-                val optionType = MkType2(WellKnownTypes.stringIndexOptionTypeDefinition).get()
-                add(TypedArg((self.expr as Rust.Expr).wrapSome(), optionType))
-                addAll(arguments.subListToEnd(1))
-            }
-            else -> arguments
-        }
-        // And cast result as i32 for be-rust int expectations.
-        return super.inlineToTree(pos, effectiveArgs, returnType, translator).infix(RustOperator.As, "i32".toId(pos))
+        // Cast result as i32 for temper Int type expectations.
+        return super.inlineToTree(pos, arguments, returnType, translator).infix(RustOperator.As, "i32".toId(pos))
```
tjpalmer, line 680: Apparently this is a TODONE situation.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -96,28 +96,25 @@ class RustBackendTest {
...
             |                        println!("{}", "Foo");
-            |                        STRINGIFY_VALUE_HERE.set(std::sync::Arc::new(stringifyHere__0.clone())).unwrap_or_else(| _ | panic!());
+            |                        let stringifyValueHere__0: std::sync::Arc<dyn Fn (i32) -> std::sync::Arc<String> + std::marker::Send + std::marker::Sync> = std::sync::Arc::new(stringifyHere__0.clone());
             |                        Ok(())
             |                }).clone()
             |            }
-            |            static STRINGIFY_VALUE_HERE: std::sync::OnceLock<std::sync::Arc<dyn Fn (i32) -> std::sync::Arc<String> + std::marker::Send + std::marker::Sync>> = std::sync::OnceLock::new();
-            |            fn stringify_value_here() -> std::sync::Arc<dyn Fn (i32) -> std::sync::Arc<String> + std::marker::Send + std::marker::Sync> {
-            |                ( * STRINGIFY_VALUE_HERE.get().unwrap()).clone()
-            |            }
```
tjpalmer, line 107: When module top-levels are needed in exported things, I have to put them at top level. Otherwise, they just go in the module init block up above. Hopefully all that is correct there. And this seems to be a case where these imported top-levels aren't needed at all, but any cleanup of that should probably be before it gets to be-rust.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -96,28 +96,25 @@ class RustBackendTest {
...
             |                let nums__0 = nums__0.to_listed();
             |                let a__0: std::sync::Arc<String> = temper_core::listed::join( & ( * nums__0), std::sync::Arc::new("".to_string()), & crate::bar::stringify.clone());
-            |                let b__0: std::sync::Arc<String> = temper_core::listed::join( & ( * nums__0), std::sync::Arc::new("".to_string()), & ( * crate::bar::stringify_value().clone()));
+            |                let b__0: std::sync::Arc<String> = temper_core::listed::join( & ( * nums__0), std::sync::Arc::new("".to_string()), & crate::bar::stringify.clone());
             |                let c__0: std::sync::Arc<String> = temper_core::listed::join( & ( * nums__0), std::sync::Arc::new("".to_string()), & stringifyHere__0.clone());
-            |                let d__0: std::sync::Arc<String> = temper_core::listed::join( & ( * nums__0), std::sync::Arc::new("".to_string()), & ( * stringify_value_here().clone()));
+            |                let d__0: std::sync::Arc<String> = temper_core::listed::join( & ( * nums__0), std::sync::Arc::new("".to_string()), & stringifyHere__0.clone());
```
tjpalmer, line 113: Old code shows `stringify`, `stringify_value`, `stringifyHere__0`, and `stringify_value_here`, but only `stringify` and `stringifyHere__0` show up any more. At a glance to me, this seems to be a case of more inlining than we had before.

mikesamuel (approved):

PR #457, opened 2026-07-08
For java & py, support userspace connected functions in functional tests

tjpalmer:
- Only works in functional tests for now
- Only for java and py so far, one with dir packages and the other with file modules
- Only for top-level functions
- See functional test for examples of things that can already be done
  - Do automate support for optional args
  - Do support access to unexported types
- Allow `@connected` in user code
- Provide `_SomeType` instead of `SomeType_123` naming for unexported types in be-py
- Also the funtest for the conventions I invented so far for java and py


be-java/src/commonMain/kotlin/lang/temper/be/java/JavaBackend.kt
```diff
@@ -152,6 +152,14 @@ class JavaBackend private constructor(
...
             add(result)
         }
+        for (file in rawBackendFiles) {
+            // Just copy raw java files into package dirs.
+            MetadataFileSpecification(
+                path = J.SourceDirectory.MainJava.filePath.resolve(file.key),
+                mimeType = sourceMimeType,
+                content = file.value,
+            ).also { add(it) }
+        }
```
tjpalmer, line 162: I'm unsure that the directory layout is always correct here, but it works for the funtest. I plan to do build testing and/or other such things in future PRs to make sure things are copied properly.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaTranslator.kt
```diff
@@ -153,7 +156,7 @@ class JavaTranslator(
             )
 
         fun module(module: TmpL.Module): List<J.Program> {
-            libraryName = module.libraryName
+            this.module = module
```
tjpalmer, line 159: Easier to ask if the module `isStd` with the full module reference.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaTranslator.kt
```diff
@@ -456,6 +459,32 @@ class JavaTranslator(
             is TmpL.BoilerplateCodeFoldStart -> J.CommentLine(t.pos, t.markerText)
         }
 
+        private fun connectedBody(fn: TmpL.FunctionDeclaration, result: J.ResultType): J.BlockLevelStatement {
+            val pos = fn.pos
+            val defaulting = fn.parameterDefaultStatementsInfo()
+            return buildList {
+                for (statement in defaulting.defaultStatements) {
+                    add(stmt(statement))
+                }
```
tjpalmer, line 468: We keep just the statements through all of the defaulting of optional parameters. I considered both here and in PyTranslator passing in these statements into normal function body translation as an override of just that part, but I've kept them both separate for now.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaTranslator.kt
```diff
@@ -456,6 +459,32 @@ class JavaTranslator(
...
+                                tmpl.optional -> {
+                                    val name = defaulting.parameterMapping.getValue(tmpl.name.name)
+                                    names.lookupRegularLocalNameObj(name).asIdentifier(tmpl.pos)
+                                }
+                                else -> java.name
+                            }.also { add(it.asNameExpr().asArgument()) }
+                        }
+                    },
+                ).exprOrReturnStatement(shouldReturn = result !is J.VoidType).also { add(it) }
+            }.let { J.BlockStatement(pos, it) }
```
tjpalmer, line 485: Here's the forwarding call. I'm still unsure what our rest argument plans are, so I've ignored that here.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaTranslator.kt
```diff
@@ -476,7 +505,7 @@ class JavaTranslator(
                 val overloadFormals = withAdjustments(parameters(overloadParameters), adj)
                 val callForwardArgs: List<J.Argument> = originalFormals.parameters.mapIndexed { ix, formal ->
                     if (ix in overloadFormals.parameters.indices) {
-                        formal.name.deepCopy().asNameExpr()
+                        formal.name.asNameExpr()
```
tjpalmer, line 508: Because `asNameExpr` already deep copies. Just happened to notice this.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaTranslator.kt
```diff
@@ -487,11 +516,15 @@ class JavaTranslator(
...
-            )
             val name = names.moduleFunction(t.name).second.toIdentifier(t.name.pos)
             val result = resultType(names, t.returnType, pos = t.returnType.pos)
+            val body = when {
+                t.metadata.any { it.key.symbol == connectedSymbol } && module?.isStdLib != true ->
+                    connectedBody(t, result)
+                else -> stmt(t.body)
+            }.asBlock(t.pos).preface(
+                translateMetadata(t.pos.leftEdge, t.metadata),
+            )
```
tjpalmer, line 527: Skip normal body if we're connecting.

be-java/src/commonMain/kotlin/lang/temper/be/java/ModuleInfo.kt
```diff
@@ -13,6 +13,10 @@ data class ModuleInfo(
         val classString = lastPart.temperToJavaClass(MODULE_GLOBAL_SUFFIX)
         OutName(classString.assertSafe(), null)
     }
+    val connectedClassName: OutName by lazy {
+        val classString = lastPart.temperToJavaClass(MODULE_CONNECTED_SUFFIX)
+        OutName(classString.assertSafe(), null)
+    }
```
tjpalmer, line 19: Tried to match global class naming style somewhat.

be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
```diff
@@ -539,3 +540,7 @@ BinaryOp.operatorDefinition = `null`;
 UnaryOp(opEnum%`UnaryOpEnum`, expressionOperatorDefinition%`PyOperatorDefinition`);
 UnaryOp.renderTo = `opEnum.renderTo(tokenSink)`;
 UnaryOp.operatorDefinition = `null`;
+
+/** Used for embedding external raw Python source. */
+Raw(source%`String`);
+Raw.renderTo = `tokenSink.value(source)`;
```
tjpalmer, line 546: I need the ability to copy raw python into the generated code, and this seemed like an ok way of doing that.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyBackend.kt
```diff
@@ -301,6 +303,17 @@ class PyBackend private constructor(
...
+            // Special __connected__.py file gets inlined.
+            file.key.last().fullName == "__connected__.py" && continue@rawBackendFiles
+            // Others get copied.
+            // TODO Make sure things get into good places.
+            MetadataFileSpecification(
+                path = file.key,
+                mimeType = mimeType,
+                content = file.value,
+            ).also { outputFileSpecifications.add(it) }
+        }
```
tjpalmer, line 316: This is similar logic to what we have in be-java except for the `__connected__.py` inlining.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyBackend.kt
```diff
@@ -392,7 +405,8 @@ class PyBackend private constructor(
 
         val translations = modules.flatMap { tmpLModule ->
             val trans = this@PyBackend.translator(names, finished.genre)
-            val programs = trans.translate(tmpLModule)
+            val connectedPath = tmpLModule.codeLocation.codeLocation.sourceFile.resolveFile("__connected__.py")
+            val programs = trans.translate(tmpLModule, connectedSource = rawBackendFiles[connectedPath])
```
tjpalmer, line 409: Here's providing that raw source.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyFormattingHints.kt
```diff
@@ -26,6 +27,7 @@ internal object PyFormattingHints : FormattingHints {
         isStarPrefix(preceding) -> false
         isFrom(preceding) && isDots(following) -> true
         isDots(preceding) && isDots(following) -> false
+        preceding.type == OutputTokenType.OtherValue -> false
```
tjpalmer, line 30: Got 1-space indentation otherwise.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -138,9 +139,9 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
         when (name) {
             is Temporary -> chooseSourceName(name, name.nameHint, name.uid, kind, reach)
             is SourceName -> chooseSourceName(name, name.baseName.nameText, name.uid, kind, reach)
-            is BuiltinName -> OutName(styleName(name.builtinKey, kind, reach), sourceName = name)
+            is BuiltinName -> OutName(styleName(name.builtinKey, kind), sourceName = name)
```
tjpalmer, line 142: Reach wasn't being used, so I removed it for now.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -160,21 +161,32 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
...
+        val styledName = styleName(safeIdent(prefix), kind)
         val safeName = when (reach) {
-            TmpL.IdReach.Internal -> concatIfVerbose(styledName, "_$uid") // numeric suffix, won't be a keyword
+            TmpL.IdReach.Internal -> when (kind) {
+                TmpL.IdKind.Type ->
+                    "_$styledName" // won't be keyword if not starting with `_`, which has other issues
+                // Type formals still need suffices for uniqueness right now.
+                TmpL.IdKind.TypeFormal, TmpL.IdKind.Value ->
+                    concatIfVerbose(styledName, "_$uid") // numeric suffix, won't be a keyword
+            }
```
tjpalmer, line 172: We need unexported type names to have reliable naming so connected functions can access them. See the functional test for an example.

I haven't done the same for unexported top-level functions in this PR, but we should do that in the future. The problem is that we generate ParsedName function names sometimes, which can cause conflicts without suffices. I don't know how to distinguish user-named functions from generated names. Maybe should just do helpers generally for translation to use suffices only when collisions happen.

And type formals also get generated in PyTranslator as top-levels with collision risk, which is why I separated out TypeFormal from Type in TmpL.IdKind.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -160,21 +161,32 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
...
             TmpL.IdReach.External -> avoidReserved(styledName)
         }
         return OutName(safeName, sourceName = name)
     }
 
+    /** Supports simple `_whatever` naming until we can arrange it more generally on names. */
+    fun choosePrettyPrivateSourceName(name: ResolvedParsedName, kind: TmpL.IdKind): String {
+        val styledName = styleName(safeIdent(name.baseName.nameText), kind)
+        return "_$styledName"
+    }
```
tjpalmer, line 182: What pretty unexported names should look like. I'm using these names for now as the user connected function names. See more in the functional test.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslator.kt
```diff
@@ -120,15 +123,16 @@ class PyTranslator(
         }
     }
 
-    private var libraryName: DashedIdentifier? = null
+    private var module: TmpL.Module? = null
 
-    fun translate(t: TmpL.Module): List<Py.Program> {
+    fun translate(t: TmpL.Module, connectedSource: String? = null): List<Py.Program> {
+        module = t
```
tjpalmer, line 129: Same reason for this as in JavaTranslator.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslator.kt
```diff
@@ -120,15 +123,16 @@ class PyTranslator(
...
         pyNames.module = t.codeLocation.codeLocation
-        libraryName = t.libraryName
         val result = mutableListOf<Py.Stmt>()
         val tests = mutableListOf<Py.Stmt>()
         val ungroupedImports = mutableListOf<Py.ImportFrom>()
         translateImports(t.imports, ungroupedImports)
+        connectedSource?.also { result.add(Py.Raw(t.pos, it)) }
```
tjpalmer, line 135: Only add if found.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslator.kt
```diff
@@ -1135,6 +1143,41 @@ class PyTranslator(
...
+        renames: List<Py.Stmt>,
+        func: TmpL.FunctionDeclaration,
+        args: Py.Arguments,
+    ): List<Py.Stmt> {
+        // Call the connected function, after substituting any default arg values.
+        // Use a pretty but private name for the connected function.
+        val name = func.name.name as ResolvedParsedName
+        val nameText = pyNames.choosePrettyPrivateSourceName(name, TmpL.IdKind.Value)
+        return buildList {
+            addAll(renames)
```
tjpalmer, line 1156: We need renames here in be-py also.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslator.kt
```diff
@@ -1135,6 +1143,41 @@ class PyTranslator(
...
+            Py.Name(func.pos, PyIdentifierName(nameText)).call(
+                args = buildList {
+                    for ((tmpl, py) in func.parameters.parameters.zip(args.args)) {
+                        when {
+                            tmpl.optional -> name(
+                                pos = tmpl.pos,
+                                name = defaulting.parameterMapping.getValue(tmpl.name.name),
+                            )
+                            else -> py.arg?.asName() // where null might be `/`, so unexpected here
+                        }?.also { add(it) }
```
tjpalmer, line 1172: This also ignores named args, but that should only apply to constructor calls, and I'm only supporting top-level function connection in this PR.

I haven't added error messages for attempting to connect other things. Maybe should.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -262,7 +262,7 @@ class PyBackendTest {
         want = """
             |from temper_std.json import parse_json as parse_json_3, JsonAdapter, JsonProducer, JsonSyntaxTree, InterchangeContext, JsonObject
             |from temper_core import cast_by_type as cast_by_type0
-            |class CJsonAdapter_8(JsonAdapter['C']):
+            |class _CJsonAdapter(JsonAdapter['C']):
```
tjpalmer, line 263: Prettier private type names now.

be-test-helpers/src/commonMain/kotlin/lang/temper/be/FunctionalTestRunner.kt
```diff
@@ -294,6 +294,10 @@ abstract class FunctionalTestRunner<BACKEND : Backend<BACKEND>>(
 
         @Suppress("UNCHECKED_CAST")
         val factory = lookupFactory(backendId) as Backend.Factory<B>
+        val extensions = factory.backendMeta.fileExtensionMap.values.toSet()
+        val rawBackendFiles = test.gatherFiles { filePath ->
+            filePath.lastOrNull()?.extension?.let { it in extensions } == true
+        }
```
tjpalmer, line 300: I automate rawBackendFiles for functional tests only at this point. And each backend only get files by relevant file extensions here. Maybe should get more sophisticated at some point, but this kind of logic might be good enough for all our current backends.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -222,6 +222,84 @@ fun <T> TmpL.CallExpression.mapParameters(
...
+
+/**
+ * Return just the statements needed to provide defaults to optional parameters,
+ * along with a list of all effective parameter names, in order, using the new
+ * default-assigned names where applicable.
+ *
+ * If no optionals, just provide an empty statement list with the original
+ * parameter names.
+ */
+fun TmpL.FunctionDeclaration.parameterDefaultStatementsInfo(): DefaultStatementsInfo {
```
tjpalmer, line 239: This generalized tmpl helper is good enough for both be-java and be-py. Maybe for all the others, too.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -222,6 +222,84 @@ fun <T> TmpL.CallExpression.mapParameters(
...
+fun <R> TmpL.Tree.firstNotNull(transform: (TmpL.Tree) -> R?): R? {
+    fun dig(tree: TmpL.Tree): R? {
+        transform(tree)?.let { return@dig it }
+        for (kid in tree.children) {
+            dig(kid)?.also { return@dig it }
+        }
+        return null
+    }
+    return dig(this)
+}
```
tjpalmer, line 301: Extra helpers that I expected to find variations on already but didn't manage to, so I wrote them up.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLTranslator.kt
```diff
@@ -3506,7 +3505,7 @@ private fun translateDeclarationMetadataSeries(
     // Skip metadata that is available via other avenues
     if (keySymbol in typeMemberMetadataSymbols) { return@mapNotNull null }
     when (keySymbol) {
-        connectedSymbol, fromTypeSymbol, fnSymbol,
+        fromTypeSymbol, fnSymbol,
```
tjpalmer, line 3508: We need to see `@connected` in translators now.

be/src/commonTest/kotlin/lang/temper/be/tmpl/TmpLBackendTest.kt
```diff
@@ -2229,19 +2226,9 @@ class TmpLBackendTest {
...
     }
 
-    private val allowConnectedDecoratorInTestInput =
-        ModuleCustomizeHook { module, isNew ->
-            if (isNew) {
-                module.addEnvironmentBindings(
-                    mapOf(connectedDecoratorName to vConnectedDecorator),
-                )
-            }
-        }
```
tjpalmer, line 2241: We always allow `@connected` is user code now.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -207,6 +208,32 @@ internal class TypeChecker(
         }
     }
 
+    private fun checkAbstractPanicContext(t: CallTree) {
```
tjpalmer, line 212: I did get in validation to require function bodies for top-level functions unless connected. This isn't exactly type checking, but it seemed to be a relatively easy place to add this in. See the stage test for an example of triggering this.

frontend/src/commonMain/kotlin/lang/temper/frontend/implicits/StandardLibraryHelpers.kt
```diff
@@ -58,15 +57,13 @@ fun accessStdWrapped(): FileSystem? {
...
-    when (module.nextStage) {
-        Stage.Lex -> if (module.isEffectivelyImplicits || module.isEffectivelyStd) {
-            module.addEnvironmentBindings(connectedDecoratorBindings)
-        }
-        else -> {}
+    if (module.isEffectivelyImplicits || module.isEffectivelyStd) {
+        // TODO Anything needed here?
+        ignore(module)
     }
 }
```
tjpalmer, line 68: I didn't feel like deleting this entirely yet.

frontend/src/commonMain/kotlin/lang/temper/frontend/syntax/SyntaxMacroStage.kt
```diff
@@ -245,6 +245,20 @@ internal fun rewriteFun(
...
+        call.add(
+            call.treeFarm.grow(call.pos.rightEdge) {
+                Fn {
+                    Block {
+                        Call { V(BuiltinFuns.vAbstractPanic) }
+                    }
+                }
+            },
+        )
+    }
```
tjpalmer, line 260: This provides missing function bodies with an easily identifiable function call marker.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeChecker.kt
```diff
@@ -207,6 +208,32 @@ internal class TypeChecker(
...
+                }
+                parent = parent.incoming?.source
+            }
+            // Not found, so bail.
+            return@checkAbstractPanicContext
+        }
+        // Abstract panic allowed only for connected functions.
+        // Pure virtual methods use a different function so don't arrive here.
+        val parts = fn.parts ?: return
+        parts.connected && return
```
tjpalmer, line 228: When I was using pureVirtual instead of abstractPanic, I also checked for a first implicitThis parameter. But I wasn't sure of an easy way to check function interfaces, so I made the new panic function. We could adjust this internal representation in the future if wanted.

frontend/src/commonTest/kotlin/lang/temper/frontend/GenerateCodeStageTest.kt
```diff
@@ -4047,6 +4042,32 @@ class GenerateCodeStageTest {
...
+            |      @fn @reach(\none) let hi__0;
+            |      hi__0 = (@stay fn hi /* return__0 */: Void {
+            |          abstractPanic();
+            |          return__0 = void
+            |      })
+            |
+            |      ```
+            |  },
+            |  errors: [
+            |    "Function body required except for virtual methods or connected functions!"
```
tjpalmer, line 4065: Error triggers for body-less unconnected above. And before this PR, this would just generate a malformed error node early on, which was less informative than this message.

frontend/src/commonTest/kotlin/lang/temper/frontend/SyntaxMacroStageTest.kt
```diff
@@ -1146,13 +1146,22 @@ class SyntaxMacroStageTest {
...
             "syntaxMacro": {
-                "body": "error (MalformedDeclaration)\n",
+                "body":
+                ```
+                @fn let f__0;
+                f__0 = fn f {
+                  fn__0: do {
+                    abstractPanic()
+                  }
+                };
```
tjpalmer, line 1163: For example, the malformed thing.

frontend/src/commonTest/kotlin/lang/temper/frontend/SyntaxMacroStageTest.kt
```diff
@@ -1166,7 +1175,7 @@ class SyntaxMacroStageTest {
         want = """
         {
             "syntaxMacro": {
-                "body": "error (MalformedDeclaration)\n",
+                "body": "error (MissingName)\n",
```
tjpalmer, line 1179: I'd say this also is better than before.

functional-test-suite/src/commonMain/kotlin/lang/temper/tests/FunctionalTestStatus.kt
```diff
@@ -25,6 +25,7 @@ val functionalTestStatus: Map<Ft, List<IssueCheck>> = buildMap {
     issue(Ft.RegexZeroAdvance, lua(166))
     issue(Ft.NamesNonascii, lua(228))
     214.let { issue(Ft.TypesNetresponse, cpp(it), interp(it), lua(it)) }
+    456.let { issue(Ft.FunctionsConnected, cpp(it), csharp(it), interp(it), js(it), lua(it), rust(it)) }
```
tjpalmer, line 28: Lots more to go, but I think this PR is a good start. C# will be like Java, and other translated backends will be like Python.

Interpreter support will get messy because we need to have a way of saying which backends we expect to be connected once we have bodies for interp. Maybe `@connected(["cpp", "js"])` etc? Or maybe if no rawBackendFiles then do the connection?

functional-test-suite/src/commonMain/resources/functions/connected/__connected__.py
```diff
@@ -0,0 +1,8 @@
+def _sum(i: int, j: int, bonus: int) -> int:
+    return i + j + bonus
```
tjpalmer, line 2: The `bonus` is an optional param, but I get to ignore that here. 🎉

And I considered making these named `_connected__sum` or some such, and maybe we should consider a more obscure naming in the future. But these shouldn't collide, at least, and maybe we can keep them as they are.

functional-test-suite/src/commonMain/resources/functions/connected/_support.py
```diff
@@ -0,0 +1,3 @@
+class Support:
+    def prod(self, i: int, j: int) -> int:
+        return i * j
```
tjpalmer, line 3: No simple way for other python modules to reach back into type definitions in the main one, but they might still be helpful, so they're supported in this PR.

functional-test-suite/src/commonMain/resources/functions/connected/connected.temper.md
```diff
@@ -0,0 +1,67 @@
+# Userspace connected methods
+
+Interfaces support calling backend code from Temper, buy requires a separate
+client project to provide that. It's much more flexible if a Temper library can
+just include and use backend code as needed.
```
tjpalmer, line 5: Probably should also write up a release note when finishing the first round of support for all backends.

functional-test-suite/src/commonMain/resources/functions/connected/connected.temper.md
```diff
@@ -0,0 +1,67 @@
...
+might also still need to generate a wrapper function in the standard location
+for calling the user-provided functiom. Specifics vary by backend.
+
+Also include an optional arg to ensure backends cope, although it's the
+responsibility of the connected code itself to provide the correct default. But,
+for example, overloads sending in `null` for optionals should be automated by
+be-java.
+
+    @connected
+    export let sum(i: Int, j: Int, bonus: Int = 0): Int;
```
tjpalmer, line 20: Here's that optional that we support automatically.

functional-test-suite/src/commonMain/resources/functions/connected/connected.temper.md
```diff
@@ -0,0 +1,67 @@
...
+
+Try both exported above and unexported below to make sure both work. Again,
+managing this varies by backend. This one also uses an unexported type that
+connected code needs to have access to.
+
+    @connected
+    /* unexported */ let prod(hidden: Hidden, j: Int): Int;
+
+    /* unexported */ class Hidden(public i: Int) {}
```
tjpalmer, line 29: And here's purposely using an unexported type. Again, I haven't yet gotten in support for reliable access to unexported top-level functions, which would be good.

Meanwhile, I just noticed that these didn't get pruned. Because they do get used by something exported. And maybe standard handling applied to connecteds is actually ok in this regard.

functional-test-suite/src/commonMain/resources/functions/connected/Support.java
```diff
@@ -0,0 +1,7 @@
+package work;
+
+class Support {
+    int prod(int i, int j) {
+        return i * j;
+    }
+}
```
tjpalmer, line 7: Separate support file still in the same package for java.

functional-test-suite/src/commonMain/resources/functions/connected/WorkConnected.java
```diff
@@ -0,0 +1,11 @@
...
+
+class WorkConnected {
+    static int sum(int i, int j, int bonus) {
+        return i + j + bonus;
+    }
+
+    static int prod(Hidden hidden, int j) {
+        return new Support().prod(hidden.i, j);
+    }
+}
```
tjpalmer, line 11: And here are the actual connected functions.

When tying up initial support for userspace connected functions, we need to document how they work in each backend.

fundamentals/src/commonMain/kotlin/lang/temper/value/NullaryNeverFn.kt
```diff
@@ -86,7 +86,7 @@ object PanicFn : NullaryNeverFn {
 }
 
 /**
- * Marker for bodies of abstract functions that must be overridden in all concrete subtypes of the
+ * Marker for bodies of abstract methods that must be overridden in all concrete subtypes of the
  * containing type.
```
tjpalmer, line 90: Although I almost supported this in non-methods in this PR.

interp/src/commonMain/kotlin/lang/temper/interp/ConnectedDecorator.kt
```diff
@@ -13,9 +12,11 @@ import lang.temper.value.symbolContained
...
 
 /**
- * `@connected methodOrPropertyDefinition...`
- * lets us connect types and type members to native code,
- * where the connection key is defined by the qname.
+ * <!-- snippet: builtin/@connected -->
+ * # `@connected` decorator
+ * Connect types, members, and functions to native code. Each backend defines
+ * specific conventions for connected backend code. Internal backend translation
+ * logic can also key on the QName of connected entities.
```
tjpalmer, line 19: I haven't yet clarified details, including that it only works for userspace on top-level functions. Which I might be ok with even for Temper 1.0. Everything else can be wrapped, as I showed in the funtest.

functional-test-suite/src/commonMain/resources/functions/connected/WorkConnected.java
mikesamuel: The class comment for *JavaBackend* ends up at https://temperlang.dev/reference/target-languages/#java-backend

Maybe it could just include

     ⎀ backend/java/connecting

And then JavaTranslator.ModuleScope.connectBody could have a method comment like the below to explain it.

    /**
     * 
     * ...
     */

functional-test-matrix.md
```diff
@@ -26,6 +26,7 @@
 | [ControlFlowLoopReenterable][] | ✅ | ❌<sup>[198][]</sup> | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [ControlFlowLoops][] | ✅ | ❌<sup>[198][]</sup> | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [FunctionsAsValues][] | ✅ | ❌<sup>[198][]</sup> | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| [FunctionsConnected][] | ❌<sup>[456][]</sup> | ❌<sup>[198][]</sup> | ❌<sup>[456][]</sup> | ❌<sup>[456][]</sup> | ✅ | ✅ | ❌<sup>[456][]</sup> | ❌<sup>[456][]</sup> | ✅ | ✅ | ❌<sup>[456][]</sup> |
```
mikesamuel, line 29: 🎉

functional-test-suite/src/commonMain/resources/functions/connected/Support.java
```diff
@@ -0,0 +1,7 @@
+package work;
```
mikesamuel, line 1: So putting them in the same package allows for package-private use.  And no problems for JAR sealing as long as `temper build -be java` pulls in all the relevant source roots.

Is there a downside to having the `work` directory under `connected` so that resources/functions/connected could be a maven source root?

functional-test-suite/src/commonMain/resources/functions/connected/WorkConnected.java
tjpalmer: Thanks for the documentation tip! I'll plan to come back to this idea when finishing things up in a later PR.

functional-test-suite/src/commonMain/resources/functions/connected/Support.java
tjpalmer: > Is there a downside to having the `work` directory under `connected` so that resources/functions/connected could be a maven source root?

What really will matter is making sure things end up in the right place in real builds. Funtests are a bit wacky sometimes, and I have trouble remembering the details. So I plan to ensure these work for real `temper build` and friends in a future PR.

Separately, this got me thinking about test-only connected functions. I didn't think about that here, but it would be nice if connected functions are only referenced (transitively) from tests that we put the related connected code in the test area also. Maybe we'll need some per-backend convention on that.

be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
mikesamuel: It's always risky if the larger system requires being able to do tree analysis.

Also, do we risk licensing confusion by copying human-authored code into generated code?

Raw is a *Stmt* which iirc can appear indented in a class or function body.

Is it something that can always appear at the beginning or end of a program?

be/src/commonMain/kotlin/lang/temper/be/Backend.kt
```diff
@@ -119,6 +120,7 @@ abstract class Backend<SELF : Backend<SELF>>(
         logSink = setup.logSink,
         config = setup.config,
         dependenciesBuilder = setup.dependenciesBuilder,
+        rawBackendFiles = setup.rawBackendFiles,
```
mikesamuel, line 123: Ok.  So backends get these files, and could do some analysis during the preAnalysis phase.

For example, be-java **could** look for `@TemperConnection("my-library/q.name")` annotations to relate QNames to (package, class, member).

Seems a good use of `setup`.

Should our recompile decisions for `temper watch` based on *SourcePartition* be aware of non-temper source files under `connected` directories?

be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
tjpalmer: > It's always risky if the larger system requires being able to do tree analysis.

Yeah, we'd rely on the humans authoring the python (or whatever other backend) code to write appropriately according to given requirements.

I don't think we should be parsing target language code. And the humans need freedom to do what they need within our requirements. And python needs the code in file for cyclic references to work properly. The java code is effectively the same, but it's code in dir instead of code in file. And I demoed some separate module offload capability in this PR also, people would often want that.

> Also, do we risk licensing confusion by copying human-authored code into generated code?

Presumably, the humans authoring the temper are the humans authoring the python, since they'll be in the same source dir.

> Raw is a _Stmt_ which iirc can appear indented in a class or function body.
> Is it something that can always appear at the beginning or end of a program?

For be-py, I think it would always go at the top. Could be before imports (and I think our requirements for now should to do all imports locally in connected functions that might be needed there), but after imports looks more natural. Current py Program looks like this:

```js
Program ::= body%Stmt*"";
```

We could make a TopLevelStmt that's either Stmt or Raw. Or we could make Program have a separate raw then body if we decide going before imports is ok. (Still probably shouldn't have their own top-level imports since the globals wouldn't be bound in the functions so our own later generated code could break them.) But I'm also ok as is. I personally don't feel a strong need to perfect our out grammar schemas.

be/src/commonMain/kotlin/lang/temper/be/Backend.kt
tjpalmer: > Ok. So backends get these files, and could do some analysis during the preAnalysis phase.
> For example, be-java **could** look for `@TemperConnection("my-library/q.name")` annotations to relate QNames to (package, class, member).

Do you mean on the java code itself? If so, I still think we shouldn't be in the business of parsing target language code. That seems like a tremendous rabbit hole. I think we should instead spec things out such that people can write proper backend code without us needing to process it in any kind of fancy way.

> Should our recompile decisions for `temper watch` based on _SourcePartition_ be aware of non-temper source files under `connected` directories?

There is no "connected" directory in the current PR, and I recommend keeping it that way. You just put the backend code next to temper code. It was only called "connected" in the funtest because the funtest is called "connected".

But beyond that, yes, we'll need `temper watch` to watch backend files in source dirs. That's part of the "this PR is only for funtests" thing.

tjpalmer: And technically all the capability is given for backends to process the backend files, but I don't want to do that ourselves and wouldn't recommend it to others.

functional-test-suite/src/commonMain/resources/functions/connected/Support.java
mikesamuel: Hmm. Are `@reach(\test)` connecteds a thing?
Do you have a use case where I'd want test only code to connect?

mikesamuel (approved):

mikesamuel:
Approving but also bumping https://github.com/temperlang/temper/pull/457#discussion_r3555153334

functional-test-suite/src/commonMain/resources/functions/connected/Support.java
tjpalmer: > Hmm. Are `@reach(\test)` connecteds a thing? Do you have a use case where I'd want test only code to connect?

I think we could already detect that, and since test dependencies are a thing in normal code, I expect some people would want test-only connecteds. My current favorite idea for that is just naming conventions on files like `__test_connected__.py` or `TestWorkConnected.java` or things like that.

tjpalmer:
> Approving but also bumping [#457 (comment)](https://github.com/temperlang/temper/pull/457#discussion_r3555153334)

Was thinking overnight that I might still do the restriction for raw to a single instance at the top of the file in python. Maybe even could call it `Connected` instead of `Raw` to avoid people wanting to abuse it.

People already can use all kinds of outside code dependencies in all kinds of languages. This is basically the same thing, but for cyclic dependencies, especially for access to data types, we need to bring this in file.

But I think I'll skip making the Raw -> Connected changes on this PR since I have other things I need to work on. And we can think about the topic meanwhile.

PR #462, opened 2026-07-10
frontend: Module.mayRun complicated predictive execution

mikesamuel:
This will eventually simplify executing a method predictively, because you optimistically believe it's side effect free, so can be reordered to compile time.

Those optimistic assumptions are something you can check in an instrumented interpreter.

And predictive execution benefits the secure composition library because there we're predicting the result of context propagation to pick escapers for untrusted inputs based on applying functions to string literals.

Predictive execution was complicated because we were tagging modules as runnable or translatable.
The idea behind that was that modules we want to translate we should not run, because running should change the AST or metadata like exported values.

This commit does two things:

- Changes *Interpreter* to panic if anything done at Stage.Run tries to use the `MacroEnvironment.replace` APIs.
- Changes *Export* so that separate values are kept for staging (when the result is statically determinable) and running (when the result may depend on sources of nondeteriminism only available at runtime).

Further commits will build on it to allow more flexible predictive executaion, by allowing a module to request that its dependencies be advanced to the run stage, and to allow executing a synthesized expression as if in the run stage.

### DIFF
--- a/common/src/commonMain/kotlin/lang/temper/common/Collections.kt
+++ b/common/src/commonMain/kotlin/lang/temper/common/Collections.kt
@@ -328,9 +328,26 @@ fun <K, V> partiallyOrder(
     items: Iterable<V>,
     afterMap: Map<K, Collection<V>>,
     depKeyOf: (V) -> K,
-): List<V> {
+): List<V> = buildList {
+    partiallyOrderTo(items, afterMap, this, depKeyOf)
+}
+
+/**
+ * A list containing [items] and values from [afterMap] such that each, v, appears once after all of
+ * [afterMap]\[k].
+ *
+ * If there is a cycle in [afterMap]'s transitive closure it is broken arbitrarily.
+ *
+ * If there is a v in afterMap\[k] such that k is a dep key for an output element, then v will end
+ * up in the partial order regardless of whether v in items.
+ */
+fun <K, V> partiallyOrderTo(
+    items: Iterable<V>,
+    afterMap: Map<K, Collection<V>>,
+    partialOrder: MutableCollection<V>,
+    depKeyOf: (V) -> K,
+) {
     val ordered = mutableSetOf<V>()
-    val partialOrder = mutableListOf<V>()
     fun addToOrdered(v: V) {
         if (v !in ordered) {
             ordered.add(v) // Arbitrarily breaks cycles.
@@ -347,7 +364,6 @@ fun <K, V> partiallyOrder(
     for (v in items) {
         addToOrdered(v)
     }
-    return partialOrder.toList()
 }
 
 fun intersect(a: IntRange, b: IntRange): IntRange {
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/FoundAfterModuleStaging.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/FoundAfterModuleStaging.kt
@@ -47,6 +47,14 @@ internal fun findExportsAndDeclaredTypes(
     val declaredTypes = mutableSetOf<TypeShape>()
     var topLevelMetadataDecl: DeclTree? = null
 
+    val exportsBefore by lazy {
+        buildMap {
+            for (export in (exporter.exports ?: listOf())) {
+                this[export.name] = export
+            }
+        }
+    }
+
     // If we have exports, try to relate them to type inferences.
     val declInfoForExports = mutableMapOf<ExportedName, DeclInfo>()
     if (exportedNames.isNotEmpty() && stage >= Stage.Type) {
@@ -101,7 +109,24 @@ internal fun findExportsAndDeclaredTypes(
             val pos = env.declarationSite(exportedName) ?: root.pos.leftEdge
             val (typeInferences, declarationMetadata) =
                 declInfoForExports[exportedName] ?: DeclInfo.empty
-            Export(exporter, exportedName, value, typeInferences, declarationMetadata, pos)
+            val valueFromStaging: Value<*>?
+            val valueFromRun: Value<*>?
+            if (stage == Stage.Run) {
+                valueFromRun = value
+                valueFromStaging = exportsBefore[exportedName]?.valueFromStaging
+            } else {
+                valueFromRun = null
+                valueFromStaging = value
+            }
+            Export(
+                exporter = exporter,
+                name = exportedName,
+                valueFromStaging = valueFromStaging,
+                valueFromRun = valueFromRun,
+                typeInferences = typeInferences,
+                declarationMetadata = declarationMetadata,
+                position = pos,
+            )
         },
         declaredTypes.toList(),
         topLevelMetadataDecl?.let { stayFor(it) },
@@ -121,7 +146,7 @@ fun isStableForMetadata(value: Value<*>): Boolean {
     if (value.stability == ValueStability.Stable) { return true }
 
     // Before we can leave values like Pairs as stable, we need to adjust the
-    // interpreter&|typer to allow us to preserve type parameter info, and
+    // interpreter&|typer to allow us to preserve type parameter info and
     // to allow un-inlining values by reconverting values to constructor
     // expressions during TmpL translation.
     // But for metadata we don't need that, and Pairs are used importantly
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/ModularNameEnvironment.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/ModularNameEnvironment.kt
@@ -18,7 +18,7 @@ import lang.temper.value.NotYet
 /**
  * Resolves names of exported or otherwise externally accessible top-level module constructs.
  *
- * Exports are assumed to be available, as are static properties of declared types, and
+ * Exports are assumed to be available, as are static properties of declared types and
  * methods defined on interface types.
  */
 internal class ModularNameEnvironment(
@@ -51,7 +51,7 @@ internal class ModularNameEnvironment(
         if (name is ExportedName) {
             val export = getExport(name)
             if (export != null) {
-                export.value ?: NotYet
+                export.value(cb.stage) ?: NotYet
             } else {
                 // If the module has not reached the export stage, return NotYet.
                 val stage = getModule(name.origin)?.stageCompleted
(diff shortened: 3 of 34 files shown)


common/src/commonMain/kotlin/lang/temper/common/Collections.kt
```diff
@@ -328,9 +328,26 @@ fun <K, V> partiallyOrder(
...
+ *
+ * If there is a v in afterMap\[k] such that k is a dep key for an output element, then v will end
+ * up in the partial order regardless of whether v in items.
+ */
+fun <K, V> partiallyOrderTo(
+    items: Iterable<V>,
+    afterMap: Map<K, Collection<V>>,
+    partialOrder: MutableCollection<V>,
+    depKeyOf: (V) -> K,
+) {
```
mikesamuel, line 349: Refactored to enable easy ordering of a set of runnable modules based on ImportRecord analysis.

frontend/src/commonMain/kotlin/lang/temper/frontend/importstage/ImportStage.kt
```diff
@@ -344,8 +344,9 @@ private fun Value<*>.loc(): CodeLocation? {
         // This actually gets the loc of the type, which is good enough for current needs.
         is TClass -> tag.typeShape.pos
         // And we don't actually even need TFunction right now, but I had an example to work from so supported that.
-        is TFunction -> (stateVector as? UserFunction)?.pos
-        is TType -> ((stateVector as? ReifiedType)?.type as? NominalType)?.definition?.pos
+        is TFunction -> (TFunction.unpack(this) as? UserFunction)?.pos
+        is TType -> (TType.unpack(this).type as? NominalType)?.definition?.pos
+        is TProblem -> TProblem.unpack(this).pos
```
mikesamuel, line 349: Opportunistic add which seemed in line with the intent.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
```diff
@@ -361,7 +360,7 @@ class ModuleAdvancer(
     fun getAllLibraryConfigurations(): List<LibraryConfiguration> = libraryConfigurations.toList()
     fun getLibraryConfiguration(libraryRoot: FilePath) = libraryConfigurations[libraryRoot]
 
-    fun advanceModules(stopBefore: Stage? = null) = advanceModules { stopBefore }
+    fun advanceModules(stopBefore: Stage?) = advanceModules { stopBefore }
```
mikesamuel, line 363: Just made this explicit in the uses of it (mostly above).

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
```diff
@@ -401,20 +400,6 @@ class ModuleAdvancer(
...
-            lazy {
-                buildStdModules(
-                    this,
-                    console,
-                    mayRun = true,
-                )
-            }
-        } else {
-            sharedStdModulesMayNotRun
-        }
```
mikesamuel, line 416: No need to create a separate std module set for running anymore.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
```diff
@@ -644,18 +629,52 @@ private class GroupOfModulesToAdvanceTogether(
...
-        val readyToAdvance = ArrayDeque(modules)
+        // We keep a queue of modules to advance and implement additional
+        // checks while looping over it to delay when waiting for imports.
+        val readyToAdvance = ArrayDeque<Module>()
+        // For Stage.Run emulation, we need to queue modules in a specific
+        // order because the run stage can perform side effects.
+        // So we collect modules that are ready to run here, off the queue,
+        // and once everything is ready to enter the run stage, we empty
+        // this set back onto the queue in the right order.
+        var readyToRun: MutableSet<Module>? = mutableSetOf()
```
mikesamuel, line 640: These comments explain the change. Basically, we hold off running while any pre-run modules are still catching up.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
```diff
@@ -944,7 +962,7 @@ private fun buildStdModules(
         classifyTemperSource = ::defaultClassifyTemperSource,
     )
     advancer.configureLibrary(tentativeStdLibraryConfiguration)
-    val stdModuleConfig = ModuleConfig.default.copy(mayRun = mayRun)
+    val stdModuleConfig = ModuleConfig.default.copy(mayRun = true)
```
mikesamuel, line 965: Could try getting rid of mayRun from ModuleConfig too.

frontend/src/commonMain/kotlin/lang/temper/frontend/FoundAfterModuleStaging.kt
```diff
@@ -47,6 +47,14 @@ internal fun findExportsAndDeclaredTypes(
     val declaredTypes = mutableSetOf<TypeShape>()
     var topLevelMetadataDecl: DeclTree? = null
 
+    val exportsBefore by lazy {
+        buildMap {
+            for (export in (exporter.exports ?: listOf())) {
+                this[export.name] = export
+            }
+        }
+    }
```
mikesamuel, line 56: If we're doing the runStage, we want to create an export object that inherits the staging value from the previous one, but only in that case.

frontend/src/commonMain/kotlin/lang/temper/frontend/ModularNameEnvironment.kt
```diff
@@ -51,7 +51,7 @@ internal class ModularNameEnvironment(
         if (name is ExportedName) {
             val export = getExport(name)
             if (export != null) {
-                export.value ?: NotYet
+                export.value(cb.stage) ?: NotYet
```
mikesamuel, line 54: Here's where the reading happens.  If we're reading for staging, we never look at the imported value from running.  And vice versa.

frontend/src/commonTest/kotlin/lang/temper/frontend/staging/ModuleAdvancerTest.kt
```diff
@@ -349,6 +349,83 @@ class ModuleAdvancerTest {
         """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
     )
 
+    @Test
+    fun runningHappensInOrder() = assertVerboseLogOutput(
```
mikesamuel, line 353: Here's a new test that the ModuleAdvancer changes above run modules in a sensible order.

fundamentals/src/commonMain/kotlin/lang/temper/env/Exporter.kt
mikesamuel, line 1: And here's where I change Export to allow storing the value from static computation, and the dynamically computed version separately.

fundamentals/src/commonMain/kotlin/lang/temper/env/Exporter.kt
```diff
@@ -50,16 +79,21 @@ data class Export(
...
+        }
+        valueFromRun?.let {
+            s.whenUnvisited(it) {
+                it.addStays(s)
             }
         }
         declarationMetadata.forEach { (_, values) ->
-            values.forEach {
-                if (value != null) {
-                    s.whenUnvisited(value) {
```
mikesamuel, line 61: `value` should have been `it` here.

interp/src/commonMain/kotlin/lang/temper/interp/Interpreter.kt
```diff
@@ -2312,6 +2312,7 @@ class Interpreter(
             try {
                 val replacementMaker = callSiteReplacementMaker
                 if (replacementMaker != null) {
+                    require(stage != Stage.Run) // Run stage should not modify the AST
```
mikesamuel, line 2315: This checks that no funny business is happening with AST edits during Stage.Run.

fundamentals/src/commonMain/kotlin/lang/temper/env/Exporter.kt
```diff
@@ -26,7 +27,27 @@ interface Exporter {
...
+     * For example, `export let profilingStartStamp = MonotonicClock.now()`
+     * depends on runtime information.  We cannot compute that statically
+     * because there are no stable compile-time semantics for the module
+     * load time.
+     *
+     * In some cases, macro function values, we might have this value but
+     * not one for [valueFromRun].  Macros are not necessary after
+     * the [Stage.GenerateCode] stage completes.
+     */
+    val valueFromStaging: Value<*>?,
```
tjpalmer, line 46: Will this enable us already to use class instances in config.temper.md? I guess recent imu work might also relate to that matter.

tjpalmer (approved):

fundamentals/src/commonMain/kotlin/lang/temper/env/Exporter.kt
mikesamuel: Actually yes.  In ModuleAdvancer it makes modules line up before they run, except for config modules so they can go through Run stage before other modules do.  It just means config modules should never try to import from same library.

tjpalmer: > Actually yes.

Maybe time to start improving config, then. And maybe good timing, since richer structure would be nice for connected dependencies.

> In ModuleAdvancer it makes modules line up before they run, except for config modules so they can go through Run stage before other modules do. It just means config modules should never try to import from same library.

Sort of sad, even though it makes sense. Even in node, I've usually imported package.json into my code and not vice versa.

mikesamuel: Yeah. I expect what we might need to do is allow backends to define their own, never translated library, and then do some hacks to allow config files to import backend stuff conditionally.  Like the below:

```ts
if (usingBackend("js")) {
  let { /* JS backend stuff */ } = import("backend-config/js/...");
  // config stuff for JS backend
}
```

> Sort of sad, even though it makes sense. Even in node, I've usually imported package.json into my code and not vice versa.

I wonder if we can enable that with a bit of indirection.

Like a library's configuration implicitly produces some bundle of information that can itself be imported but without actually importing the config module.

Like a lot of the complaints about non-declarative config fall away if the non-declarative config produces declarative artifacts.  A lot of times it doesn't matter that they're not available as declarative in the source tree if they're to tools when they need it.

PR #463, opened 2026-07-12
frontend: use operator form for remaining arithmetic operators

mikesamuel:
This eliminates most remaining uses of cover functions in BuiltinFuns. The remaining ones are the comparison operators which I plan to do in a subsequent PR.

This opportunistically fixes a bug where control flow simplification inserted uses of the BuiltinName `!` after generate code stage inlined it which prevented pure user functions from functioning properly since they close over *EmptyEnvironment* due to not requiring any global names. This broke some of the test running tests.

This opportunistically also gets be-cpp working on MacOS by redoing the g++-means-clang fix on be-cpp.


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/RunCpp.kt
```diff
@@ -60,7 +60,8 @@ fun runCpp(
 }
 
 object GppCommand : VersionedTool {
-    override val cliNames = listOf("g++")
+    override val cliNames = listOf("g++-16", "g++-15", "g++-14", "g++")
+    // On MacOS, g++ is an alias for clang, but the numbered ones are legit
```
mikesamuel, line 64: Copied from be-cppv/.../RunCpp.kt

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -926,408 +928,132 @@ object BuiltinFuns {
         null,
     ) { a -> a }
 
-    /**
```
mikesamuel, line 929: All these doc comments moved to BuiltinOperatorSpecs.kt.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -926,408 +928,132 @@ object BuiltinFuns {
...
-     * 1 + 1.0
-     * ```
-     *
-     * `+` does not work on [snippet/type/String]s.  Use [snippet/syntax/string/interpolation] instead.
-     *
-     * ```temper FAIL
-     * "foo" + "bar"
-     * ```
-     */
     val plusFn = CoverFunction(
```
mikesamuel, line 931: plusFn and minusFn are still used in BuiltinEnvironment in the definition of `++` and `--`.

I plan to fold that into DesugarOperation in a future PR.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinLogicalOperators.kt
mikesamuel, line 1: The place in fundamentals where interface LogicalOperators is defined explains why this exists.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinOperatorSpecs.kt
```diff
@@ -8,6 +8,37 @@ import lang.temper.value.CallableValue
...
+     * 1 + 1.0
+     * ```
+     *
+     * `+` does not work on [snippet/type/String]s.  Use [snippet/syntax/string/interpolation] instead.
+     *
+     * ```temper FAIL
+     * "foo" + "bar"
+     * ```
+     */
     "+_" to listOf(
```
tjpalmer, line 42: Could be nice sometime just to define these in core/implicits, then just connect them.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinOperatorSpecs.kt
```diff
@@ -18,6 +49,342 @@ val builtinOperatorSpecs: Map<String, List<CallableValue>> = mapOf(
...
+     * var x = 1;
+     * +x == -(-x) &&  // double negation is identity
+     * --x == 0        // but two adjacent `-` means pre-decrement
+     * ```
+     */
+    "-_" to listOf(
+        BuiltinFuns.minusIntFn,
+        BuiltinFuns.minusLongFn,
+        BuiltinFuns.minusFloatFn,
+    ),
```
tjpalmer, line 90: Meanwhile, these still seem a bit odd vs what was in my mind of being defined as different methods on different classes vs being in common lists here.

tjpalmer (approved):

tjpalmer:
And glad the overall line count didn't go up, even if it also didn't really go down.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinOperatorSpecs.kt
mikesamuel: We might need to be careful in implicits not to use `+` in `Implicits.temper`.  We'd need to refer to the definition.  And testing the interpreter by doing arithmetic would be hard because it'd lack it's own arithmetic operators.

mikesamuel:
> And glad the overall line count didn't go up, even if it also didn't really go down.

Yeah. Added test code to SyntaxMacroStageTest, so prod code went down a bit.

PR #464, opened 2026-07-18
frontend: `++` and `--` are regular operators

mikesamuel:
More work to eliminate cover functions.  This time, `++` and `--`.

Previously, `++` and `--` corresponded to `+= 1` which is problematic for non-Int32 operands because `1` is an `Int32` literal.

This commit changess the desugaring of `++` to be a reassigning use of the `.succ()` method which is now implemented on `Int32`, `Int64`, and `Float64` to add the right kind of `1`.  The `--` operator desugars to a call to `.pred()`, both of those names reminiscent of the *S* and *P* operators from Peano arithmetic.  It's intended that this will allow persistent cursors that can move forward and back with `++` and `--`.

Previously, `DesugarPrefixOperatorMacro` desugared `++i` calls, and `DesugarPostfixOperatorMacro` was used to turn `i++` into ones that returned the value of `i` before the reassignment.  Now, that logic is folded into `DesugarOperation` which handles regular operators and both kinds of reassigning operators.

Since the old implementation just desugared to `i += 1` there were no `.succ()` or `.pred()` method calls that needed to be connected by backends.  To avoid increasing the backend implementors burden, `SimplifyDotHelper` now inlines calls to those methods for the well known numeric types.

This commit fixes a bug in `CleanupTemporaries` shaken out by uses of `++` in some control flow functional tests in nested loops.


frontend/src/commonMain/kotlin/lang/temper/frontend/implicits/FloatFns.kt
```diff
@@ -175,6 +175,12 @@ internal interface FloatFns {
         }
     }
 
+    object Pred : SigFnBuilder("core.type Float64.pred()") {
```
mikesamuel, line 178: I don't know if these are necessary.  There are implementations in Implicits.temper.

be/src/commonMain/kotlin/lang/temper/be/README.md
```diff
@@ -63,11 +63,13 @@ The set of *\@connected* methods from Temper's *ImplicitsModule* are:
 - `core.type Float64.min()`
 - `core.type Float64.near()`
 - `core.type Float64.pi`
+- `core.type Float64.pred()`
```
mikesamuel, line 66: Numeric types get `pred()` and `succ()` which is how we get access to them without a concept of unit step: `1`, `1.0` or `1I64` or something else as appropriate to the operand.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/int.js
```diff
@@ -1,11 +1,39 @@
...
+ * @returns {bigint}
+ */
+export const int64Succ = (x) => x + 1n;
+
+/**
+ * Predecessor function.
+ * @param {bigint} n
+ * @returns {bigint}
+ */
+export const int64Pred = (x) => x - 1n;
```
tjpalmer, line 27: Do we do the wrapping elsewhere on these?

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/float.js
```diff
@@ -3,6 +3,20 @@
...
+ * @returns {number}
+ */
+export const float64Succ = (x) => x + 1.0;
+
+/**
+ * Predecessor function.
+ * @param {number} n
+ * @returns {number}
+ */
+export const float64Pred = (x) => x - 1.0;
```
tjpalmer, line 18: Could be nice to inline these but maybe not urgent in this PR.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinOperatorSpecs.kt
```diff
@@ -20,14 +21,16 @@ val builtinOperatorSpecs: Map<String, List<CallableValue>> = mapOf(
      * - *Prefix* with one [snippet/type/Float64]: numeric identity
      *
      * ```temper
-     * 1   + 2   == 3   &&
-     * 1.0 + 2.0 == 3.0 &&
-     * +1        == 1   &&
-     * +1.0      == 1.0
+     * 1    + 2    == 3    &&
+     * 1.0  + 2.0  == 3.0  &&
+     * 1I64 + 2I64 == 3I64 &&
```
tjpalmer: I think we encourage `1i64` rather than `1I64`.

frontend/src/commonMain/kotlin/lang/temper/frontend/implicits/FloatFns.kt
tjpalmer: > I don't know if these are necessary. There are implementations in Implicits.temper.

I'll guess they aren't, but sometimes I have to try things out.

tjpalmer (commented):
Overall, I like the strategy. Do we need any extra tests on wrapping arithmetic, or are we good already? Beyond that, I'd just recommend the `1i64` syntax.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/float.js
mikesamuel: The SimplifyDotHelper pass does that.  Not sure this is necessary, but be-js:check was flagging it. I can remove.

tjpalmer: > The SimplifyDotHelper pass does that. Not sure this is necessary, but be-js:check was flagging it. I can remove.

We have an annoying be-js test that requires either implementing the functions or listing them as not needed. I'd like to remove that test.

tjpalmer: Maybe that's connecteds-supported-test.mjs.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinOperatorSpecs.kt
mikesamuel: Recased in Implicits.temper implementations of succ and pred fot Int64 and in docs for `+`

tjpalmer (approved):

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/float.js
mikesamuel: Yeah.  I think that test has outlived its usefulness.

I think you've expressed this before, and maybe I have a bit of a hoarder mindset regarding tests that have caught problems in the past.

That test was super useful when I was initially working on the (now obsolete) connection mechanism.  Now, our functional-test-suite has largely taken over the job of checking mandatory connections.

PR #465, opened 2026-07-21
Pass connected funtest on all but js and interp

tjpalmer:
- Pass connected functions funtest on all backends except interp and js
- Also refine be-py connecteds some
- To some extent, this is drafting the plan for userspace connecteds, maybe good approximately as is, but some forms of adjustment wouldn't be too hard
- Js is hard because standard import is only top level and our "whatever.internal.js" handling is still awkward
- Interp needs some way for backends to know when to use temper impl vs connected code

### DIFF
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/Cpp.out-grammar
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/Cpp.out-grammar
@@ -12,8 +12,11 @@ Program ::= stmts%Global*"\n" & "\n";
 // top-levels
 Global = StructDecl | FuncDecl | VarDecl | StructDef | DerivedStructDef | TemplateStructDef | TemplateFuncDef | TemplateFuncDecl | TypeDef | FuncDef | VarDef | PreProc | Comment | Namespace;
 
-PreProc = Define | Undef | Pragma | IncludeGuard | Include | IfPreProc;
+PreProc = Define | Undef | Pragma | IncludeGuard | Include | IncludeLocal | IfPreProc;
+// For more on includes, see https://en.cppreference.com/cpp/preprocessor/include#Notes
 Include ::= "#" & "include" & "<" & path%Raw & ">" & "\n";
+/** For string literals, although escapes don't work. */
+IncludeLocal ::= "#" & "include" & path%LiteralExpr & "\n";
 Define ::= "#" & "define" & name%SingleName & value%Expr? & "\n";
 Undef ::= "#" & "undef" & name%SingleName & "\n";
 Pragma ::= "#" & "pragma" & text%Raw & "\n";
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
@@ -224,10 +224,12 @@ class CppBackend private constructor(
                 fileExtensionMap = mapOf(
                     FileType.Module to lang.ext,
                     FileType.Script to lang.ext,
+                    FileType.Header to HPP_EXT,
                 ),
                 mimeTypeMap = mapOf(
                     FileType.Module to MimeType.cppSource,
                     FileType.Script to MimeType.cppSource,
+                    FileType.Header to MimeType.cppSource,
                 ),
             )
 
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
@@ -11,10 +11,12 @@ import lang.temper.be.tmpl.asReceiverMember
 import lang.temper.be.tmpl.injectSuperCallMethods
 import lang.temper.be.tmpl.mutatingMemberNames
 import lang.temper.common.MimeType
+import lang.temper.common.subListToEnd
 import lang.temper.fs.ResourceDescriptor
 import lang.temper.fs.declareResources
 import lang.temper.log.FilePath
 import lang.temper.log.FilePathSegment
+import lang.temper.log.asFilePath
 import lang.temper.log.dirPath
 import lang.temper.log.filePath
 import lang.temper.log.plus
@@ -131,6 +133,18 @@ class CppBackend private constructor(
             result
         }
 
+        // Connected code.
+        val connectedFiles = buildList {
+            for (file in rawBackendFiles) {
+                MetadataFileSpecification(
+                    // Skip the library name part.
+                    path = file.key.segments.subListToEnd(1).asFilePath(),
+                    mimeType = MimeType.cppSource,
+                    content = file.value,
+                ).also { add(it) }
+            }
+        }
+
         val initPath = filePath(INIT_NAME)
 
         dependenciesBuilder.addMetadata(
@@ -149,7 +163,7 @@ class CppBackend private constructor(
             testNs = testNs,
         )
 
-        return translations + listOf(
+        return translations + connectedFiles + listOf(
             MetadataFileSpecification(
                 path = filePath(MAIN_CPP_FILE),
                 mimeType = MimeType.cppSource,
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
@@ -3,7 +3,9 @@ package lang.temper.be.cpp
 import lang.temper.be.Backend
 import lang.temper.be.tmpl.TmpL
 import lang.temper.be.tmpl.TypedArg
+import lang.temper.be.tmpl.isStdLib
 import lang.temper.be.tmpl.mapParameters
+import lang.temper.be.tmpl.parameterDefaultStatementsInfo
 import lang.temper.be.tmpl.referencedNames
 import lang.temper.common.MimeType
 import lang.temper.common.ignore
@@ -43,6 +45,7 @@ import lang.temper.value.TInt64
 import lang.temper.value.TNull
 import lang.temper.value.TProblem
 import lang.temper.value.TString
+import lang.temper.value.connectedSymbol
 
 /** How the C++ backend renders Temper function types. Emitted and matched in one place. */
 private const val STD_FUNCTION_PREFIX = "std::function"
@@ -664,15 +667,16 @@ class CppTranslator(
         val bindings = type.bindings
         val typeArgs = bindings.map { translateType2(it) }
         if (typeArgs.isNotEmpty()) {
-            when {
-                def == WellKnownTypes.functionTypeDefinition ->
+            when (def) {
+                WellKnownTypes.functionTypeDefinition ->
                     return stdFunction(typeArgs.last(), typeArgs.dropLast(1))
-                def in listLikeDefs ->
+                in listLikeDefs ->
                     return sharedPtr(stdVector(typeArgs.first()))
-                def in mapLikeDefs ->
+                in mapLikeDefs ->
                     return sharedPtr(cpp.template(cppBaseTypeForDefinition(def)!!, typeArgs))
-                def == WellKnownTypes.dequeTypeDefinition ->
+                WellKnownTypes.dequeTypeDefinition ->
                     return sharedPtr(cpp.template(cppBaseTypeForDefinition(def)!!, typeArgs))
+                else -> {}
             }
         }
         cppBaseTypeForDefinition(def)?.let { base ->
@@ -746,9 +750,7 @@ class CppTranslator(
             cppBaseTypeForDefinition(def) ?: run {
                 // Check if this is a type formal with a known template parameter name
                 val typeFormalName = typeFormalNames[def] ?: typeFormalNamesByText[typeFormalKey(def)]
-                if (typeFormalName != null) {
-                    typeFormalName
-                } else {
+                typeFormalName ?: run {
                     when (val loc = def.sourceLocation) {
                         ImplicitsCodeLocation -> when (val defName = def.name) {
                             is ExportedName -> translateImplicitsType(defName.baseName.builtinKey)
@@ -2797,7 +2799,7 @@ class CppTranslator(
         // member ordering in TmpL.
         prepopulatePropertyDotNames(topLevel)
         val declaredSetters = mutableSetOf<String>()
-        val structFields = buildList<Cpp.StructPart> {
+        val structFields = buildList {
             // Add virtual destructor for interfaces to enable dynamic_pointer_cast
             if (isInterface) {
                 val dtorName = "~${cpp.name(topLevel.name).id.text}"
@@ -2903,12 +2905,17 @@ class CppTranslator(
         } else {
             formals.map { cpp.name(it.name) }
         }
+        val block = when {
+            topLevel.metadata.any { it.key.symbol == connectedSymbol } && !mod!!.isStdLib ->
+                translateConnectedBody(topLevel, allParamNames)
+            else -> translateBlock(topLevel.body)
+        }
         val func = cpp.func(
             cpp.name(topLevel.name),
             translateType(topLevel.returnType),
             allParamTypes,
             allParamNames,
-            translateBlock(topLevel.body),
+            block,
         )
         if (typeFormals.isNotEmpty()) {
             val templateParams = typeFormals.map { formal ->
@@ -2967,6 +2974,36 @@ class CppTranslator(
         for (key in savedTypeFormalKeys) {
             typeFormalNamesByText.remove(key)
         }
+    }
+
+    private fun translateConnectedBody(
+        fn: TmpL.FunctionDeclaration,
+        paramNames: List<Cpp.SingleName>,
+    ): Cpp.BlockStmt = cpp.pos(fn) {
+        hasConnected = true
+        val defaulting = fn.parameterDefaultStatementsInfo()
+        buildList {
+            for (statement in defaulting.defaultStatements) {
+                addAll(translateStatement(statement))
+            }
+            cpp.callExpr(
+                expr = cpp.name(cpp.name("_connected"), cpp.name(fn.name)),
+                args = buildList {
+                    for ((tmpl, cppName) in fn.parameters.parameters.zip(paramNames)) {
+                        when {
+                            tmpl.optional ->
+                                cpp.name(defaulting.parameterMapping.getValue(tmpl.name.name))
+                            else -> cppName.deepCopy()
+                        }.also { add(it) }
+                    }
+                },
+            ).let { call ->
+                when ((fn.returnType.ot as? TmpL.NominalType)?.typeName?.sourceDefinition) {
+                    WellKnownTypes.voidTypeDefinition -> cpp.exprStmt(call)
+                    else -> cpp.returnStmt(call)
+                }
+            }.also { add(it) }
+        }.let { cpp.blockStmt(it) }
     }
 
     /**
@@ -3182,7 +3219,13 @@ class CppTranslator(
         impl.add(initFuncDef)
     }
 
+    private var mod: TmpL.Module? = null
+
+    /** If any userspace connected functions have been defined. */
+    private var hasConnected: Boolean = false
+
     fun translateModule(mod: TmpL.Module): List<Backend.TranslatedFileSpecification> {
+        this.mod = mod
         includes.clear()
         currentModuleLocation = mod.codeLocation.codeLocation
         preprocessImports(mod)
@@ -3317,6 +3360,9 @@ class CppTranslator(
                     content = cpp.program(
                         buildList {
                             add(cpp.include("$modPath$hppName"))
+                            if (hasConnected) {
+                                add(cpp.includeLocal("_connected.hpp"))
+                            }
                             addAll(namespaced(implVarDecls + impl))
                         },
                     ),
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
@@ -2993,7 +2993,7 @@ class CppTranslator(
                         when {
                             tmpl.optional ->
                                 cpp.name(defaulting.parameterMapping.getValue(tmpl.name.name))
-                            else -> cppName.deepCopy()
+                            else -> cppName
                         }.also { add(it) }
                     }
                 },
--- a/be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpBackend.kt
+++ b/be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpBackend.kt
@@ -29,6 +29,7 @@ import lang.temper.log.FilePath
 import lang.temper.log.dirPath
 import lang.temper.log.filePath
 import lang.temper.log.last
+import lang.temper.log.resolveFile
 import lang.temper.name.BackendId
 import lang.temper.name.BackendMeta
 import lang.temper.name.FileType
@@ -183,6 +184,16 @@ class CSharpBackend(setup: BackendSetup<CSharpBackend>) : Backend<CSharpBackend>
                         else -> {}
                     }
                 }
+            }
+            // Connected source.
+            for (file in rawBackendFiles) {
+                // Skip the library name dir and the file name, and keep the middle
+                val subspace = chooseSubspace(file.key.segments).subList(1, file.key.segments.size - 1)
+                MetadataFileSpecification(
+                    path = dirPath(SRC_PROJECT_DIR).resolve(dirPath(subspace)).resolveFile(file.key.segments.last()),
+                    mimeType = mimeType,
+                    content = file.value,
+                ).also { add(it) }
             }
             // Proj.
             val rootNamespaceByName = names.rootNamespaces.associate { it.first.libraryName.text to it.second }
--- a/be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpTranslator.kt
+++ b/be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpTranslator.kt
@@ -10,10 +10,12 @@ import lang.temper.be.tmpl.TypedArg
 import lang.temper.be.tmpl.canBeNull
 import lang.temper.be.tmpl.dependencyCategory
 import lang.temper.be.tmpl.isCommonlyImplied
+import lang.temper.be.tmpl.isStdLib
 import lang.temper.be.tmpl.isYieldingStatement
 import lang.temper.be.tmpl.libraryName
 import lang.temper.be.tmpl.mapParameters
 import lang.temper.be.tmpl.orInvalid
+import lang.temper.be.tmpl.parameterDefaultStatementsInfo
 import lang.temper.be.tmpl.toSigBestEffort
 import lang.temper.be.tmpl.typeOrInvalid
 import lang.temper.be.tmpl.withoutNullOrBubble
@@ -24,6 +26,7 @@ import lang.temper.lexer.Genre
 import lang.temper.lexer.withTemperAwareExtension
 import lang.temper.library.LibraryConfigurations
 import lang.temper.log.FilePath
+import lang.temper.log.FilePathSegment
 import lang.temper.log.Position
 import lang.temper.log.dirPath
 import lang.temper.log.resolveFile
@@ -57,7 +60,6 @@ import lang.temper.type2.isVoidLike
 import lang.temper.type2.withNullity
 import lang.temper.type2.withType
 import lang.temper.value.DependencyCategory
-import lang.temper.value.MetadataValueMapHelpers.get
 import lang.temper.value.TBoolean
 import lang.temper.value.TClass
 import lang.temper.value.TClosureRecord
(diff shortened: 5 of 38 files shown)


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/Cpp.out-grammar
```diff
@@ -12,8 +12,11 @@ Program ::= stmts%Global*"\n" & "\n";
 // top-levels
 Global = StructDecl | FuncDecl | VarDecl | StructDef | DerivedStructDef | TemplateStructDef | TemplateFuncDef | TemplateFuncDecl | TypeDef | FuncDef | VarDef | PreProc | Comment | Namespace;
 
-PreProc = Define | Undef | Pragma | IncludeGuard | Include | IfPreProc;
+PreProc = Define | Undef | Pragma | IncludeGuard | Include | IncludeLocal | IfPreProc;
+// For more on includes, see https://en.cppreference.com/cpp/preprocessor/include#Notes
 Include ::= "#" & "include" & "<" & path%Raw & ">" & "\n";
+/** For string literals, although escapes don't work. */
+IncludeLocal ::= "#" & "include" & path%LiteralExpr & "\n";
```
tjpalmer, line 19: Technically not necessarily local-priority includes, but typically all compilers treat them that way, and it at least provides a different convention that can be used by compilers.

Separately, I tried explicit double-quotes around raw to make differences from strings more explicit, but the spacing was off, and it doesn't really accomplish anything.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
```diff
@@ -131,6 +133,18 @@ class CppBackend private constructor(
...
+        val connectedFiles = buildList {
+            for (file in rawBackendFiles) {
+                MetadataFileSpecification(
+                    // Skip the library name part.
+                    path = file.key.segments.subListToEnd(1).asFilePath(),
+                    mimeType = MimeType.cppSource,
+                    content = file.value,
+                ).also { add(it) }
+            }
+        }
```
tjpalmer, line 146: For all the backends, I haven't tested submodules yet. Putting that off for build tests in a future PR.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
```diff
@@ -224,10 +238,12 @@ class CppBackend private constructor(
                 fileExtensionMap = mapOf(
                     FileType.Module to lang.ext,
                     FileType.Script to lang.ext,
+                    FileType.Header to HPP_EXT,
```
tjpalmer, line 241: We need extensions registered for all files we might want to copy.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
```diff
@@ -224,10 +238,12 @@ class CppBackend private constructor(
...
                 ),
                 mimeTypeMap = mapOf(
                     FileType.Module to MimeType.cppSource,
                     FileType.Script to MimeType.cppSource,
+                    FileType.Header to MimeType.cppSource,
```
tjpalmer, line 246: We were already supplying this mime type for header files, just not using this map.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
```diff
@@ -664,15 +667,16 @@ class CppTranslator(
...
-                def in listLikeDefs ->
+                in listLikeDefs ->
                     return sharedPtr(stdVector(typeArgs.first()))
-                def in mapLikeDefs ->
+                in mapLikeDefs ->
                     return sharedPtr(cpp.template(cppBaseTypeForDefinition(def)!!, typeArgs))
-                def == WellKnownTypes.dequeTypeDefinition ->
+                WellKnownTypes.dequeTypeDefinition ->
                     return sharedPtr(cpp.template(cppBaseTypeForDefinition(def)!!, typeArgs))
+                else -> {}
```
tjpalmer, line 679: The overall change recommended by ij, and it's cleaner overall except for needing the `else`. Went along with it.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
```diff
@@ -746,9 +750,7 @@ class CppTranslator(
             cppBaseTypeForDefinition(def) ?: run {
                 // Check if this is a type formal with a known template parameter name
                 val typeFormalName = typeFormalNames[def] ?: typeFormalNamesByText[typeFormalKey(def)]
-                if (typeFormalName != null) {
-                    typeFormalName
-                } else {
+                typeFormalName ?: run {
```
tjpalmer, line 753: Also recommended by ij.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
```diff
@@ -2903,12 +2905,17 @@ class CppTranslator(
         } else {
             formals.map { cpp.name(it.name) }
         }
+        val block = when {
+            topLevel.metadata.any { it.key.symbol == connectedSymbol } && !mod!!.isStdLib ->
```
tjpalmer, line 2909: The `!mod!!` is unfortunate, but referencing the modules is the easiest way to check this, and I didn't feel like adjusting when the module was passed in, so it's technically nullable. Usage guarantees non-null here, though.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
```diff
@@ -2969,6 +2976,36 @@ class CppTranslator(
         }
     }
 
+    private fun translateConnectedBody(
+        fn: TmpL.FunctionDeclaration,
+        paramNames: List<Cpp.SingleName>,
+    ): Cpp.BlockStmt = cpp.pos(fn) {
```
tjpalmer, line 2982: These connected body translations are very similar across backends but with subtle differences because of different ways of doing things in different backends.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
```diff
@@ -2969,6 +2976,36 @@ class CppTranslator(
...
+                addAll(translateStatement(statement))
+            }
+            cpp.callExpr(
+                expr = cpp.name(cpp.name("_connected"), cpp.name(fn.name)),
+                args = buildList {
+                    for ((tmpl, cppName) in fn.parameters.parameters.zip(paramNames)) {
+                        when {
+                            tmpl.optional ->
+                                cpp.name(defaulting.parameterMapping.getValue(tmpl.name.name))
+                            else -> cppName
```
tjpalmer, line 2996: Managing parameter names for forwarding as args definitely varies by backend. And I mostly haven't thought hard about rest args yet, for example. I've commented variously around this pr on this or related matters. We can either be more robust about it in the future or else see how it relates to rest arg plans in the future and/or validate in the front end, but I don't want to deal with it in this pr.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpBackend.kt
```diff
@@ -184,6 +185,16 @@ class CSharpBackend(setup: BackendSetup<CSharpBackend>) : Backend<CSharpBackend>
...
+            // Connected source.
+            for (file in rawBackendFiles) {
+                // Skip the library name dir and the file name, and keep the middle
+                val subspace = chooseSubspace(file.key.segments).subList(1, file.key.segments.size - 1)
+                MetadataFileSpecification(
+                    path = dirPath(SRC_PROJECT_DIR).resolve(dirPath(subspace)).resolveFile(file.key.segments.last()),
+                    mimeType = mimeType,
+                    content = file.value,
+                ).also { add(it) }
+            }
```
tjpalmer, line 197: This logic also is fairly similar across backends, but file layout or usage varies somewhat.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpTranslator.kt
```diff
@@ -254,10 +259,7 @@ internal class CSharpTranslator(
         val modulePath = Backend.defaultFilePathForSource(
             libraryConfiguration, moduleName, "",
         )
-        val subspace = modulePath.segments.map { segment ->
-            segment.withTemperAwareExtension("").fullName.dashToPascal()
-        }
-        return subspace
+        return chooseSubspace(modulePath.segments)
```
tjpalmer, line 262: Factored out for use also with connected files.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/lua.out-grammar
```diff
@@ -9,9 +9,16 @@ Program = Chunk;
...
-Stmt = LabelStmt | DoStmt | WhileStmt | IfStmt | SetStmt | FunctionStmt | LocalDeclStmt | LocalStmt | LocalFunctionStmt | CallStmt | Comment | GotoStmt;
+Stmt = LabelStmt | DoStmt | WhileStmt | IfStmt | SetStmt | FunctionStmt | LocalDeclStmt | LocalStmt | LocalFunctionStmt | CallStmt | Comment | GotoStmt | Connected;
 LastStmt = ReturnStmt | BreakStmt | GotoStmt;
 
+/**
+ * For embedding external raw `@connected` Lua source. Should be placed after
+ * predeclarations and before other top-level statements.
+ */
+Connected(source%`String`);
+Connected.renderTo = `tokenSink.value(source)`;
```
tjpalmer, line 20: Without restructuring this more, it would be hard to restrict exactly where Connected gets used, but I named it Connected and put this comment here to encourage proper usage instead of abuse.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaFormattingHints.kt
```diff
@@ -33,6 +34,9 @@ internal class LuaFormattingHints : FormattingHints {
         if (noSpace.matches(preceding.text) || noSpace.matches(following.text)) {
             return false
         }
+        if (preceding.type == OutputTokenType.OtherValue) {
+            return false
+        }
```
tjpalmer, line 39: Similar rule as I put in PyFormattingHints.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaLocalCounter.kt
```diff
@@ -108,6 +108,7 @@ class LuaLocalCounter(val cb: (Int) -> Unit) {
     }
 
     private fun countLocalsBy(stmt: Lua.Stmt): Int = when (stmt) {
+        is Lua.Connected -> 1 // Require all in one local.
```
tjpalmer, line 111: We'll need to document that inlined connected code for be-lua (and also really for be-py, though less vital) must expose only one top-level local called `_connected`. I haven't actually tested edge cases of what happens when we overrun the limit for locals at the top level with connecteds involved, but it might be worth checking that edge case sometime.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaLocalRemover.kt
```diff
@@ -285,6 +285,7 @@ class LuaLocalRemover {
     }
 
     internal fun scan(stmt: Lua.Stmt): Lua.Stmt? = when (stmt) {
+        is Lua.Connected -> stmt.deepCopy() // TODO Anything we can/should do here?
```
tjpalmer, line 288: Since we can't do anything here without modifying the connected lua code, it's hard to accommodate. I suspect worse case we could just change the max handling to give space for one more local.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaSupportNetwork.kt
```diff
@@ -526,18 +526,21 @@ internal object LuaSupportNetwork : SupportNetwork {
...
-                .split(".", "::")
-                // Skip module name.
-                .subListToEnd(1)
-                .joinToString("_") { part ->
-                    // Skip qualifiers and parens.
-                    part.split(" ").last().trimEnd('(', ')')
-                }
-                .lowercase(),
-        )
+        else if connectedKey.startsWith("core.") || connectedKey.startsWith("std/") ->
```
tjpalmer, line 529: This was trying to connect user code to our temper core before this change.

be-py/src/commonMain/kotlin/lang/temper/be/py/py.out-grammar
```diff
@@ -22,12 +22,20 @@ let imports = """
...
 Program(
     dependencyCategory%`DependencyCategory`,
     genre%`Genre`,
     outputPath%`FilePath`,
 );
+Program.connected.default = `null`;
+
+/** For embedding external raw `@connected` Python source. */
+Connected(source%`String`);
+Connected.renderTo = `tokenSink.value(source)`;
```
tjpalmer, line 38: This is me moving the Raw to a more constrained grammar rule, per earlier discussions.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyAst.kt
```diff
@@ -151,9 +151,9 @@ fun garbageStmt(pos: Position, d: TmpL.Diagnostic?): Py.Stmt = garbageStmt(pos,
 fun Py.Program?.orEmpty(outputPath: FilePath): Py.Program =
     this ?: Py.Program(
         unknownPos,
-        listOf(),
-        DependencyCategory.Production,
-        Genre.Library,
+        body = listOf(),
+        dependencyCategory = DependencyCategory.Production,
+        genre = Genre.Library,
         outputPath = outputPath,
```
tjpalmer, line 157: Needed because `connected` is now before these.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -175,10 +175,9 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
...
     }
 
-    /** Supports simple `_whatever` naming until we can arrange it more generally on names. */
-    fun choosePrettyPrivateSourceName(name: ResolvedParsedName, kind: TmpL.IdKind): String {
-        val styledName = styleName(safeIdent(name.baseName.nameText), kind)
-        return "_$styledName"
+    /** Provide a pretty name, whether exported or not. */
+    fun choosePrettyName(name: ResolvedParsedName, kind: TmpL.IdKind): String {
+        return avoidReserved(styleName(safeIdent(name.baseName.nameText), kind))
     }
```
tjpalmer, line 181: And now saying `_connected.whatever(...)` instead of `_whatever(...)`.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
```diff
@@ -122,9 +127,23 @@ class RustBackend(setup: BackendSetup<RustBackend>) : Backend<RustBackend>(Facto
...
                 }
             }
+            // Copy connected code.
+            for (file in rawBackendFiles) {
+                // Put rust connected modules into a subdir, under "src/" but skipping library name.
+                val connectedDir = file.key.dirName().resolveDir("_connected").segments.subListToEnd(1)
+                val fileName = when (val fileName = file.key.last().fullName) {
+                    "_connected.rs" -> "mod.rs"
+                    else -> fileName
+                }
```
tjpalmer, line 137: To let connected code have its own submodules, we need a subdir for the connected module, and I didn't like adding a separate connected subdir for rust in the temper source tree when no other backends need that. So I conjure the subdir here and rename `_connected.rs` to `mod.rs` under that.

Once we stabilize all the userspace connected code, we need to document the rules under each individual backend.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustBackend.kt
```diff
@@ -315,7 +333,7 @@ private fun MutableList<Backend.OutputFileSpecification>.addLib(
             attrs = allowWarnings(pos),
             items = buildList {
                 // Separate mods.
-                declareSubmods(pos, allModKids[libraryConfiguration.libraryRoot] ?: setOf())
+                declareSubmods(pos, modKids)
```
tjpalmer, line 336: Might be customized to include the connected submodule under root now.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -81,8 +81,18 @@ internal fun whereForAnyValueImpl(pos: Position, translatedGenerics: List<Rust.G
...
+            // The main problem with this convention is that leading underscore typically
+            // means unused in Rust rather than private, but it still works.
+            // TODO Just say "connected" and recognize that here explicitly for non-pub?
+            when {
+                name.startsWith("_") -> name.toId(pos) to null
+                else -> name.dashToSnake().toId(pos) to Rust.VisibilityPub(pos)
+            }
+        }
+        add(Rust.Module(pos, id = modId, pub = pub).toItem())
     }
```
tjpalmer, line 96: Super not happy with the above, but it might be ok despite that.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustExt.kt
```diff
@@ -369,11 +379,26 @@ internal fun Rust.Expr.wrapOkOrElse(pos: Position = this.pos) =
...
+                is Rust.RefType -> when (val type = this.type) {
+                    is Rust.Id -> type
+                    // This is a total punt. Shouldn't get here on valid Temper code.
+                    else -> type.toString().toId(pos)
+                }
+                else -> error(this) // Can't actually happen, but Kotlin fails to realize.
+            }
+            else -> error(this)
+        }
     }
```
tjpalmer, line 402: And this is me trying to avoid exceptions in translators for any cases that might possibly happen. I'd rather generate code that won't compile than to crash the translator.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1932,6 +1955,39 @@ class RustTranslator(
...
+            statements = buildList {
+                for (statement in defaulting.defaultStatements) {
+                    addAll(translateStatement(statement))
+                }
+            },
+            result = "_connected".toId(pos).let { connectedModule ->
+                when {
+                    isRoot -> "crate".toKeyId(pos).extendWith(connectedModule)
+                    else -> connectedModule
+                }
```
tjpalmer, line 1973: My awkward root lib handling in be-rust. Maybe ok, but it takes some effort.

functional-test-suite/src/commonMain/kotlin/lang/temper/tests/FunctionalTestStatus.kt
```diff
@@ -25,7 +25,7 @@ val functionalTestStatus: Map<Ft, List<IssueCheck>> = buildMap {
     issue(Ft.RegexZeroAdvance, lua(166))
     issue(Ft.NamesNonascii, lua(228))
     214.let { issue(Ft.TypesNetresponse, cpp(it), interp(it), lua(it)) }
-    456.let { issue(Ft.FunctionsConnected, cpp(it), csharp(it), interp(it), js(it), lua(it), rust(it)) }
+    456.let { issue(Ft.FunctionsConnected, interp(it), js(it)) }
```
tjpalmer, line 28: Narrowing it down.

functional-test-suite/src/commonMain/resources/functions/connected/__connected__.py
```diff
@@ -1,8 +1,8 @@
-def _sum(i: int, j: int, bonus: int) -> int:
-    return i + j + bonus
-
-
-def _prod(hidden: "_Hidden", j: int) -> int:
+class _connected:
     from ._support import Support
```
tjpalmer, line 2: Lower-cased class names in python user code aren't typically idiomatic, but this also technically doesn't need to be a class. And I still haven't made all backends exactly the same in this pr, but some others already say `_connected`, so I did that here, too.

Anyway, after every other backend, including lua, were getting a single subnamespace for connected functions, I felt bad with python not doing the same. Even though I think there could be some value in top-level connected impls for top-level connected declarations.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.cpp
```diff
@@ -0,0 +1,17 @@
+#include "_connected.hpp"
+#include "support.hpp"
+
+namespace work {
+namespace _connected {
```
tjpalmer, line 5: Felt like going with a subnamespace in c++ connected code.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <cstdint>
+#include <work/init.hpp>
```
tjpalmer, line 4: Shaw was using global-ish includes for our main generated headers, so I retained that here. Again, I'm imagining that some of this could be in a different dir in the future. But we'd have to work out actual details later if/when we make that change.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.cpp
```diff
@@ -0,0 +1,17 @@
+#include "_connected.hpp"
```
tjpalmer, line 1: I also generate local-ish includes from generated code to `"_connected.hpp"` because abstractly the exported includes might could be in a different dir than impl code.

@ShawSumma I'd actually like to change our header generation to split exported decls from internal decls, so that they're in different files, then put them in different dirs so that people would need to go to effort to get at the internals.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.lua
```diff
@@ -0,0 +1,21 @@
+local _connected = {};
+
+do
+  local s = require("work._support")
```
tjpalmer, line 4: Lua makes it rather easy to keep requires in a sub-namespace while exported `_connected` members below into the single top-level local. This is an easy way for people to conform to "just one top-level local called `_connected`" in the inlined connected code.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.rs
```diff
@@ -0,0 +1,13 @@
+use super::Hidden;
+use more::Support;
+
+mod more;
```
tjpalmer, line 4: The file as a whole ends up in `_connected/mod.rs`, and the "more" submodule ends up in `_connected/more.rs`.

functional-test-suite/src/commonMain/resources/functions/connected/_support.lua
```diff
@@ -0,0 +1,19 @@
...
+end
+
+---@param i integer
+---@param j integer
+---@return integer
+function Support:prod(i, j)
+    return i * j
+end
+
+return exports
```
tjpalmer, line 19: I made "support" or "more" code for each backend just to prove we have access to them. And made class-like things in each also just for kicks.

functional-test-suite/src/commonMain/resources/functions/connected/WorkConnected.cs
```diff
@@ -0,0 +1,15 @@
+namespace Work
+{
+    static class WorkConnected
```
tjpalmer, line 3: We already had `<Lib/Mod>Global` in be-csharp and be-java, so I reused the same style for `<Lib/Mod>Connected` for both. But after everyone else is just getting `_connected`, it's almost tempting to do that for csharp and java also. Or maybe I should make others more idiomatic instead of making these worse. Don't know. These can be refined in future PRs, if we decide for sure to change from what I've made so far.

functional-test-suite/src/commonMain/resources/functions/connected/WorkConnected.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        internal static int Sum(int i, int j, int bonus)
+        {
+            return i + j + bonus;
+        }
+
+        internal static int Prod(Hidden hidden, int j)
+        {
+            return new MoreSupport().Prod(hidden.I, j);
+        }
```
tjpalmer, line 13: I am trying to make the functions themselves use idiomatic naming. And I've proved I can use the `Hidden` type in each backend.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.lua
```diff
@@ -0,0 +1,21 @@
...
+
+  ---@param i integer
+  ---@param j integer
+  ---@param bonus integer
+  ---@return integer
+  function _connected.sum(i, j, bonus)
+    return i + j + bonus
+  end
+
+  ---@param hidden work.Hidden TODO Actually define types in our Lua.
```
tjpalmer, line 14: We don't yet generated static EmmyLua (or rather Lua Language Server) typing in our LuaTranslator, and we also don't strip numeric suffices yet. But we also don't need static typing for Lua to work. I just wanted to do static type annotations where possible in the lua connected code in this PR, anyway. This is just the case I *can't* do yet.

Shaw also would like to make generating type annotations optional in the future, if we do get around to generating them (which I personally would greatly like, as luals is actually pretty high quality in my using it so far in hobby things).

name/src/commonMain/kotlin/lang/temper/name/BackendId.kt
```diff
@@ -85,6 +85,9 @@ enum class FileType {
 
     /** a script or application to run */
     Script,
+
+    /** a header file for pre-declarations */
+    Header,
```
tjpalmer, line 90: Felt a bit bad adding this, but C and/or OCaml and/or others might also want header files in the future.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/Cpp.out-grammar
```diff
@@ -12,8 +12,11 @@ Program ::= stmts%Global*"\n" & "\n";
 // top-levels
 Global = StructDecl | FuncDecl | VarDecl | StructDef | DerivedStructDef | TemplateStructDef | TemplateFuncDef | TemplateFuncDecl | TypeDef | FuncDef | VarDef | PreProc | Comment | Namespace;
 
-PreProc = Define | Undef | Pragma | IncludeGuard | Include | IfPreProc;
+PreProc = Define | Undef | Pragma | IncludeGuard | Include | IncludeLocal | IfPreProc;
+// For more on includes, see https://en.cppreference.com/cpp/preprocessor/include#Notes
 Include ::= "#" & "include" & "<" & path%Raw & ">" & "\n";
```
mikesamuel, line 17: If you run into formatting glitches you could try token references for some of these with specified operator association.  Like `<` as bracket operators instead of as infix.

functional-test-suite/src/commonMain/resources/functions/connected/__connected__.py
mikesamuel, line 1: Note to self: these resources are just for the connected functional test.

name/src/commonMain/kotlin/lang/temper/name/BackendId.kt
mikesamuel: So this just keys into a lookup table for mime types so helps in picking accurate mime-types?

functional-test-matrix.md
```diff
@@ -26,7 +26,7 @@
 | [ControlFlowLoopReenterable][] | ✅ | ❌<sup>[198][]</sup> | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [ControlFlowLoops][] | ✅ | ❌<sup>[198][]</sup> | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | [FunctionsAsValues][] | ✅ | ❌<sup>[198][]</sup> | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
-| [FunctionsConnected][] | ❌<sup>[456][]</sup> | ❌<sup>[198][]</sup> | ❌<sup>[456][]</sup> | ❌<sup>[456][]</sup> | ✅ | ✅ | ❌<sup>[456][]</sup> | ❌<sup>[456][]</sup> | ✅ | ✅ | ❌<sup>[456][]</sup> |
+| [FunctionsConnected][] | ✅ | ❌<sup>[198][]</sup> | ✅ | ❌<sup>[456][]</sup> | ✅ | ✅ | ❌<sup>[456][]</sup> | ✅ | ✅ | ✅ | ✅ |
```
mikesamuel, line 29: 🎉

mikesamuel (approved):
What would it take to have our backend test framework test how backends and translators deal with connected files?

tjpalmer:
> What would it take to have our backend test framework test how backends and translators deal with connected files?

Do you mean just the ...BackendTest suites? If so, it would be fairly easy to include them. I didn't feel like I would get much out of them, but it sounds like you would, so I can add some in future PRs, at least.

name/src/commonMain/kotlin/lang/temper/name/BackendId.kt
tjpalmer: > So this just keys into a lookup table for mime types so helps in picking accurate mime-types?

I don't think all backends even use them meaningfully. Looks like FileType.Module gets used in a couple of places. I'm not sure others do at all. But fileExtensionMap and mimeTypeMap in BackendMeta need them. We maybe could adjust in the future. But I needed something other than Module to list .hpp, and JsModule and Script both seemed very wrong.

PR #466, opened 2026-07-21
frontend: simplify uses of comparison operators

mikesamuel:
This is preparatory work to changing the IR of comparison operators.  Just making sure that non-overridden operators, especially the Int32 variants, are directly available.

Also, instead of using the `==` and `!=` cover functions in other desugarings, use simpler tests like `isNull`.

This required changing frontend and backend test goldens so I thought I'd get it out of the way first.

Specific changes:

- Coalescing operator, `??`, now uses `isNull` in its desugaring, not `!= null`.
- Comparison operators like `<` are no longer null accepting. be-rust no longer wraps f64 as options to compare them but `==` and `!=` still are.
- CoverFunction.uncover is now a static method so users like `new` and DotHelper no longer have to create CoverFunctions under the hood.
- Hardcoded ASTs in tests like ControlFlowTest use specific comparison operators, not the cover function values.


mikesamuel:
https://hackmd.io/@temper/SknfNmaVGl

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -275,9 +275,13 @@ private class Float64Compare(
         returnType: Type2,
         translator: RustTranslator,
     ): Rust.Expr {
+        val cmpFn = when (operator) {
+            RustOperator.Equals, RustOperator.NotEquals -> "cmp_option"
```
mikesamuel, line 279: I want `==` and `!=` for ints to eventually be non-null accepting, but for now, they still are.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -1196,6 +1225,40 @@ object BuiltinFuns {
         a % b
     }
 
+    val ltIntFn: CallableValue = IntCompareFun(
+        "<",
+        BuiltinOperatorId.LtIntInt,
+    ) { cmp ->
+        TBoolean.value(cmp < 0)
+    }
```
mikesamuel, line 1233: Exposing these out so that the non-overloaded version is readily available.

This will make it easier to desugar overloaded comparisons to a use of `<=>` nested inside a guaranteed int comparison of its output.

builtin/src/commonMain/kotlin/lang/temper/builtin/CoalesceMacro.kt
```diff
@@ -52,13 +52,12 @@ internal object CoalesceMacro : BuiltinMacro("??", null) {
             IfThenElse(
                 {
                     Call {
-                        V(Value(BuiltinFuns.notEqualsFn))
+                        V(vIsNullFn)
```
mikesamuel, line 55: One fewer use of a cover function.

format/src/commonMain/kotlin/lang/temper/format/FormattingHints.kt
```diff
@@ -255,6 +255,7 @@ private val spaceBetween = mapOf<Pair<Any, Any>, Boolean>(
     ("..." to OutputTokenType.Name) to false, // ...Int
     (OutputTokenType.Word to "(") to true, // if (
     (OutputTokenType.Name to "{") to true, // f {
+    (OutputTokenType.NumericValue to "{") to true, // if x == 0 {
```
mikesamuel, line 258: Rust was formatting oddly: `if x < 0{`.  Now there's a space before the bracket.

frontend/src/commonMain/resources/std/json/json.temper.md
```diff
@@ -1014,7 +1014,7 @@ where such a distinction is not possible to represent.
       var i = start;
       while (i.compareTo(limit) < 0) {
         let cp = sourceText[i];
-        let digit = if (char'0' <= cp && cp <= char'0') {
+        let digit = if (char'0' <= cp && cp <= char'9') {
```
mikesamuel, line 1017: 🤦

fundamentals/src/commonMain/kotlin/lang/temper/type/DotHelper.kt
```diff
@@ -234,8 +234,9 @@ class DotHelper(
             accessibleMembers.isEmpty() && extensions.isNotEmpty() &&
             extensions.all { (it as? FunctionResolution)?.fn is CallableValue }
         ) {
-            val cover = CoverFunction(extensions.map { (it as FunctionResolution).fn as CallableValue })
```
mikesamuel, line 237: Fewer creations of cover functions under the hood.

fundamentals/src/commonMain/kotlin/lang/temper/value/Actuals.kt
```diff
@@ -383,7 +383,9 @@ private class MacroActualsView(
         return if (uIndex >= 0) {
             underlying.result(uIndex, interpMode = interpMode, computeInOrder = computeInOrder)
         } else {
-            TODO("Link to a macro environment so we can evaluate the value tree")
+            val (_, newValue) = newValues[index]
+                ?: TODO("Link to a macro environment so we can evaluate the value tree $index, $interpMode")
+            newValue.valueContained ?: Fail
```
mikesamuel, line 388: `new` now cherry picks arguments to pass and the this argument reaches this path.

fundamentals/src/commonMain/kotlin/lang/temper/value/CoverFunction.kt
```diff
@@ -172,6 +110,72 @@ class CoverFunction(
...
         }
     }.toList()
+
+    companion object {
+        fun uncover(
+            args: Actuals,
+            cb: InterpreterCallback,
+            interpMode: InterpMode,
+            covered: List<CallableValue>,
+            otherwise: CallableValue? = null,
```
mikesamuel, line 120: Only change from above is these two arguments passed in instead of being accessed as fields, and a change to the recursive call at L129.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -2803,28 +2831,28 @@ class RustBackendTest {
             |        let mut n__0: i32 = 0;
             |        let mut i__0: usize = start__0;
             |        'loop___0: loop {
-            |            if ! (Some(Some(i__0).cmp( & Some(limit__0)) as i32) < Some(0)) {
+            |            if ! (Some(i__0).cmp( & Some(limit__0)) as i32 < 0) {
```
tjpalmer: Looks like the outer `Some(...)` was masking the now exposed `as i32 < 0`.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -441,7 +441,7 @@ class RustBackendTest {
             |}
             |pub fn hi(n__0: i32) {
             |    let mut i__0: i32 = 0;
-            |    'loop___0: while Some(i__0) < Some(n__0) {
+            |    'loop___0: while i__0 < n__0 {
```
tjpalmer, line 444: Losing the `Some` is nice.

tjpalmer (approved):

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
mikesamuel: Yeah.  Comparisons were just doing too much.  Uses of non-nullable types shouldn't pay for null-safety.

PR #467, opened 2026-07-21
be-rust: Require parens for as inside cmp

tjpalmer:
Address this error:

```rs
    --> C:\Users\work2\AppData\Local\Temp\temper-test3896171247360590248\rust\std\src\json\mod.rs:1782:64
     |
1782 |             if ! (Some(i__713).cmp( & Some(limit__710)) as i32 < 0) {
     |                                                                ^ -- interpreted as generic arguments
     |                                                                |
     |                                                                not interpreted as comparison
     |
help: try comparing the cast value
     |
1782 |             if ! ((Some(i__713).cmp( & Some(limit__710)) as i32) < 0) {
     |                   +                                            +
```

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustOperatorDefinition.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustOperatorDefinition.kt
@@ -28,6 +28,7 @@ enum class RustOperatorDefinition(
     // Copied from CSharpOperatorDefinition.
     override fun canNest(inner: OperatorDefinition, childIndex: Int) = when {
         inner !is RustOperatorDefinition -> false
+        this == Relational && inner == As -> false // `as T < ...` interpreted as generic arguments
         ordinal < inner.ordinal -> true
         ordinal > inner.ordinal -> false
         // TODO Does this allow `==` to nest?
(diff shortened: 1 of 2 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustOperatorDefinition.kt
```diff
@@ -28,6 +28,7 @@ enum class RustOperatorDefinition(
     // Copied from CSharpOperatorDefinition.
     override fun canNest(inner: OperatorDefinition, childIndex: Int) = when {
         inner !is RustOperatorDefinition -> false
+        this == Relational && inner == As -> false // `as T < ...` interpreted as generic arguments
```
tjpalmer, line 31: Might be overkill but shouldn't hurt to have extra parens sometimes.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -2831,7 +2831,7 @@ class RustBackendTest {
             |        let mut n__0: i32 = 0;
             |        let mut i__0: usize = start__0;
             |        'loop___0: loop {
-            |            if ! (Some(i__0).cmp( & Some(limit__0)) as i32 < 0) {
+            |            if ! ((Some(i__0).cmp( & Some(limit__0)) as i32) < 0) {
```
tjpalmer, line 2834: We had an outer `Some(...)` before that was masking the `as i32 < 0` issue that was now being exposed. This PR puts parens back.

PR #468, opened 2026-07-22
Rename implicits to core

tjpalmer:
- Also delete be-js connecteds supported test instead of updating it
- Don't yet rename `builtins` to `core`, and maybe shouldn't until/unless we can defer recognizing it


be-java/src/commonMain/kotlin/lang/temper/be/java/SimplifyNames.kt
```diff
@@ -44,7 +44,7 @@ fun simplifyNames(stmt: J.BlockLevelStatement): J.BlockLevelStatement {
 class SimplifyNames(private val top: J.TopLevelClassDeclaration) {
     private val importables = mutableListOf<Importable>()
     private val topLevel: Scope = Scope()
-    private val implicits = buildList {
+    private val core = buildList {
```
tjpalmer, line 47: Part of me worries that the name is less unique, but I didn't hit arbitrary name collisions in these changes. (I did hit some specifically on the implicits/core dual naming, but that'll be later in self review.)

be-js/src/commonTest/js/runtime-support/connecteds-supported-test.mjs
```diff
@@ -1,156 +0,0 @@
...
-import {describe, it} from 'mocha';
-import {expect} from 'chai';
-
-// The README for the backend package contains a list of connected methods that backends should
-// implement.
-
-const __dirname = dirname(fileURLToPath(import.meta.url));
-const connectedsList = ((content) => {
-    const [_, markdownList] =
-        /<!-- start ImplicitsModule-connected -->([\s\S]*)<!-- end ImplicitsModule-connected -->/
```
tjpalmer, line 16: I have an earlier commit with Implicits renamed to Core here, but then I just deleted the file, based on [this comment from @mikesamuel in discord](https://discord.com/channels/1108828567439163503/1195122951536771162/1529219236403216615):

> ...
> That test was super useful when I was initially working on the (now obsolete) connection mechanism. Now, our functional-test-suite has largely taken over the job of checking mandatory connections.
> All that to say, I agree, belatedly, that that test is more trouble than it's worth.

It would be better to delete in a separate PR, but then I was imagining merge conflicts, depending on the merge order. And this PR being core/std-related, this file isn't entirely off topic.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -349,7 +349,7 @@ fun standardResolver(name: ResolvedName) = when (name) {
 
 fun CodeLocation.source() = when (this) {
     is ModuleName -> this.sourceFile
-    is ImplicitsCodeLocation -> filePath("implicits.temper")
+    is CoreCodeLocation -> filePath("core.temper")
```
tjpalmer, line 352: That's an odd duck case for saying lowercase `implicits`.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -803,7 +803,7 @@ fun TmpL.Type.withoutAtom(predicate: (TmpL.Type) -> Boolean): TmpL.Type = when (
 private fun Iterable<TmpL.DeclarationMetadata>.documentationOf(): OccasionallyHelpful? =
     this.lookupMetaData(docStringSymbol, unpackVal = ::helpfulFromMetadataValue)
 
-/** An enum that allows switching on frequently used types from Implicits ignoring type parameters */
+/** An enum that allows switching on frequently used types from Core ignoring type parameters */
```
tjpalmer, line 806: And this is an odd duck case for saying title-case `Core`. I'm somewhat ok treating it as a proper name in some contexts. Either that or with backticks or other punctuation or extra words, again to make it extra clear sometimes.

be/src/commonMain/kotlin/lang/temper/be/FinishTmpLImports.kt
```diff
@@ -15,7 +15,7 @@ import lang.temper.name.ResolvedNameMaker
 import lang.temper.name.ResolvedParsedName
 import lang.temper.name.Temporary
 import lang.temper.value.TType
-import lang.temper.value.isImplicits
+import lang.temper.value.isCore
```
tjpalmer, line 18: I seriously considered saying something longer than `isCore`, but the shorter form worked fine. As I mentioned, it feels like there could be conflicts, but I didn't really hit any. Straight grep in source is probably worse now, though.

be/src/commonMain/kotlin/lang/temper/be/README.md
```diff
@@ -191,7 +191,7 @@ The set of *\@connected* methods from Temper's *ImplicitsModule* are:
 - `core.type StringIndexOption.compareTo()`
 - `core.type ValueResult`
 - `core.type ValueResult.constructor()`
-<!-- end ImplicitsModule-connected -->
+<!-- end CoreModule-connected -->
```
tjpalmer, line 194: I think this automatically updated file still has value for documentation.

build-user-docs/skeletal-docs/temper-docs/docs/tutorial/05-builtin.md
```diff
@@ -16,7 +16,7 @@ We also diverge here from our example geometric shapes library, though it
 served us well for context to get started.
 
 For the types below, some of them can be found as class or interface definitions
-in the file [temper/**/Implicits.temper] in the Temper source code. The
+in the file [temper/**/core.temper] in the Temper source code. The
```
tjpalmer, line 19: And I did lowercase the file name, since we've standardized on that for temper source file names.

build-user-docs/src/main/kotlin/lang/temper/docbuild/TypeShapeExtractor.kt
```diff
@@ -9,7 +9,7 @@ import lang.temper.common.putMultiList
 import lang.temper.common.splitAfter
 import lang.temper.common.toStringViaBuilder
 import lang.temper.format.OutToks
-import lang.temper.frontend.implicits.ImplicitsModule
+import lang.temper.frontend.core.CoreModule
```
tjpalmer, line 12: Being shorter is nice.

cli/src/main/resources/META-INF/native-image/resource-config.json
```diff
@@ -12,7 +12,7 @@
         "pattern": "\\QMETA-INF/services/kotlin.reflect.jvm.internal.impl.resolve.ExternalOverridabilityCondition\\E"
       },
       {
-        "pattern": "\\Qimplicits\\E"
+        "pattern": "\\Qcore\\E"
```
tjpalmer, line 15: Maybe this could end up keeping too much? But that wouldn't likely break anything, and we never got this working, anyway.

frontend/src/commonMain/kotlin/lang/temper/frontend/core/CoreModule.kt
```diff
@@ -75,14 +75,14 @@ object ImplicitsModule {
             )
         }
 
-        val content = loadResource(this, "implicits/Implicits.temper")
+        val content = loadResource(this, "core/core.temper")
```
tjpalmer, line 78: If we change our `core` processing, we could split this into multiple files, but I didn't chase that here.

frontend/src/commonMain/kotlin/lang/temper/frontend/Module.kt
```diff
@@ -752,18 +752,18 @@ class Module(
...
-        private val implicitsFilePositions get() = ImplicitsModule.implicitsFilePositions
+        private val coreFilePositions get() = CoreModule.coreFilePositions
     }
 
-    /** Whether this module represents Temper implicits/builtins, including for source editing purposes. */
-    val isEffectivelyImplicits = when (loc) {
-        // Recognize fake implicits, such as when editing them as ordinary files.
+    /** Whether this module represents Temper core/builtins, including for source editing purposes. */
+    val isEffectivelyCore = when (loc) {
+        // Recognize fake core, such as when editing them as ordinary files.
```
tjpalmer, line 760: I looked at implicits/core under both old and new regimes in our language server (which still uses way too much ram), and I got the same squiggly errors in both cases. I'm unsure I ever cleaned them all out before, and we've added more since. I did a side branch of cleaning some, but cleaning all would be beyond the scope of this PR, so I didn't worry that here.

frontend/src/commonMain/resources/core/config.temper.md
```diff
@@ -0,0 +1,6 @@
+# Temper Core
+
+Core definitions are partially internal to the Temper frontend, but some
+definitions are also defined here in Temper source.
+
+    export let name = "temper-core";
```
tjpalmer, line 6: This used to say "temper-implicits", but with my added comment and the dir rename, it saw this as a whole new file. This name is relevant in other changes in this PR, although I'm unsure we have any automated testing of it.

frontend/src/commonMain/resources/implicits/config.temper.md
```diff
@@ -1,3 +0,0 @@
-# Temper Implicits
-
-    export let name = "temper-implicits";
```
tjpalmer, line 3: The old name.

frontend/src/commonMain/resources/core/core.temper
tjpalmer, line 1: I guess we could have kept any name on this file, but "core.temper" seemed appropriate.

name/src/commonMain/kotlin/lang/temper/name/ModuleLocation.kt
```diff
@@ -328,9 +328,6 @@ data class DashedIdentifier(
          */
         const val temperCoreLibraryVersion = "0.6.0"
 
-        /** Specifically refers to the implicits library used by Temper code. */
-        val temperImplicitsLibraryIdentifier = DashedIdentifier("temper-implicits")
```
tjpalmer, line 332: This one evaporated. Automatic rename conflicts with `temperCoreLibraryIdentifier` that was already in this file. One of a couple of such situations. This name ought to match the `name` given in "config.temper.md" for core, which now says "temper-core". So I just changed dependent code (specifically, `isEffectivelyCore`) to use the single definition.

tooling/src/commonMain/kotlin/lang/temper/tooling/ModuleData.kt
```diff
@@ -668,17 +668,17 @@ private fun gatherDefs(
...
     }
-    // Implicits, letting these overwrite core where applicable.
-    val implicits = ImplicitsModule.module.exports?.asSequence()?.map { export ->
+    // Core module, letting these overwrite builtins where applicable.
+    val core = CoreModule.module.exports?.asSequence()?.map { export ->
         val name = export.name.baseName.nameText
         name to Mention.def(name = name, pos = export.position, text = name)
     } ?: emptySequence()
-    (core + implicits).toMap()
+    (builtins + core).toMap()
```
tjpalmer, line 681: And here was the other conflict, which again was specifically because of our explicit naming. At this point, it seemed more appropriate to rename the old `core` def here to `builtins`, letting the old `implicits` become `core`. If we do manage to rename `builtins` to `core` as well, the val def here might still keep the name `builtins` anyway.

mikesamuel (approved):

Issue #456, opened 2026-07-02
Userspace `@connected` methods


tjpalmer:
Add some decoration (maybe just `@connected`?) tells temper to copy in backend code to call instead of translating temper source.

(This is part of the request in #130, but that also suggests a different form of customizing per-backend logic.)

Some TODO items:

- [x] Backend tests to present better and to lock in translation somewhat
- [ ] Finalize translation/api decisions
- [ ] Frontend errors for unsupported features: rest params, generics, generators, anything but top-level functions (where we might be more flexible in the future, but I've worked within a subset for now just to make the common cases simple)
- [ ] Interp/fallback code
- [ ] Per-backend dependencies in config.temper.md
- [ ] Stable internal/unexported method names
- [ ] Verifying things work in standard builds, not just funtests
- [ ] What happens to null-default default args like `@connected let blah(a: String? = null);`
- [ ] Decorator to avoid pruning otherwise unexported things

tjpalmer:
Related PRs so far include #457, #465, and #470.

PR #473, opened 2026-08-10
Backend test connected code

tjpalmer:
- Work toward #456
- Fix some issues, especially paths relative to the library root
- Err on the side of avoiding major changes to backends in this pr
  - But some things demonstrated here might suggest possible further changes in the future
- Some changes:
  - Slightly move connected code injection point in be-lua
  - Count raw files for expanding `foo.py` to `foo/__init.__py` in be-py

### DIFF
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
@@ -135,10 +135,10 @@ class CppBackend private constructor(
 
         // Connected code.
         val connectedFiles = buildList {
+            val rootSize = libraryConfigurations.currentLibraryConfiguration.libraryRoot.segments.size
             for (file in rawBackendFiles) {
                 MetadataFileSpecification(
-                    // Skip the library name part.
-                    path = file.key.segments.subListToEnd(1).asFilePath(),
+                    path = file.key.segments.subListToEnd(rootSize).asFilePath(),
                     mimeType = MimeType.cppSource,
                     content = file.value,
                 ).also { add(it) }
--- a/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
+++ b/be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
@@ -3361,7 +3361,13 @@ class CppTranslator(
                         buildList {
                             add(cpp.include("$modPath$hppName"))
                             if (hasConnected) {
-                                add(cpp.includeLocal("_connected.hpp"))
+                                val connectedIncludePath = "_connected.hpp".let { connectedIncludeName ->
+                                    when {
+                                        relPath.segments.isEmpty() -> connectedIncludeName
+                                        else -> "$moduleBaseName/$connectedIncludeName"
+                                    }
+                                }
+                                add(cpp.includeLocal(connectedIncludePath))
                             }
                             addAll(namespaced(implVarDecls + impl))
                         },
--- a/be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
+++ b/be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
@@ -9,6 +9,7 @@ import lang.temper.lexer.Genre
 import lang.temper.log.filePath
 import kotlin.test.Test
 import kotlin.test.assertTrue
+import kotlin.text.trimMargin
 
 @SuppressWarnings("MaxLineLength")
 class CppBackendTest {
@@ -643,6 +644,148 @@ class CppBackendTest {
                 "void set_value(int32_t)",
                 "setBox",
             ),
+        )
+    }
+
+    @Test
+    fun connected() {
+        val connecteds = listOf(
+            filePath("something", "_connected.cpp") to """
+                    |#include "_connected.hpp"
+                    |#include "support.hpp"
+                    |
+                    |namespace work {
+                    |namespace _connected {
+                    |
+                    |std::int32_t sum(std::int32_t i, std::int32_t j, std::int32_t bonus) {
+                    |    return i + j + bonus;
+                    |}
+                    |
+                    |} // namespace _connected
+                    |} // namespace work
+                """.trimMargin(),
+            filePath("something", "_connected.hpp") to """
+                    |#pragma once
+                    |
+                    |#include <cstdint>
+                    |#include <work/init.hpp>
+                    |
+                    |namespace work {
+                    |namespace _connected {
+                    |
+                    |using namespace work;
+                    |
+                    |std::int32_t sum(std::int32_t i, std::int32_t j, std::int32_t bonus);
+                    |
+                    |} // namespace _connected
+                    |} // namespace work
+                """.trimMargin(),
+            filePath("other", "thing", "whatever.cpp") to """
+                    |// Content doesn't really matter here.
+                """.trimMargin(),
+        )
+        assertGeneratedCode(
+            backendConfig = Backend.Config.production,
+            factory = CppBackend.Cpp,
+            inputs = listOf(
+                filePath("something", "fun.temper") to """
+                    |let { prod } = import("./deeper");
+                    |
+                    |@connected
+                    |export let sum(i: Int, j: Int, bonus: Int = 0): Int;
+                    |
+                    |export let inc(i: Int): Int {
+                    |    sum(i, 1)
+                    |}
+                    |
+                    |export let twice(i: Int): Int {
+                    |  prod(i, 2)
+                    |}
+                """.trimMargin(),
+                filePath("something", "deeper", "more-fun.temper") to """
+                    |export let prod(i: Int, j: Int): Int {
+                    |    i * j
+                    |}
+                """.trimMargin(),
+            ) + connecteds,
+            want = """
+                |{
+                |    "cpp": {
+                |        "my-test-library": {
+                |            "something.cpp": {
+                |                content: ```
+                |                  #include <my-test-library/something.hpp>
+                |                  #include "something/_connected.hpp"
+                |                  namespace my_test_library {
+                |                    int32_t sum(int32_t i_4, int32_t j_5, temper::core::NullableParam<int32_t> bonus) {
+                |                      int32_t bonus_6;
+                |                      if(temper::core::is_null(bonus)) {
+                |                        {
+                |                          bonus_6 = 0;
+                |                        }
+                |                      }else {
+                |                        {
+                |                          bonus_6 = temper::core::not_null(bonus);
+                |                        }
+                |                      }
+                |                      return _connected::sum(i_4, j_5, bonus_6);
+                |                    }
+                |                    int32_t sum(int32_t i_4, int32_t j_5) {
+                |                      return sum(i_4, j_5, nullptr);
+                |                    }
+                |                    int32_t inc(int32_t i_8) {
+                |                      return sum(i_8, 1);
+                |                    }
+                |                    int32_t twice(int32_t i_10) {
+                |                      return my_test_library::prod(i_10, 2);
+                |                    }
+                |                    void global_init_something() {
+                |                      static bool initialized = false;
+                |                      if(initialized) {
+                |                        return;
+                |                      }
+                |                      initialized = true;
+                |                      my_test_library::global_init_deeper();
+                |                    }
+                |                  }
+                |
+                |                  ```
+                |            },
+                |            "something.hpp": {
+                |                "content": ```
+                |                  #pragma once
+                |                  #include <temper-core/core.hpp>
+                |                  #include <my-test-library/something/deeper.hpp>
+                |                  namespace my_test_library {
+                |                    int32_t sum(int32_t, int32_t, temper::core::NullableParam<int32_t>);
+                |                    int32_t sum(int32_t, int32_t);
+                |                    int32_t inc(int32_t);
+                |                    int32_t twice(int32_t);
+                |                    void global_init_something();
+                |                  }
+                |
+                |                  ```
+                |            },
+                |            "something": {
+                |                "_connected.cpp": "__DO_NOT_CARE__",
+                |                "_connected.hpp": "__DO_NOT_CARE__",
+                |                "deeper.cpp": "__DO_NOT_CARE__",
+                |                "deeper.hpp": "__DO_NOT_CARE__",
+                |                "deeper.hpp.map": "__DO_NOT_CARE__",
+                |                "deeper.cpp.map": "__DO_NOT_CARE__",
+                |            },
+                |            "something.cpp.map": "__DO_NOT_CARE__",
+                |            "something.hpp.map": "__DO_NOT_CARE__",
+                |            "main.cpp": "__DO_NOT_CARE__",
+                |            "other": {
+                |                "thing": {
+                |                    "whatever.cpp": "__DO_NOT_CARE__",
+                |                },
+                |            },
+                |        },
+                |    },
+                |}
+            """.trimMargin(),
         )
     }
 }
--- a/be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
+++ b/be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
@@ -4,6 +4,7 @@ import lang.temper.be.Backend
 import lang.temper.be.assertGeneratedCode
 import lang.temper.be.generateCode
 import lang.temper.common.ListBackedLogSink
+import lang.temper.common.stripDoubleHashCommentLinesToPutCommentsInlineBelow
 import lang.temper.fs.MemoryFileSystem
 import lang.temper.lexer.Genre
 import lang.temper.log.filePath
@@ -653,6 +654,7 @@ class CppBackendTest {
             backendConfig = Backend.Config.production,
             factory = CppBackend.Cpp,
             inputs = listOf(
+                // Test using a submodule.
                 filePath("something", "fun.temper") to """
                     |let { prod } = import("./deeper");
                     |export let twice(i: Int): Int {
@@ -670,6 +672,7 @@ class CppBackendTest {
                     |    i * j
                     |}
                 """.trimMargin(),
+                // Connected code needs explicit cpp and hpp files and explicit namespaces inside.
                 filePath("something", "_connected.cpp") to """
                     |#include "_connected.hpp"
                     |
@@ -707,6 +710,7 @@ class CppBackendTest {
                 |{
                 |    "cpp": {
                 |        "my-test-library": {
+                |## Submodule something gets translated for now as "something.cpp|hpp".
                 |            "something.cpp": {
                 |                content: ```
                 |                  #include <my-test-library/something.hpp>
@@ -761,6 +765,7 @@ class CppBackendTest {
                 |
                 |                  ```
                 |            },
+                |## Other submodule content goes in a "something" subdir but retains the namespace given.
                 |            "something": {
                 |                "_connected.cpp": "__DO_NOT_CARE__",
                 |                "_connected.hpp": "__DO_NOT_CARE__",
@@ -780,7 +785,7 @@ class CppBackendTest {
                 |        },
                 |    },
                 |}
-            """.trimMargin(),
+            """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
         )
     }
 }
--- a/be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
+++ b/be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
@@ -655,16 +655,14 @@ class CppBackendTest {
             inputs = listOf(
                 filePath("something", "fun.temper") to """
                     |let { prod } = import("./deeper");
+                    |export let twice(i: Int): Int {
+                    |  prod(i, 2)
+                    |}
                     |
                     |@connected
                     |export let sum(i: Int, j: Int, bonus: Int = 0): Int;
-                    |
                     |export let inc(i: Int): Int {
                     |    sum(i, 1)
-                    |}
-                    |
-                    |export let twice(i: Int): Int {
-                    |  prod(i, 2)
                     |}
                 """.trimMargin(),
                 filePath("something", "deeper", "more-fun.temper") to """
@@ -674,7 +672,6 @@ class CppBackendTest {
                 """.trimMargin(),
                 filePath("something", "_connected.cpp") to """
                     |#include "_connected.hpp"
-                    |#include "support.hpp"
                     |
                     |namespace work {
                     |namespace _connected {
@@ -690,7 +687,7 @@ class CppBackendTest {
                     |#pragma once
                     |
                     |#include <cstdint>
-                    |#include <work/init.hpp>
+                    |#include <my-test-library/something.hpp>
                     |
                     |namespace work {
                     |namespace _connected {
@@ -715,27 +712,27 @@ class CppBackendTest {
                 |                  #include <my-test-library/something.hpp>
                 |                  #include "something/_connected.hpp"
                 |                  namespace my_test_library {
-                |                    int32_t sum(int32_t i_4, int32_t j_5, temper::core::NullableParam<int32_t> bonus) {
-                |                      int32_t bonus_6;
+                |                    int32_t twice(int32_t i_4) {
+                |                      return my_test_library::prod(i_4, 2);
+                |                    }
+                |                    int32_t sum(int32_t i_6, int32_t j_7, temper::core::NullableParam<int32_t> bonus) {
+                |                      int32_t bonus_8;
                 |                      if(temper::core::is_null(bonus)) {
                 |                        {
-                |                          bonus_6 = 0;
+                |                          bonus_8 = 0;
                 |                        }
                 |                      }else {
                 |                        {
-                |                          bonus_6 = temper::core::not_null(bonus);
+                |                          bonus_8 = temper::core::not_null(bonus);
                 |                        }
                 |                      }
-                |                      return _connected::sum(i_4, j_5, bonus_6);
+                |                      return _connected::sum(i_6, j_7, bonus_8);
                 |                    }
-                |                    int32_t sum(int32_t i_4, int32_t j_5) {
-                |                      return sum(i_4, j_5, nullptr);
+                |                    int32_t sum(int32_t i_6, int32_t j_7) {
+                |                      return sum(i_6, j_7, nullptr);
                 |                    }
-                |                    int32_t inc(int32_t i_8) {
-                |                      return sum(i_8, 1);
-                |                    }
-                |                    int32_t twice(int32_t i_10) {
-                |                      return my_test_library::prod(i_10, 2);
+                |                    int32_t inc(int32_t i_10) {
+                |                      return sum(i_10, 1);
                 |                    }
                 |                    void global_init_something() {
                 |                      static bool initialized = false;
@@ -755,10 +752,10 @@ class CppBackendTest {
                 |                  #include <temper-core/core.hpp>
                 |                  #include <my-test-library/something/deeper.hpp>
                 |                  namespace my_test_library {
+                |                    int32_t twice(int32_t);
                 |                    int32_t sum(int32_t, int32_t, temper::core::NullableParam<int32_t>);
                 |                    int32_t sum(int32_t, int32_t);
                 |                    int32_t inc(int32_t);
-                |                    int32_t twice(int32_t);
                 |                    void global_init_something();
                 |                  }
                 |
--- a/be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
+++ b/be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
@@ -649,41 +649,6 @@ class CppBackendTest {
 
     @Test
     fun connected() {
-        val connecteds = listOf(
-            filePath("something", "_connected.cpp") to """
-                    |#include "_connected.hpp"
-                    |#include "support.hpp"
-                    |
-                    |namespace work {
-                    |namespace _connected {
-                    |
-                    |std::int32_t sum(std::int32_t i, std::int32_t j, std::int32_t bonus) {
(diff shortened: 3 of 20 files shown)


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppBackend.kt
```diff
@@ -135,10 +135,10 @@ class CppBackend private constructor(
 
         // Connected code.
         val connectedFiles = buildList {
+            val rootSize = libraryConfigurations.currentLibraryConfiguration.libraryRoot.segments.size
             for (file in rawBackendFiles) {
                 MetadataFileSpecification(
-                    // Skip the library name part.
-                    path = file.key.segments.subListToEnd(1).asFilePath(),
+                    path = file.key.segments.subListToEnd(rootSize).asFilePath(),
```
tjpalmer, line 141: This is one of the most common changes in this pr. The previous `1` worked for the funtest, but it doesn't work generally. Some testing of submodules in this pr is also related to path testing.

be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
```diff
@@ -3361,7 +3361,13 @@ class CppTranslator(
...
                             add(cpp.include("$modPath$hppName"))
                             if (hasConnected) {
-                                add(cpp.includeLocal("_connected.hpp"))
+                                val connectedIncludePath = "_connected.hpp".let { connectedIncludeName ->
+                                    when {
+                                        relPath.segments.isEmpty() -> connectedIncludeName
+                                        else -> "$moduleBaseName/$connectedIncludeName"
+                                    }
+                                }
+                                add(cpp.includeLocal(connectedIncludePath))
```
tjpalmer, line 3370: This deals with "something.cpp/hpp" vs "something/..." placement. I've made connected code go in the subdir without changing the cpp/hpp placement. But top-level is different, which is why the `isEmpty()` check. That case is being tested in the funtest, noting that I didn't have the qualified include in the old code here.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -645,6 +647,147 @@ class CppBackendTest {
...
     }
+
+    @Test
+    fun connected() {
+        assertGeneratedCode(
+            backendConfig = Backend.Config.production,
+            factory = CppBackend.Cpp,
+            inputs = listOf(
+                // Test using a submodule.
+                filePath("something", "fun.temper") to """
```
tjpalmer, line 658: Some places where already using `inputFileMapFromJson` for nested dirs, but I noticed that after making these changes here. I didn't go back and change to using it here or some other places. So my code in this pr has a mix of both. I'm sort of ok with that.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -645,6 +647,147 @@ class CppBackendTest {
...
+                    |    sum(i, 1)
+                    |}
+                """.trimMargin(),
+                filePath("something", "deeper", "more-fun.temper") to """
+                    |export let prod(i: Int, j: Int): Int {
+                    |    i * j
+                    |}
+                """.trimMargin(),
+                // Connected code needs explicit cpp and hpp files and explicit namespaces inside.
+                filePath("something", "_connected.cpp") to """
```
tjpalmer, line 676: I've left comments around in a variety of forms, although I didn't comment all of my thinking for the tests. Tried to gt some of the key points, at least.

Meanwhile, the backend tests vary across backends somewhat. I put more into what seemed more interesting in each case. I guess if we had a common source of such things for all backend tests, that still might be fine someday. And I might not always have chosen *the* most interesting things per backend in this pr, but hopefully I got sufficiently interesting things for now.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -645,6 +647,147 @@ class CppBackendTest {
...
+                    |
+                    |using namespace work;
+                    |
+                    |std::int32_t sum(std::int32_t i, std::int32_t j, std::int32_t bonus);
+                    |
+                    |} // namespace _connected
+                    |} // namespace work
+                """.trimMargin(),
+                filePath("other", "thing", "whatever.cpp") to """
+                    |// Content doesn't really matter here.
```
tjpalmer, line 706: Here I'm often just trying to see where things end up. And because backend tests don't run, it's possible I have mistakes in the raw target language code in some of these cases. Hopefully the content is sufficiently interesting at least to motivate things we might care about.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -645,6 +647,147 @@ class CppBackendTest {
...
+                |                      if(temper::core::is_null(bonus)) {
+                |                        {
+                |                          bonus_8 = 0;
+                |                        }
+                |                      }else {
+                |                        {
+                |                          bonus_8 = temper::core::not_null(bonus);
+                |                        }
+                |                      }
+                |                      return _connected::sum(i_6, j_7, bonus_8);
```
tjpalmer, line 733: I purposely included a function with a default arg in all backends so we can see the effect. Maybe only needed in one, but meh. In any case, seeing the actual connected calls here is interesting and varies somewhat by backend.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -645,6 +647,147 @@ class CppBackendTest {
...
+                |                    }
+                |                  }
+                |
+                |                  ```
+                |            },
+                |            "something.hpp": {
+                |                "content": ```
+                |                  #pragma once
+                |                  #include <temper-core/core.hpp>
+                |                  #include <my-test-library/something/deeper.hpp>
```
tjpalmer, line 757: Among other things, I wanted to see the include path here.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -645,6 +647,147 @@ class CppBackendTest {
...
+                |                  }
+                |
+                |                  ```
+                |            },
+                |## Other submodule content goes in a "something" subdir but retains the namespace given.
+                |            "something": {
+                |                "_connected.cpp": "__DO_NOT_CARE__",
+                |                "_connected.hpp": "__DO_NOT_CARE__",
+                |                "deeper.cpp": "__DO_NOT_CARE__",
+                |                "deeper.hpp": "__DO_NOT_CARE__",
```
tjpalmer, line 773: But in many cases, I just wanted to see where files went.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -645,6 +647,147 @@ class CppBackendTest {
...
+            ),
+            want = """
+                |{
+                |    "cpp": {
+                |        "my-test-library": {
+                |## Submodule something gets translated for now as "something.cpp|hpp".
+                |            "something.cpp": {
+                |                content: ```
+                |                  #include <my-test-library/something.hpp>
+                |                  #include "something/_connected.hpp"
```
tjpalmer, line 717: One of the interesting includes. I purposely used double-quote includes here. Even though the spec doesn't say what to do with these, I expect all modern compilers to cope with this for relative imports. In discussion with Shaw, he's suggested possibly separating header trees from source trees, among other things. If we did that, this might also need to change. So far, I've done it this way with the idea that it's an internal things, where we might keep internal headers with cpp source vs public headers that might be elsewhere. For now, everything is in one place. If we change layout, this might also need to change, depending on what changes we choose to make.

But I think this is ok here for the scope of this pr and lets us think about related matters.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpBackend.kt
```diff
@@ -186,9 +186,10 @@ class CSharpBackend(setup: BackendSetup<CSharpBackend>) : Backend<CSharpBackend>
                 }
             }
             // Connected source.
+            val rootSize = libraryConfig.libraryRoot.segments.size
             for (file in rawBackendFiles) {
                 // Skip the library name dir and the file name, and keep the middle
-                val subspace = chooseSubspace(file.key.segments).subList(1, file.key.segments.size - 1)
+                val subspace = chooseSubspace(file.key.segments).subList(rootSize, file.key.segments.size - 1)
```
tjpalmer, line 192: The whole 1 -> rootSize thing that will keep showing up in this pr.

be-csharp/src/commonTest/kotlin/lang/temper/be/csharp/CSharpBackendTest.kt
```diff
@@ -540,15 +544,15 @@ class CSharpBackendTest {
...
                 |export class Test(public var name: String) {
                 |  // Custom getter is nice for contrast with automated sometimes.
                 |  public get that(): String { thing() }
                 |  public thing(): String {
-                |    punctuate("Hi, ${"$"}{name}")
+                |    punctuate("Hi, ${name}")
                 |  }
                 |  private punctuate(message: String): String {
-                |    "${"$"}{message}!"
+                |    "${message}!"
```
tjpalmer, line 555: Recommended by ij.

be-csharp/src/commonTest/kotlin/lang/temper/be/csharp/CSharpBackendTest.kt
```diff
@@ -1498,6 +1502,98 @@ class CSharpBackendTest {
...
+                    |        return Sum(i__1, 1);
+                    |    }
+                    |}
+                    """.trimMargin(),
+                ),
+            ),
+            outputs = connecteds.map { (filePath, content) ->
+                // Automatically place files we expect to copy over.
+                // They go in same namespace dirs here as translated code.
+                val segments = buildList {
```
tjpalmer, line 1589: I did a fair amount of comments here. And note that csharp and java are rather like each other so far. Both still say `WhateverConnected` (rather than `_connected`, although maybe we should change that in the future). And both always just put connected code in the dir for the namespace/package/module, since that works well for them.

be-csharp/src/commonTest/kotlin/lang/temper/be/csharp/CSharpBackendTest.kt
```diff
@@ -1519,7 +1615,14 @@ private fun assertGenerateWanted(
     errors: List<String> = listOf(),
     moduleName: String = "test",
     namespaceName: String = moduleName.camelToPascal(),
+    outputs: List<Pair<FilePath, String>> = listOf(),
 ) {
+    // Build a tree-structured representation of outputs for easier nested handling.
+    val outputsFs = MemoryFileSystem()
+    for ((srcPath, content) in outputs) {
+        outputsFs.write(srcPath, content.toByteArray())
+    }
```
tjpalmer, line 1624: Backend testing for csharp has some extra automation compared to other backends, and I used and expanded that here rather than just using raw json testing. For good or ill. This is part of that expanded automation, and I just reached for the tools that seemed handiest that I ran across.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaBackend.kt
```diff
@@ -152,10 +154,15 @@ class JavaBackend private constructor(
             )
             add(result)
         }
+        // Copy raw java files into package dirs.
+        val javaLibConfig = JavaLibraryConfig(libraryConfigurations.currentLibraryConfiguration)
         for (file in rawBackendFiles) {
-            // Just copy raw java files into package dirs.
+            val modulePath = file.key.segments.subList(0, file.key.segments.size - 1).asFilePath()
+            val packageName = javaLibConfigs.packageNameFor(modulePath, javaLibConfig)
```
tjpalmer, line 161: Logic in be-java for file placement was somewhat factored, so I tried to use that here. It ends up a bit messier looking than other backends, but if I did it right, the code here should be well tied to prior be-java code placement logic.

be-java/src/commonTest/kotlin/lang/temper/be/java/JavaBackendTest.kt
```diff
@@ -2142,6 +2142,91 @@ class JavaBackendTest {
         """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
     )
 
+    @Test
+    fun connected() = assertGeneratedJavaRaw(
+        inputs = inputFileMapFromJson(
+            // Test using a submodule.
+            """
+                |{
+                |  sub: {
```
tjpalmer, line 2151: Submodule here still also.

be-js/src/commonMain/kotlin/lang/temper/be/js/JsBackend.kt
```diff
@@ -221,9 +221,10 @@ class JsBackend private constructor(
             }
         }
         // Connected files.
+        val rootSize = libraryConfigurations.currentLibraryConfiguration.libraryRoot.segments.size
         val connectedFiles = rawBackendFiles.map { file ->
             MetadataFileSpecification(
-                path = FilePath(file.key.segments.subListToEnd(1), isDir = false),
+                path = FilePath(file.key.segments.subListToEnd(rootSize), isDir = false),
```
tjpalmer, line 227: And be-js is back to being simple here.

be-js/src/commonMain/kotlin/lang/temper/be/js/JsTranslator.kt
```diff
@@ -210,13 +210,16 @@ internal class JsTranslator(
...
+            // All except top-level connecteds go down a subdir along with temper-built submodules.
+            val dirName = relativePath.segments.lastOrNull()?.fullName?.let { "./$it" } ?: "."
             Js.ImportDeclaration(
                 t.pos,
                 specifiers = Js.ImportNamespaceSpecifier(
                     t.pos,
                     Js.Identifier(t.pos, connectedName, null),
                 ).let { listOf(it) },
-                source = Js.StringLiteral(t.pos, "./_connected.js"),
+                source = Js.StringLiteral(t.pos, "$dirName/_connected.js"),
```
tjpalmer, line 222: This is approximately the same logic as in be-cpp changes in this pr that imports from a subdir if not for the root module of a library.

be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
```diff
@@ -2454,6 +2454,138 @@ class JsBackendTest {
...
+                |
+                |        @connected
+                |        export let sum(i: Int, j: Int, bonus: Int = 0): Int;
+                |        export let inc(i: Int): Int {
+                |            sum(i, 1)
+                |        }
+                |        ```,
+                |      _connected.js: ```
+                |        // Importing with "../" required here. See layout later for more.
+                |        import "../foo.internal.js";
```
tjpalmer, line 2479: And for the root module, you'd say `./whatever.internal.js`, as seen in the earlier functional test:

https://github.com/temperlang/temper/blob/18c0cf0383381175e4224d87bffbf3e87896e5ee/functional-test-suite/src/commonMain/resources/functions/connected/_connected.js#L2

This is an example of something ugly we'll need to document if we don't adjust translated js file placement in the future, such as always saying `foo/index.js` instead of `foo.js`.

be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
```diff
@@ -2454,6 +2454,138 @@ class JsBackendTest {
...
+            |        },
+            |## As for cpp, we currently put foo(.internal?).js at the upper level.
+            |        "foo.js": "__DO_NOT_CARE__",
+            |## And because we don't export internal modules in package.json, we can't
+            |## do a outer-namespaced import of the internal module, so we currently
+            |## have to pay attention to when we need to "../" import things from this
+            |## subdir. Maybe we can always make "foo/index.js" in the future instead.
+            |        "foo": {
+            |          "_connected.js": "__DO_NOT_CARE__",
+            |          "deeper.js": "__DO_NOT_CARE__",
```
tjpalmer, line 2555: Above is more explanation on why the imports are awkward.

be-js/src/commonTest/kotlin/lang/temper/be/js/JsBackendTest.kt
```diff
@@ -2454,6 +2454,138 @@ class JsBackendTest {
...
+            |      },
+            |      "package.json": {
+            |        jsonContent: {
+            |          "name": "my-test-library",
+            |          "type": "module",
+            |          "exports": {
+            |            // Showing json to emphasize what's exported, which excludes connected.
+            |            "./src/foo/deeper": "./src/foo/deeper.js",
+            |            "./src/foo": "./src/foo.js",
+            |            ".": "./index.js"
```
tjpalmer, line 2576: These exports also exclude `.internal.js` files.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaBackend.kt
```diff
@@ -97,10 +97,10 @@ class LuaBackend private constructor(
                 // Special _connected.lua file gets inlined.
                 file.key.last().fullName == "_connected.lua" && continue@rawBackendFiles
                 // Others get copied.
-                // TODO Make sure things get into good places.
+                val rootSize = libraryConfigurations.currentLibraryConfiguration.libraryRoot.segments.size
```
tjpalmer, line 100: I think we are doing at least much better about right places with the changes in this pr.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaTranslator.kt
```diff
@@ -2515,11 +2515,11 @@ private class ModuleParts(
         val withLocals = luaChunk(
             pos,
             buildList {
-                addAll(imports)
-                addAll(preDecls)
                 if (connectedSource != null) {
                     add(Lua.Connected(pos, connectedSource))
                 }
+                addAll(imports)
+                addAll(preDecls)
```
tjpalmer, line 2522: I felt like moving the connected code higher. This means just temper-core import and top-level local *decls* (not assignments/definitions) go before connected code now. We'll see more of that in backend testing.

be-lua/src/commonTest/kotlin/lang/temper/be/lua/LuaBackendTest.kt
```diff
@@ -364,6 +366,185 @@ class LuaBackendTest {
...
+                |export let inc(i: Int): Int {
+                |  sum(i, 1)
+                |}
+                |test("sum") {
+                |  assert(sum(1, 2) == 3);
+                |}
+                |let sum(i: Int, j: Int): Int {
+                |  console.log("hi");
+                |  i + j
+                |}
```
tjpalmer, line 383: I added this testing backend test because we didn't have any such for be-lua so far, and I wanted to have it here for reference.

be-lua/src/commonTest/kotlin/lang/temper/be/lua/LuaBackendTest.kt
```diff
@@ -364,6 +366,185 @@ class LuaBackendTest {
...
+            |        "my-test-library": {
+            |            "something.lua": {
+            |                "content": ```
+            |## Use absolute imports consistently in lua to help proper caching.
+            |                  local imports = require('my-test-library/something-internal');
+            |                  return {inc = imports.inc};
+            |
+            |                  ```
+            |            },
+            |            "something-internal.lua": {
```
tjpalmer, line 398: At the moment, I especially wanted to see the name of this file and how we were importing it. Unlink be-js, I haven't changed be-lua to always split out an internal. It happens as needed still for now.

be-lua/src/commonTest/kotlin/lang/temper/be/lua/LuaBackendTest.kt
```diff
@@ -364,6 +366,185 @@ class LuaBackendTest {
...
+            |                  return exports;
+            |
+            |                  ```
+            |            },
+            |            "tests": {
+            |                "something-test.lua": {
+            |                    content: ```
+            |                      local temper = require('temper-core');
+            |                      local sum__0, local_4, local_5, exports;
+            |                      sum__0 = temper.import('my-test-library/something-internal', 'sum__0');
```
tjpalmer, line 422: And note that imports in be-lua are always absolute paths in nature rather than relative. Some comments on that below.

be-lua/src/commonTest/kotlin/lang/temper/be/lua/LuaBackendTest.kt
```diff
@@ -364,6 +366,185 @@ class LuaBackendTest {
...
+                |    i * j
+                |}
+            """.trimMargin(),
+            filePath("something", "_connected.lua") to """
+                |## Only `_connected` should be defined at top level here.
+                |local _connected = {}
+                |do
+                |## Show example require in connected code, even if we don't use it here.
+                |## And note that absolute rather than relative imports work here.
+                |  local s = require("my-test-library.something._support")
```
tjpalmer, line 483: Again, more on absolute imports. Lua wants the import string to be the same each time for each module for easier caching.

be-lua/src/commonTest/kotlin/lang/temper/be/lua/LuaBackendTest.kt
```diff
@@ -364,6 +366,185 @@ class LuaBackendTest {
...
+        want = """
+            |{
+            |    "lua": {
+            |        "my-test-library": {
+            |            "something.lua": {
+            |                "content": ```
+            |                  local temper = require('temper-core');
+            |                  local prod, twice, sum, inc, exports;
+            |## Inline connected code *after* locals are defined to allow access from connected code.
+            |                  local _connected = {}
```
tjpalmer, line 503: A little example also here on how local decl positions matter in lua:

```lua
> local function f(); print(i); end; local i = 5; f()
nil
> local i; local function f(); print(i); end; i = 5; f()
5
```

And we want to be able to call module members from connected code.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyBackend.kt
```diff
@@ -306,13 +308,19 @@ class PyBackend private constructor(
...
+                add(FilePathSegment(pyLibraryName.text))
+                addAll(file.key.segments.subListToEnd(rootSize))
+            }.asFilePath()
+            topModule.setOutputFile(
+                MetadataFileSpecification(
+                    path = path,
+                    mimeType = mimeType,
+                    content = file.value,
+                ),
+            )
```
tjpalmer, line 323: Python had to get fancy here to get things in reliable places at all. Maybe should just change all `blah.py` to `blah/__init__.py`, but I didn't do that here. I *did* make it so the presence of raw backend code causes that expansion, though. More on that below.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyBackend.kt
```diff
@@ -329,7 +337,7 @@ class PyBackend private constructor(
         }
         topModule.mapNotNullTo(outputFileSpecifications) { mod ->
             if (mod.moduleId != null) {
-                mod.write().toTreeFile()
+                mod.toOutputFile()
```
tjpalmer, line 340: I needed this slightly more abstracted now to accommodate either raw backend files or translated files.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyBackend.kt
```diff
@@ -441,7 +449,7 @@ class PyBackend private constructor(
                                 val relImportPath = listOf("", program.outputPath.segments.last().baseName)
                                 Py.ImportWildcardFrom(
                                     unknownPos,
-                                    Py.ImportDotted(unknownPos, PyDottedIdentifier.dotted(relImportPath)),
+                                    Py.ImportDotted(unknownPos, dotted(relImportPath)),
```
tjpalmer, line 452: Recommended by ij.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyModule.kt
```diff
@@ -25,14 +30,26 @@ class PyModule(var moduleId: PyDottedIdentifier?) : Iterable<PyModule> {
...
+        node.program = program
+        return node
+    }
+
+    /** Intending `this` for a top level module, set a program based on its path. */
+    fun setOutputFile(outputFile: Backend.OutputFileSpecification): PyModule {
+        val node = expand(outputFile.path)
+        node.outputFile = outputFile
+        return node
+    }
```
tjpalmer, line 45: So now you can set either one or the other. I could have used some sealed class or interface, but instead I was sloppy and had both fields. But the idea is you'd set either one or the other.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyModule.kt
tjpalmer, line 118: And this now notices if you have raw backend files present.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -699,6 +699,87 @@ class PyBackendTest {
...
+            |      class _connected:
+            |          from ._support import Support
+            |
+            |          def sum(i: int, j: int, bonus: int) -> int:
+            |              return i + j + bonus
+            |
+            |      ```,
+            |## Include this bonus file *without* an explicit temper module at the
+            |## same level to prove we still get a dir for the translated module.
+            |    _support.py: ```
```
tjpalmer, line 727: Meaning this raw file is the only thing causing `foo/__init__.py` to happen.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -699,6 +699,87 @@ class PyBackendTest {
...
+            |}
+            """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
+        ),
+        want = """
+            |{
+            |  py: {
+            |    my-test-library: {
+            |      my_test_library: {
+            |        foo: {
+            |          __init__.py: {
```
tjpalmer, line 741: And the subdir does happen.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -699,6 +699,87 @@ class PyBackendTest {
...
+            |                      bonus_4 = _bonus_8
+            |## Here's the connected call.
+            |                  return _connected.sum(i_2, j_3, bonus_4)
+            |              def inc(i_6: 'int1', /) -> 'int1':
+            |                  return sum(i_6, 1)
+            |
+            |              ```,
+            |          },
+            |## And here's the support file next to the primary module.
+            |          _support.py: "__DO_NOT_CARE__",
```
tjpalmer, line 771: Here's the file it had to accommodat.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -186,6 +187,92 @@ class RustBackendTest {
...
+            |
+            |              ```
+            |          },
+            |          _connected: {
+            |## Raw rust code goes into this subdir so other raw files are below it.
+            |## The original "_connected.rs" gets renamed to "mod.rs", which leaves it
+            |## effectively still just named `_connected`, and "whatever.rs" is now
+            |## `_connected::whatever` relative to anything above.
+            |            mod.rs: "__DO_NOT_CARE__",
+            |            whatever.rs: "__DO_NOT_CARE__",
```
tjpalmer, line 264: The comment above is why I do the `_connected/` dir. I retried not doing that while working this pr, and it would cause trouble, including that I'd have to inject mod declarations in the translated module for everything raw rust source file. And that seemed overmuch. So people just have to expect this layout, and we'll need to document it.

be-test-helpers/src/commonMain/kotlin/lang/temper/be/GeneratedCodeHelper.kt
```diff
@@ -187,6 +189,10 @@ fun <BACKEND : Backend<BACKEND>> generateCode(
             } else {
                 NullSystemAccess(outputRoot.path.resolve(backendLib), cancelGroup)
             }
+            val extensions = factory.backendMeta.fileExtensionMap.values.toSet()
+            val rawBackendFiles = sourceTree.gatherFilesSync { filePath ->
+                filePath.lastOrNull()?.extension?.let { it in extensions } == true
+            }
```
tjpalmer, line 195: I still haven't implemented this for cli build yet, but this now covers the backend test case. Before, only funtests gathered raw backend files.

be-test-helpers/src/commonMain/kotlin/lang/temper/be/GeneratedCodeHelper.kt
```diff
@@ -216,3 +223,24 @@ fun <BACKEND : Backend<BACKEND>> generateCode(
...
+
+/**
+ * Map file paths to the textual content of files. Only [keep] files that can be
+ * decoded as UTF-8.
+ */
+fun FileSystem.gatherFilesSync(keep: (FilePath) -> Boolean): Map<FilePath, String> = buildMap {
+    fun addFiles(path: FilePath) {
+        when (classify(path)) {
+            FileClassification.File -> if (keep(path)) {
+                put(path, readBinaryFileContentSync(path).result!!.decodeToString())
```
tjpalmer, line 235: Everything else in this area is already sync, so this seemed ok here, too.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.lua
```diff
@@ -1,4 +1,4 @@
-local _connected = {};
+local _connected = {}
```
tjpalmer, line 1: Just a small cleanup. Lua doesn't typically have semicolons, and this is just a manual file that I accidentally put a semi in.

mikesamuel (approved):

PR #475, opened 2026-08-19
Add checks that overrides are compatible with overridden signatures

tjpalmer:
- Continues #350. Maybe should have updated that PR, but I didn't think about that when first posting this draft PR.


mikesamuel (approved):

mikesamuel:
It's still in draft, but CI/CD is green. If you want to do more drafty stuff just let me know and I'll rereview when it comes out of draft.

tjpalmer:
> It's still in draft, but CI/CD is green. If you want to do more drafty stuff just let me know and I'll rereview when it comes out of draft.

I called it draft because I'm not entirely happy with all the naming, but we could potentially merge this for now and change naming later.

Part of my sadness is having to say something longer than `.next()` for what is so far the common case of no throwing.

PR #476, opened 2026-08-26
Check connected limits as they stand today

tjpalmer:
- Relates to #456
- Report error on user `@connected` except for top-level functions
- Don't check for either core or std, and move some std checking helper logic
- Call decorator macros only once each, even with multiple insertion points
- Test error cases
- In existing tests, allow previously applied cases that we don't really support to give errors for now
  - There's actually one ("rewrite-connected-decorator") syntax macro stage test (at least) that fakes being std to allow connected even in the past, so that's a nice side test for the std dodge, especially since we don't fail funtests with compiler errors in std

### DIFF
--- a/be/src/commonTest/kotlin/lang/temper/be/tmpl/TmpLBackendTest.kt
+++ b/be/src/commonTest/kotlin/lang/temper/be/tmpl/TmpLBackendTest.kt
@@ -2270,7 +2270,13 @@ class TmpLBackendTest {
             |    },
             |    "defines.tmpl.map": "__DO_NOT_CARE__",
             |    "uses.tmpl.map": "__DO_NOT_CARE__",
-            |  }
+            |  },
+            |  errors: [
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |  ]
             |}
         """.trimMargin(),
         supportNetwork = defaultTestSupportNetwork.copy(
@@ -2316,7 +2322,13 @@ class TmpLBackendTest {
             |    },
             |    "defines.tmpl.map": "__DO_NOT_CARE__",
             |    "uses.tmpl.map": "__DO_NOT_CARE__",
-            |  }
+            |  },
+            |  errors: [
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |  ]
             |}
         """.trimMargin(),
         supportNetwork = defaultTestSupportNetwork.copy(
@@ -2363,7 +2375,13 @@ class TmpLBackendTest {
             |    },
             |    "defines.tmpl.map": "__DO_NOT_CARE__",
             |    "uses.tmpl.map": "__DO_NOT_CARE__",
-            |  }
+            |  },
+            |  errors: [
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |  ]
             |}
         """.trimMargin(),
         supportNetwork = defaultTestSupportNetwork.copy(
@@ -2434,6 +2452,13 @@ class TmpLBackendTest {
             |    },
             |    foo.tmpl.map: "__DO_NOT_CARE__",
             |  },
+            |  errors: [
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |  ]
             |}
         """.trimMargin(),
     )
@@ -3624,7 +3649,12 @@ class TmpLBackendTest {
             |        ```
             |    },
             |    "foo.tmpl.map": "__DO_NOT_CARE__",
-            |  }
+            |  },
+            |  errors: [
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |  ]
             |}
         """.trimMargin().stripDoubleHashCommentLinesToPutCommentsInlineBelow(),
     )
--- a/frontend/src/commonMain/kotlin/lang/temper/frontend/define/AddTemperTestInstructionsTo.kt
+++ b/frontend/src/commonMain/kotlin/lang/temper/frontend/define/AddTemperTestInstructionsTo.kt
@@ -3,7 +3,7 @@ package lang.temper.frontend.define
 import lang.temper.builtin.BuiltinFuns
 import lang.temper.common.Log
 import lang.temper.frontend.Module
-import lang.temper.frontend.isEffectivelyStd
+import lang.temper.interp.importExport.isEffectivelyStd
 import lang.temper.log.MessageTemplate
 import lang.temper.name.BuiltinName
 import lang.temper.name.ExportedName
--- a/frontend/src/commonTest/kotlin/lang/temper/frontend/SyntaxMacroStageTest.kt
+++ b/frontend/src/commonTest/kotlin/lang/temper/frontend/SyntaxMacroStageTest.kt
@@ -260,6 +260,12 @@ class SyntaxMacroStageTest {
     )
 
     @Test
+    fun connectedUnsupported() = assertModuleAtStage(
+        stageTestDir = StageTestDir("syntax-macro/connected-unsupported"),
+        moduleResultNeeded = true,
+    )
+
+    @Test
     fun reorder() = assertModuleAtStage(
         stageTestDir = StageTestDir("syntax-macro/reorder"),
     )
--- a/frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/syntax-macro/connected-unsupported/expect/errors.json
+++ b/frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/syntax-macro/connected-unsupported/expect/errors.json
@@ -1,0 +1,5 @@
+[
+    "At this time, users can connect only functions!",
+    "At this time, users can connect only functions!",
+    "At this time, rest parameters aren't supported in user connected functions!"
+]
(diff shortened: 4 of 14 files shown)


be/src/commonTest/kotlin/lang/temper/be/tmpl/TmpLBackendTest.kt
```diff
@@ -2270,7 +2270,13 @@ class TmpLBackendTest {
...
             |    "defines.tmpl.map": "__DO_NOT_CARE__",
             |    "uses.tmpl.map": "__DO_NOT_CARE__",
-            |  }
+            |  },
+            |  errors: [
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |    "At this time, users can connect only top-level functions!",
+            |  ]
```
tjpalmer, line 2279: Currently unsupported things were already being tested even before any user connected function support. I think it's ok just to allow these errors until such time as we support more connected code.

frontend/src/commonMain/kotlin/lang/temper/frontend/define/AddTemperTestInstructionsTo.kt
```diff
@@ -3,7 +3,7 @@ package lang.temper.frontend.define
 import lang.temper.builtin.BuiltinFuns
 import lang.temper.common.Log
 import lang.temper.frontend.Module
-import lang.temper.frontend.isEffectivelyStd
+import lang.temper.interp.importExport.isEffectivelyStd
```
tjpalmer, line 6: Moved to `interp` for access from ConnectedDecorator.

frontend/src/commonTest/kotlin/lang/temper/frontend/SyntaxMacroStageTest.kt
```diff
@@ -259,6 +259,12 @@ class SyntaxMacroStageTest {
         pseudoCodeDetail = PseudoCodeDetail.default.copy(showTypeMemberMetadata = true),
     )
 
+    @Test
+    fun connectedUnsupported() = assertModuleAtStage(
+        stageTestDir = StageTestDir("syntax-macro/connected-unsupported"),
+        moduleResultNeeded = true,
+    )
```
tjpalmer, line 266: Added this before learning we already had violations elsewhere, but this test is more precise about checking various unsupported cases.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/syntax-macro/connected-unsupported/expect/errors.json
```diff
@@ -0,0 +1,7 @@
+[
+    "At this time, users can connect only top-level functions!",
+    "At this time, users can connect only top-level functions!",
+    "At this time, users can connect only top-level functions!",
+    "At this time, users can connect only top-level functions!",
+    "At this time, rest parameters aren't supported in user connected functions!"
+]
```
tjpalmer, line 7: Maybe should add a newline to avoid annoying glyphs.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/syntax-macro/connected-unsupported/work/test/test.temper
```diff
@@ -0,0 +1,20 @@
...
+
+  @connected
+  public static notEvenStaticMethodsYet();
+}
+
+@connected
+export let alsoNoRestArgsYet(i: Int, ...more: List<Int>);
+
+@connected
+export let alsoNoOtherDeclsYet: Int = 5;
```
tjpalmer, line 14: Various currently unsupported cases above and one currently supported case below for contrast.

fundamentals/src/commonMain/kotlin/lang/temper/value/MacroEnvironment.kt
```diff
@@ -180,6 +181,12 @@ interface MacroEnvironment : InterpreterCallback, Positioned, ConfigurationKey.H
      */
     val isProcessingCore: Boolean get() = false
 
+    /**
+     * Depending on the situation, [sharedLocationContext] might be needed for
+     * accurate results.
+     */
+    fun isProcessingStd(sharedLocationContext: SharedLocationContext?): Boolean
```
tjpalmer, line 188: For more convenient access from ConnectedDecorator. I left the parameter here even though I don't need it right now because I want it to be a conscious choice to pass in null.

interp/src/commonMain/kotlin/lang/temper/interp/importExport/ResolvedModuleSpecifier.kt
```diff
@@ -17,3 +23,21 @@ val STANDARD_LIBRARY_FILEPATH = FilePath(listOf(FilePathSegment(STANDARD_LIBRARY
...
+    }
+    if (this is FileRelatedCodeLocation) {
+        // Recognize fake std, such as when editing them as ordinary files.
+        val sourceFile = this.sourceFile
+        if (sourceFile.segments.firstOrNull()?.fullName == STANDARD_LIBRARY_NAME) {
+            return true
+        }
+    }
+    return false
+}
```
tjpalmer, line 43: Just moved this to where `STANDARD_LIBRARY_NAME` is already defined, making it available to ConnectedDecorator.

interp/src/commonMain/kotlin/lang/temper/interp/ConnectedDecorator.kt
```diff
@@ -21,8 +26,40 @@ import lang.temper.value.void
...
     connectedSymbol,
     findDecoratorInsertions = ::findConnectedDecoratorInsertions,
-) {
-    void
+) { args ->
+    // Whether we have errors or not, move on with a void value.
+    void.also check@{
+        isProcessingCore && return@check
+        // At this stage, we don't need the location context.
+        isProcessingStd(sharedLocationContext = null) && return@check
```
tjpalmer, line 34: Check when not in core or std. And apparently some other `@connected` checking already happens elsewhere, based on some other testing. But it still feels not terrible to check a macro's applicability while in the macro itself.

interp/src/commonMain/kotlin/lang/temper/interp/ConnectedDecorator.kt
```diff
@@ -21,8 +26,40 @@ import lang.temper.value.void
...
+        val metadata = (args.rawTreeList.first() as? DeclTree)?.parts?.metadataSymbolMap ?: return@check
+        when {
+            typeDeclSymbol in metadata -> {
+                log(Log.Error, MessageTemplate.UserConnectedNotFun, pos, listOf())
```
tjpalmer, line 38: I could give a more precise error here about not being allowed on types, but this gets the job done.

interp/src/commonMain/kotlin/lang/temper/interp/ConnectedDecorator.kt
```diff
@@ -21,8 +26,40 @@ import lang.temper.value.void
...
+                    init.children.any { maybeParam ->
+                        when (val maybeParamParts = (maybeParam as? DeclTree)?.parts) {
+                            null -> false
+                            else -> restFormalSymbol in maybeParamParts.metadataSymbolMap
+                        }
+                    } -> {
+                        log(Log.Error, MessageTemplate.UserConnectedFunHasRest, pos, listOf())
+                    }
+                    else -> {
+                        // We support connected functions without rest params at this time.
```
tjpalmer, line 54: Here's the ok case with a comment.

interp/src/commonMain/kotlin/lang/temper/interp/ConnectedDecorator.kt
```diff
@@ -21,8 +26,40 @@ import lang.temper.value.void
...
+                    // Neither instance nor static methods get in here because of different tree structures.
+                    // That's good for now.
+                    // Also, we don't get formal params yet when this macro is called, so loop trees.
+                    init.children.any { maybeParam ->
+                        when (val maybeParamParts = (maybeParam as? DeclTree)?.parts) {
+                            null -> false
+                            else -> restFormalSymbol in maybeParamParts.metadataSymbolMap
+                        }
+                    } -> {
+                        log(Log.Error, MessageTemplate.UserConnectedFunHasRest, pos, listOf())
```
tjpalmer, line 51: The one precise case I give an error for at the moment, because it doesn't get as easily swept into a simple explanation.

interp/src/commonMain/kotlin/lang/temper/interp/MetadataDecorator.kt
```diff
@@ -80,11 +80,16 @@ class MetadataDecorator(
...
+        var valuerResultCache: PartialResult? = null
         for (insertion in insertions) {
             val (decorated, insertionPoint) = insertion
             if (insertionPoint in 0..decorated.size) {
-                val calleePos = macroEnv.callee.pos
-                when (val valuerResult = macroEnv.valuer(args)) {
+                val valuerResult = when (valuerResultCache) {
+                    null -> macroEnv.valuer(args).also { valuerResultCache = it }
+                    else -> valuerResultCache
+                }
```
tjpalmer, line 91: Prevent calling the macro multiple times for a single application, just because of multiple insertion points. And apparently so far our Value instances don't seem to stress out about this. I think that's likely to be resilient against future changes also.

interp/src/commonMain/kotlin/lang/temper/interp/MetadataDecorator.kt
```diff
@@ -96,6 +101,7 @@ class MetadataDecorator(
                         return valuerResult
                     }
                     is Value<*> -> {
+                        val calleePos = macroEnv.callee.pos
```
tjpalmer, line 104: Noticed this was only being used down here anyway.

log/src/commonMain/kotlin/lang/temper/log/MessageTemplate.kt
```diff
@@ -233,6 +233,14 @@ enum class MessageTemplate(
...
         CompilationPhase.CodeGeneration,
     ),
+    UserConnectedNotFun(
+        "At this time, users can connect only top-level functions",
+        CompilationPhase.CodeGeneration,
+    ),
+    UserConnectedFunHasRest(
+        "At this time, rest parameters aren't supported in user connected functions",
+        CompilationPhase.CodeGeneration,
+    ),
```
tjpalmer, line 243: One general and one precise message, both mentioning "user".

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/syntax-macro/connected-unsupported/expect/errors.json
mikesamuel: 🤷

mikesamuel (approved):

PR #478, opened 2026-09-01
Fix some rust things for new result handling

tjpalmer:
- Add result ops to support network
- Expect `return void` instead of just `return`
- Delete some commented code
- Update rust backend test
- Lots of funtests still fail, I think many because of receiving bad type info
- The async backend test also fails, at least in part because an `await` op is being received


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -1124,6 +1124,25 @@ private val timesIntInt = MethodCall("TimesIntInt", "wrapping_mul", BuiltinOpera
...
+    baseName = "IsOkResult",
+    memberName = "is_ok",
+    builtinOperatorId = BuiltinOperatorId.IsOkResult,
+)
+
+private object PackOkResult : FunctionCall(
+    baseName = "PackOkResult",
+    functionName = "Ok",
+    builtinOperatorId = BuiltinOperatorId.PackOkResult,
+    cloneEvenIfFirst = true,
```
tjpalmer, line 1137: Too much cloning sometimes happens here, but it's still semantically correct, even if wasteful. I think without extra juggling, the only other option in my current handling is passing a reference, and that's semantically wrong at least sometimes.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -2262,97 +2263,6 @@ class RustTranslator(
 
     private fun translateGetterId(getter: TmpL.Getter) = translateDotName(getter.dotName)
 
-    /* do not commit
```
tjpalmer, line 2265: Yeah, this shouldn't really be needed anymore, even if some cleverness in here made smaller output code.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -3010,8 +2920,9 @@ class RustTranslator(
         val pos = statement.pos
         val context = functionContextStack.last()
         val returnType = context.returnType
-        val value = when (val value = statement.expression) {
-            null -> when {
+        val expr = statement.expression
+        val value = when {
+            expr.isVoidish() -> when {
```
tjpalmer, line 2925: Because `return void` now.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -421,35 +421,42 @@ class RustBackendTest {
...
                 |    } else {
                 |        t___0 = false;
                 |    }
                 |    if t___0 {
+                |        let t___1: i32;
                 |        if i__0.is_none() {
                 |            return Err(temper_core::Error::new());
                 |        } else {
-                |            t___1 = i__0.unwrap();
+                |            let result___0: temper_core::Result<i32> = Ok(i__0.unwrap());
```
tjpalmer, line 449: I think in some cases we're getting needless Result juggling, and in other cases we're wrapping things that already are Result. Some type info is being provided wrong here at least sometimes.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -421,35 +421,42 @@ class RustBackendTest {
...
+                |            if ! result___0.is_ok() {
+                |                return result___0;
+                |            }
+                |            t___1 = result___0.unwrap();
+                |        }
+                |        let return___0: temper_core::Result<i32> = Ok(temper_core::int_rem(5, t___1));
```
tjpalmer, line 455: For example, `int_rem` already returns Result and doesn't need Ok-wrapped, but the wrong type is being provided for the call expression (String instead of String | Bubble), so RustTranslator wraps it.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -2458,8 +2467,7 @@ class RustBackendTest {
             |}
             |pub fn hi(n__0: i32, things__0: impl temper_core::ToList<std::sync::Arc<String>>) -> i32 {
             |    let things__0 = things__0.to_list();
-            |    let mut t___0: i32 = temper_core::ListedTrait::len( & things__0);
-            |    return n__0.wrapping_add(t___0);
+            |    return n__0.wrapping_add(temper_core::ListedTrait::len( & things__0));
```
tjpalmer, line 2484: Lots of reduced temporaries are nice.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -943,6 +943,12 @@ fun List<TmpL.Statement>.splitConstructorBody(): Pair<List<TmpL.Statement>, List
     return initStatements to useStatements
 }
 
+/** True if either null or a void constant. */
+fun TmpL.Expression?.isVoidish(): Boolean = when (this) {
+    null -> true
+    else -> this.isVoidConstant
+}
```
tjpalmer, line 950: If we *never* produce plain `return;` anymore, we could just use `.isVoidConstant`, but I wasn't sure.

mikesamuel (approved):

PR #479, opened 2026-09-04
Support typed config objects

tjpalmer:
- Specifically support typed config for be-java
- Inject bindings during import stage from all registered backends
- Also inject std config.temper.md source from backends so it's not present when a backend isn't active and the config bindings are also missing
- Also evaluate `new` calls during partial interp for imu types
- Eval function defs in export stage so we get export values
- Still support old top-level be-java fields so unchanged libraries still work (and also still has `<groupId>dev.temperlang</groupId>` and `package temper.regex_parser;` in generated files):

```
temper-regex-parser> temper test -b java
Tests passed: 22 of 22
```


be-java/src/commonMain/kotlin/lang/temper/be/java/JavaBackend.kt
```diff
@@ -335,21 +336,6 @@ class JavaBackend private constructor(
...
-        val javaLibraryGroupConfigKey = Symbol("javaGroup")
-
-        /** Config files may export a name with this text to specify the Maven artifact id */
-        val javaLibraryArtifactConfigKey = Symbol("javaArtifact")
-
-        /** Config files may export a name with this text to specify the Java `package` name */
-        val javaPackageConfigKey = Symbol("javaPackage")
-
-        /** Config key to specify Maven dependencies */
-        val javaDependenciesKey = Symbol("javaDependencies")
```
tjpalmer, line 351: Moved to config description in frontend to be with the new config type for use in std.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaBackend.kt
```diff
@@ -411,6 +397,8 @@ class JavaBackend private constructor(
         val pomPath = filePath("pom.xml")
         val pomMime = MimeType("text", "xml")
 
+        override val configBindingsInjector: BindingsInjector = JavaConfigInjector
```
tjpalmer, line 403: We still want these configured on each backend so that plugged backends also can register them, even if we don't yet have a way for them to do custom config for std.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaLibraryConfig.kt
```diff
@@ -64,21 +70,48 @@ open class JavaLibraryConfigs(
...
+        // We currently generate within the context of the config module, so the origin isn't special.
+        (typeShape.name as? ExportedName)?.baseName?.nameText == JavaConfigKeys.CONFIG_CLASS_NAME || run {
+            console.error("Expected config class ${JavaConfigKeys.CONFIG_CLASS_NAME}, not ${typeShape.name}")
+            return@value null
+        }
+        // Good enough for now. Extract a pretty map.
+        (value.stateVector as InstancePropertyRecord).properties.map { instance ->
+            (instance.key as SourceName).baseName.nameText to instance.value
+        }.toMap()
+    }
```
tjpalmer, line 100: Even if the definitions are in the frontend, we use them from the java backend.

And while sloppy checks here for config type are better than nothing, somehow sneaking this properly into frontend type checking might be nice.

What *does* get checked in frontend now, though, is the the field types themselves for the JavaConfig instance, and that's likely to be helpful.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaLibraryConfig.kt
```diff
@@ -64,21 +70,48 @@ open class JavaLibraryConfigs(
...
-                ?: cfg(JavaBackend.javaPackageConfigKey)
+            cfg(JavaConfigKeys.GROUP, JavaConfigKeys.libraryGroupGlobal)
+                ?: cfgPackage()
                 // Dashes not allowed in group names per
                 // maven.apache.org/guides/mini/guide-naming-conventions.html
                 ?: libraryName.safeIdentifier()
-    private val libraryArtifact: String get() = cfg(JavaBackend.javaLibraryArtifactConfigKey) ?: libraryName
+
+    private val libraryArtifact: String
+        get() = cfg(JavaConfigKeys.ARTIFACT, JavaConfigKeys.libraryArtifactGlobal) ?: libraryName
```
tjpalmer, line 116: To support fallback for existing temper libs, check both the typed config and if missing then the global config var.

be-java/src/commonTest/kotlin/lang/temper/be/java/JavaBackendTest.kt
```diff
@@ -2073,7 +2073,10 @@ class JavaBackendTest {
                 |  config.temper.md: ```
                 |    # My test library
                 |
-                |        export let javaPackage = "com.example";
+                |        export let java = {
+                |          class: JavaConfig,
+                |          package: "com.example",
+                |        };
                 |    ```,
```
tjpalmer, line 2080: If we get target-typed prop bags, we could say this instead:

```ts
export let java: JavaConfig = {
  package: "com.example",
};
```

I'd rather not rely on unique keys, even though `package` is unique right now, because I suspect different backend configs could conflict easily in the general sense.

be-test-helpers/src/commonMain/kotlin/lang/temper/be/FunctionalTestRunner.kt
```diff
@@ -141,10 +141,12 @@ abstract class FunctionalTestRunner<BACKEND : Backend<BACKEND>>(
         val preparedModules = prepareModulesForFunctionalTest(
             test,
             projectLogSink = logSink,
-            customizeModule = { module, _ ->
-                module.addEnvironmentBindings(server.makeBindings())
-                for (factory in backendOrganization.factoriesById.values) {
-                    module.addEnvironmentBindings(factory.environmentBindings)
+            customizeModule = { module, isNew ->
+                if (isNew) {
```
tjpalmer, line 146: Added `isNew` checks around because that seems appropriate. I think our specific usage for env bindings was ok because it went into a Map, but we didn't need to be resetting at each stage. And for injectors, even though I've made that a Set, there's still risk of duplicates, depending on hash/equality definitions on them.

be/src/commonMain/kotlin/lang/temper/be/Backend.kt
```diff
@@ -812,6 +814,19 @@ abstract class Backend<SELF : Backend<SELF>>(
...
+
+        /** Add environment bindings to the module as appropriate. */
+        fun addEnvironmentBindings(module: Module) {
+            module.addEnvironmentBindings(environmentBindings)
+            if (module.isConfigModule) {
+                configBindingsInjector?.also { injector ->
+                    module.addBindingsInjector(injector)
+                }
+            }
+        }
```
tjpalmer, line 832: Centralize the env bindings and the config injector into a central place.

I'm still unsure how efficient it would be to try to generate lots of independent modules for explicit import in config vs just injecting in each. Might depend on the number of libraries being built at a go.

cli/src/test/kotlin/lang/temper/cli/DoTestTest.kt
```diff
@@ -135,7 +135,7 @@ class DoTestTest {
 
     @Test
     @Timeout(JAVA_TIMEOUT_SECONDS)
-    fun connectedsJava17All() {
+    fun connectedsJava17() {
```
tjpalmer, line 138: Slightly off name before. Maybe I did that from copy-paste?

cli/src/test/kotlin/lang/temper/cli/DoTestTest.kt
```diff
@@ -409,26 +409,22 @@ private fun doTestResult(
     }
 
     val realBackends = backendIds.filter { it != interpBackendId }
-    val harness = BuildHarness(
+    val libraries: Set<DashedIdentifier>? = libraryName?.let { setOf(it) }
+    val build = prepareBuild(
```
tjpalmer, line 413: Without going through prepareBuild, we weren't getting all the same setup. I don't remember if this was needed generally or just for std. And if for std, maybe that was an intermediate form and not needed now? But probably doesn't hurt to be more uniform in our setup.

cli/src/test/resources/testing/connecteds/config.temper.md
```diff
@@ -8,6 +8,9 @@ Test extra things about user-space connecteds here.
...
 
-    export let javaDependencies = [
-      "org.roaringbitmap:RoaringBitmap:1.6.20",
-    ];
+    export let java = {
+      class: JavaConfig,
+      dependencies: [
+        "org.roaringbitmap:RoaringBitmap:1.6.20",
+      ],
+    };
```
tjpalmer, line 16: And this still works, which is nice.

And again, frontend checks that things like `dependencies` have the right type on them. And presumably autocomplete on property names could work also, although I'm unsure I have that implemented for property bags so far.

frontend/src/commonMain/kotlin/lang/temper/frontend/importstage/ImportStage.kt
```diff
@@ -90,6 +90,9 @@ internal class ImportStage(
                 },
                 afterInterpretation = { (root), _ ->
                     flipDeclaredNames(root)
+                    for (injector in module.bindingsInjectors()) {
+                        injector.inject(module, root, logSink)
+                    }
```
tjpalmer, line 95: Here's where we actually inject in import stage. Providing some import resolver instead could be clever, but I didn't try working through that. And note that I'm taking advantage of my injected defs running through follow-on stages with the rest of the module.

And note that here nothing is specific to config modules. But code elsewhere currently only puts injectors in config modules so far.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/BuildConfigType.kt
```diff
@@ -0,0 +1,55 @@
...
+fun Planting.buildConfigType(
+    name: String,
+    properties: Map<String, Planting.() -> Any?>,
+) {
+    Call {
+        Rn(ParsedName(imuDecorator.name))
+        // Using the value directly here doesn't work.
+        // Call(ExportDecorator) { ... }
+        Call {
+            Rn(ParsedName(ExportDecorator.name))
```
tjpalmer, line 32: I didn't try figuring out why the value doesn't work above. Throughout here, I tried using values everywhere I could get them to work. Beyond that, the structure is based on what the frontend usually generates at import stage for manual definitions of a similar type directly in temper code.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/backend/JavaConfig.kt
```diff
@@ -0,0 +1,60 @@
+package lang.temper.frontend.staging.backend
```
tjpalmer: And yes, I have both frontend and backend in the package name here.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/backend/JavaConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+    const val ARTIFACT = "artifact"
+    val libraryArtifactGlobal = Symbol("javaArtifact")
+
+    /** Config files may export a name with this text to specify the Java `package` name */
+    const val PACKAGE = "package"
+    val packageGlobal = Symbol("javaPackage")
+
+    /** Config key to specify Maven dependencies */
+    const val DEPENDENCIES = "dependencies"
+    val dependenciesGlobal = Symbol("javaDependencies")
```
tjpalmer: Here we have both the moved config global names and the new property names.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/backend/JavaConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+                        Call(BuiltinFuns.angleFn) {
+                            V(Value(Types.list))
+                            V(Value(Types.string))
+                        }
+                    },
+                ),
+            )
+        }
+    }
+}
```
tjpalmer: And while I could try to define more of a schema instead of all the callbacks here, I find this still rather convenient to write and read while being flexible and needing very little support helpers.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/backend/README.md
```diff
@@ -0,0 +1,6 @@
+# Backend Config Helpers
+
+We configure std for bundled backends, so we need basic backend config
+definitions in the frontend.
+
+That's all that should be here.
```
tjpalmer: Felt it was worth putting explicit rules here for what goes in the "frontend backend" code at this time. And working out some way to inject actual config code into config.temper.md from backends for std could possibly remove the need for this code being here. And could also help for plugged in backends. But unless there's a quick fix for that, I'd rather move forward with this for now.

And just the gradle submodule dependencies likely help us avoid putting too much backend-specific code here. We can't reference any actual backend kotlin code, for example.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
```diff
@@ -672,7 +673,7 @@ private class GroupOfModulesToAdvanceTogether(
             val lastStageCompleted = m.stageCompleted
             val isConfigModule = m.isConfigModule
             if (lastStageCompleted == stageBeforeRun && readyToRun != null && !isConfigModule) {
-                readyToRun!!.add(m)
+                readyToRun.add(m)
```
tjpalmer, line 678: Recommended by ij.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/ModuleAdvancer.kt
```diff
@@ -1066,6 +1073,11 @@ private class ModuleAdvancerContinueConditionImpl : ContinueCondition {
     override fun toString(): String = "ModuleAdvancerContinueConditionImpl(${count[0]})"
 }
 
+/** Needed for including config for our bundled backends in std. */
+private val sharedStdConfigInjectors = setOf(
+    JavaConfigInjector,
+)
+
 private val sharedStdModules = lazy {
```
tjpalmer, line 1094: With sharedStdModules themselves already global here, I didn't feel too bad about making the injectors also global. Note that both are private to this file.

frontend/src/commonMain/kotlin/lang/temper/frontend/Module.kt
```diff
@@ -195,6 +195,19 @@ class Module(
...
+
+    /**
+     * Injectors called during import stage. The primary initial use is for
+     * injecting config classes into config modules.
+     *
+     * TODO Instead define virtual modules or ImportResolvers for import?
+     */
+    fun addBindingsInjector(injector: BindingsInjector) {
+        mutableBindingsInjectors.add(injector)
+    }
```
tjpalmer, line 209: Here's comments about me musing on import resolvers.

frontend/src/commonMain/resources/std/config.temper.md
```diff
@@ -37,17 +37,20 @@ We've chosen to use `temper` as a package prefix for official Temper packages,
...
 dependency of `std`, so its internals can use it. For now, usage is done via
 specialized connected methods.
 
-    export let javaDependencies = [
-      "org.junit.jupiter:junit-jupiter:5.9.2",
-    ];
+      dependencies: [
+        "org.junit.jupiter:junit-jupiter:5.9.2",
+      ],
+    };
```
tjpalmer: Converted std to new style here, which is what needed the schema code being in frontend.

interp/src/commonMain/kotlin/lang/temper/interp/Interpreter.kt
```diff
@@ -1869,6 +1869,7 @@ class Interpreter(
             goingOutOfStyle -> true
             (stageRange != null && stage in stageRange) -> true
             // Produce exports
+            stage == Stage.Export -> true
             stage == Stage.GenerateCode -> true
             else -> false
```
tjpalmer, line 1874: And we discussed this some on discord, but I can't get values into exports unless either:

- Export stage actually provides function values or
- We wait until generate code stage to make exports.

Note that the already existing comment above suggests someone was thinking about exports coming from generate code stage, even though that's not what happens elsewhere.

interp/src/commonMain/kotlin/lang/temper/interp/New.kt
```diff
@@ -150,3 +158,7 @@ object New : NamedBuiltinFun, SpecialFunction {
 
     override fun toString(): String = "builtin New"
 }
+
+private fun TypeShape.isImu(): Boolean {
+    return (stayLeaf?.incoming?.source as? DeclTree)?.parts?.let { imuSymbol in it.metadataSymbolMap } == true
+}
```
tjpalmer, line 164: Not much code needed to produce `new` values in partial interp for `@imu`. I don't know what all other functions run, but at least basic values and lists work as shown in this pr. And supporting proper `@partialImu` usage would be correct here, too, but I didn't look into that in this pr.

I didn't look to see if partial interp support for lists values check `@partialImu` rules correctly. Lists of strings were already working before any of my recent changes.

be/src/commonMain/kotlin/lang/temper/be/Backend.kt
mikesamuel: Do they need to have the authority to affect all modules, or can we just have their effect limited to config modules.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/BuildConfigType.kt
```diff
@@ -0,0 +1,55 @@
...
+import lang.temper.value.vWordSymbol
+import lang.temper.value.visibilitySymbol
+
+/** Builds an imu data class with nullable constructor properties only. */
+fun Planting.buildConfigType(
+    name: String,
+    properties: Map<String, Planting.() -> Any?>,
+) {
+    Call {
+        Rn(ParsedName(imuDecorator.name))
```
mikesamuel, line 28: Should this be a builtinname?  And ExportDecorator.name below?

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/backend/JavaConfig.kt
mikesamuel, line 1: It seems less than ideal to have backend-specific stuff outside of that directory.

Is there any way to avoid that?

tjpalmer: > It seems less than ideal to have backend-specific stuff outside of that directory.
> 
> Is there any way to avoid that?

As I've mentioned, I can't put java config info in be-java and also use it in std, because we don't always have be-java on the classpath.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/BuildConfigType.kt
tjpalmer: > Should this be a builtinname? And ExportDecorator.name below?

In at least some cases, I think BuiltinName was failing on me, but I can do a fresh pass to try things out.

be/src/commonMain/kotlin/lang/temper/be/Backend.kt
tjpalmer: > Do they need to have the authority to affect all modules, or can we just have their effect limited to config modules.

While I allow for injectors in all modules with this infrastructure, I only define them for config modules. We could wrap the ImportStage injection code in a check there also if you want, and I could at least comment that they're only applied on config modules. But they already are limited to that from what a backend can do.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/backend/JavaConfig.kt
tjpalmer: > As I've mentioned, I can't put java config info in be-java and also use it in std, because we don't always have be-java on the classpath.

And I already linked in the pr description my failed attempt at controlling config injectors from backends.

tjpalmer: @ShawSumma recommends possibly injecting std config content and not just type definitions. Well, full recommendations could be much more elaborate, but that's the simple scope version that maybe I can fit in this pr. So I'll look into that.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaBackend.kt
tjpalmer: > Moved to config description in frontend to be with the new config type for use in std.

Now in a different file in be-java.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaLibraryConfig.kt
```diff
@@ -98,9 +138,59 @@ class JavaLibraryConfig(
...
+    val libraryArtifactGlobal = Symbol("javaArtifact")
+
+    /** Config files may export a name with this text to specify the Java `package` name */
+    const val PACKAGE = "package"
+    val packageGlobal = Symbol("javaPackage")
+
+    /** Config key to specify Maven dependencies */
+    const val DEPENDENCIES = "dependencies"
+    val dependenciesGlobal = Symbol("javaDependencies")
+}
```
tjpalmer, line 174: Separated these keys out.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaLibraryConfig.kt
```diff
@@ -98,9 +138,59 @@ class JavaLibraryConfig(
...
+                        Call(BuiltinFuns.angleFn) {
+                            V(Value(Types.list))
+                            V(Value(Types.string))
+                        }
+                    },
+                ),
+            )
+        }
+    }
+}
```
tjpalmer, line 196: And here's the type definition for java config. Maybe could make this temper source like the bonus config section, but I already had this part working before that was done. That source injection is only for std config. This here is for config typing for any library, including std.

be-java/src/commonMain/resources/lang/temper/be/java/temper-std/config.temper.md
```diff
@@ -0,0 +1,20 @@
...
+      package: "temper.std",
+
+And as long as the `testing` module is part of `std`, we need junit as a core
+dependency of `std`, so its internals can use it. For now, usage is done via
+specialized connected methods.
+
+      dependencies: [
+        "org.junit.jupiter:junit-jupiter:5.9.2",
+      ],
+    };
```
tjpalmer, line 20: Backends now can add their own separate config chunk for std, so this also works for plugged in configs. Alternatively, they could use custom backend logic, but seeing config for std that matches config for other libraries can be nice.

be-test-helpers/src/commonMain/kotlin/lang/temper/be/GeneratedCodeHelper.kt
```diff
@@ -94,6 +95,7 @@ fun <BACKEND : Backend<BACKEND>> generateCode(
         lookupFactory = lookupFactory,
         onError = { error(it) },
     )
+    backendOrganization.addSharedStdConfigInjectors(supportedBackends)
```
tjpalmer: Here's the new clever way of ensuring shared std has what it needs, including from backends that might not be active on the first run. This primarily matters for unit test classes that run different backends in different test cases.

be/src/commonMain/kotlin/lang/temper/be/Backend.kt
```diff
@@ -997,12 +1016,29 @@ data class BackendOrganization(
...
+        for (id in allIds) {
+            // And use the factories we already have before looking up others.
+            val factory = factoriesById[id] ?: lookupFactory(id)
+            factory?.configBindingsInjector?.also { injector ->
+                val plugin = SharedStdConfigPlugin(injector, factory.loadStdConfigSource())
+                plugInSharedStdConfig(factory.backendId, plugin)
+            }
+        }
+    }
+}
```
tjpalmer, line 1041: And the actual helper method that gets called from anywhere that organizes backends (currently 3 places). Most importantly, this provides config for supported backends from the temper cli.

It's also called for funtests and backend tests but without any extra backends listed.

cli/src/test/kotlin/lang/temper/cli/RunTest.kt
```diff
@@ -115,19 +114,17 @@ class RunTest {
     fun runWithImportJsBackend() = runRunTest("runWithImport", JsBackend.Factory.backendId)
 
     @Test
-    fun runWithStdImportDefaultBackend() {
-        val defaultBackend = defaultBackend.value
+    fun runWithStdImportJsBackend() {
```
tjpalmer, line 117: I removed `defaultBackend`, so I changed these cases to js specifically, which was the net effect before anyway.

frontend/src/commonMain/kotlin/lang/temper/frontend/core/StandardLibrary.kt
```diff
@@ -3,4 +3,4 @@ package lang.temper.frontend.core
 import lang.temper.fs.FileSystem
 
 /** Returns a file system providing the standard library, which is available by import. */
-expect fun accessStd(): FileSystem?
+expect fun accessStd(configPluginSource: String): FileSystem?
```
tjpalmer, line 6: For passing down the concatenated contributed backends bonus config source chunks.

frontend/src/commonMain/resources/std/config.temper.md
```diff
@@ -31,24 +31,6 @@ libraries.
...
-    export let javaPackage = "temper.std";
-
-And as long as the `testing` module is part of `std`, we need junit as a core
-dependency of `std`, so its internals can use it. For now, usage is done via
-specialized connected methods.
-
-    export let javaDependencies = [
-      "org.junit.jupiter:junit-jupiter:5.9.2",
-    ];
-
```
tjpalmer, line 51: Now injected by be-java so it can use (also injected) JavaConfig.

frontend/src/jvmMain/kotlin/lang/temper/frontend/core/StandardLibrary.kt
```diff
@@ -10,11 +10,22 @@ import java.nio.file.FileSystems
...
     fun copyFrom(path: Path) = MemoryFileSystem().also {
-        copyRecursive(from = RealFileSystem(path), to = it)
+        copyRecursive(from = RealFileSystem(path), to = it) { filePath, bytes ->
+            when {
+                configPluginSource.isNotEmpty() && filePath.lastOrNull()?.fullName == "config.temper.md" -> {
+                    // Prefix an extra newline for convenience here. This is a rare cost.
+                    // We even could just load as bytes in the first place, but this is easier to look at.
+                    // This does presume our main std config.temper.md is utf-8, but we can manage that.
+                    val pluginBytes = "\n$configPluginSource".encodeToByteArray()
+                    bytes + pluginBytes
```
tjpalmer, line 24: Here's where we actually append the backend-contributed std config source.

fs/src/commonMain/kotlin/lang/temper/fs/FileSystem.kt
```diff
@@ -139,7 +143,7 @@ fun copyRecursive(
                     is RSuccess -> {
                         to.write(
                             path = toPath.resolve(fromKid.segments.last(), isDir = false),
-                            bytes = it.result.copyOf(),
+                            bytes = munge(fromKid, it.result.copyOf()),
```
tjpalmer, line 146: Support for contributing std chunks. This seemed easier than other options.

supported-backends/src/commonMain/kotlin/lang/temper/supportedBackends/SupportedBackends.kt
```diff
@@ -156,9 +156,10 @@ private fun backendsFromJson(
                     }
                 }
                 is RFailure -> {
-                    console.group("Error loading Temper plugin $jvmClassName", Log.Error) {
-                        console.error(r.failure)
-                    }
+                    // This typically is for tests where not all registered backends are on the classpath.
+                    // In those cases, this message isn't much interesting, but someone's actual environment could be
+                    // broken, and even in that case, the message should be enough info without a stack dump.
+                    console.error("Error loading Temper plugin $jvmClassName: ${r.failure}")
```
tjpalmer, line 162: Shrink error messages for when backends aren't available. This still usually is clear about the issue.

supported-backends/src/commonMain/kotlin/lang/temper/supportedBackends/SupportedBackends.kt
```diff
@@ -223,9 +224,6 @@ val testedBackends: List<BackendId> = buildList {
     sort()
 }
 
-val defaultBackend: Lazy<Backend.Factory<*>> =
-    supportedBackendsGlobal.factories.getValue(BackendId("js")).backend
```
tjpalmer, line 227: We were only using this in RunTest and not even in the cli, and it was breaking anyone trying to use anything else in this module if be-js wasn't on the classpath, so I removed it.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/BuildConfigType.kt
tjpalmer: > > Should this be a builtinname? And ExportDecorator.name below?

In latest checks, BuiltinName worked for imu but not export, so I made the change for imu.

be-test-helpers/src/commonMain/kotlin/lang/temper/be/GeneratedCodeHelper.kt
```diff
@@ -94,6 +94,8 @@ fun <BACKEND : Backend<BACKEND>> generateCode(
         lookupFactory = lookupFactory,
         onError = { error(it) },
     )
+    // We only get here from backend-specific test classes, so no need for other backends.
+    backendOrganization.addSharedStdConfigInjectors(listOf())
```
tjpalmer, line 98: For cli, we pass in supportedBackends, but here and for funtests we don't need that, and it causes error messages.

tooling/src/commonMain/kotlin/lang/temper/tooling/buildrun/Build.kt
```diff
@@ -356,6 +358,7 @@ fun doOneBuild(build: Build): BuildResult {
             }
         },
     )
+    backendOrganization.addSharedStdConfigInjectors(supportedBackends)
```
tjpalmer, line 361: For main build, inject all extra std config for all supported backends. Otherwise, a unit test case for one backend misses config for others in the same test class. This logic should also be fine for cli needs.

frontend/src/commonMain/kotlin/lang/temper/frontend/staging/BuildConfigType.kt
mikesamuel: Thanks for giving that a try.

mikesamuel (approved):

PR #483, opened 2026-09-10
Fix more rust things

tjpalmer:
- Simplify and fix err repacking
- Avoid deref in ok packing
- Fix constructor splitting

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
@@ -297,6 +297,7 @@ private class Float64Compare(
 internal open class FunctionCall(
     connectedNames: List<String>,
     val functionName: String,
+    val avoidDeref: Boolean = false,
     builtinOperatorId: BuiltinOperatorId? = null,
     cloneEvenIfFirst: Boolean = false,
     /** Non-null means the indicated param has a special-tailored fn borrow type. */
@@ -316,6 +317,7 @@ internal open class FunctionCall(
         baseName: String,
         functionName: String,
         builtinOperatorId: BuiltinOperatorId? = null,
+        avoidDeref: Boolean = false,
         cloneEvenIfFirst: Boolean = false,
         fnIndex: Int? = null,
         hasGeneric: Boolean = false,
@@ -324,6 +326,7 @@ internal open class FunctionCall(
     ) : this(
         connectedNames = listOf(baseName),
         functionName = functionName,
+        avoidDeref = avoidDeref,
         builtinOperatorId = builtinOperatorId,
         cloneEvenIfFirst = cloneEvenIfFirst,
         fnIndex = fnIndex,
@@ -350,7 +353,7 @@ internal open class FunctionCall(
                     // We don't do this in user code because we have less promises about how they intend to use it.
                     // TODO If we do add borrows to Temper, we could generalize better.
                     when {
-                        selfArg.type.described().isInterface() -> self.deref()
+                        !avoidDeref && selfArg.type.described().isInterface() -> self.deref()
                         else -> self
                     }
                 }.let { self ->
@@ -1133,6 +1136,7 @@ private object IsOkResult : MethodCall(
 private object PackOkResult : FunctionCall(
     baseName = "PackOkResult",
     functionName = "Ok",
+    avoidDeref = true,
     builtinOperatorId = BuiltinOperatorId.PackOkResult,
     cloneEvenIfFirst = true,
 )
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
@@ -169,10 +169,10 @@ private fun supportCodeByOperatorId(builtinOperatorId: BuiltinOperatorId?): Supp
         BuiltinOperatorId.SafeAdaptGeneratorFn -> adaptGeneratorFnSafe
 
         // Required since using results for failure recovery
-        BuiltinOperatorId.IsOkResult -> IsOkResult
-        BuiltinOperatorId.PackOkResult -> PackOkResult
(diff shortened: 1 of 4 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -169,10 +169,10 @@ private fun supportCodeByOperatorId(builtinOperatorId: BuiltinOperatorId?): Supp
...
 
         // Required since using results for failure recovery
-        BuiltinOperatorId.IsOkResult -> IsOkResult
-        BuiltinOperatorId.PackOkResult -> PackOkResult
-        BuiltinOperatorId.RepackErrResult -> repackErrResult
-        BuiltinOperatorId.UnpackOkResult -> UnpackOkResult
+        BuiltinOperatorId.IsOkResult -> isOkResult
+        BuiltinOperatorId.PackOkResult -> packOkResult
+        BuiltinOperatorId.RepackErrResult -> RepackErrResult
+        BuiltinOperatorId.UnpackOkResult -> unpackOkResult
```
tjpalmer, line 175: Included some changes from `object` to `val`, because that was appropriate.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -1125,31 +1128,45 @@ private val timesIntInt = MethodCall("TimesIntInt", "wrapping_mul", BuiltinOpera
...
     baseName = "IsOkResult",
     memberName = "is_ok",
     builtinOperatorId = BuiltinOperatorId.IsOkResult,
 )
 
-private object PackOkResult : FunctionCall(
+private val packOkResult = FunctionCall(
     baseName = "PackOkResult",
     functionName = "Ok",
+    avoidDeref = true,
```
tjpalmer, line 1140: We don't want to deref in this case. For many connected methods, I cheat and don't take the Arc, so I can be more efficient, but for Ok packing, we want the Arc still.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -1125,31 +1128,45 @@ private val timesIntInt = MethodCall("TimesIntInt", "wrapping_mul", BuiltinOpera
...
+        returnType: Type2,
+        translator: RustTranslator,
+    ): Rust.Tree {
+        return Rust.Call(
+            pos,
+            callee = "Err".toId(pos),
+            // `.expect_err` and `.unwrap_err` both require the success type
+            // implements the Debug trait so that they can produce a panic message,
+            // `.unwrap_err_unchecked` is unsafe, and `.into_err` is nightly only.
+            // So live on `.err().unwrap()` instead.
```
tjpalmer, line 1162: Moved and adjusted your earlier comment.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustSupportNetwork.kt
```diff
@@ -1125,31 +1128,45 @@ private val timesIntInt = MethodCall("TimesIntInt", "wrapping_mul", BuiltinOpera
...
+            // Meanwhile, we use these just for finished results, so avoid clone here.
+            args = listOf((arguments[0].expr as Rust.Expr).methodCall("err").methodCall("unwrap")),
```
tjpalmer, line 1164: Because `blah.err().unwrap()` doesn't need blah to be Debug like `.unwrap_err()` does. And this also dodges any reference and/or clone issues. Note that I had to dig for that alternative. I'd also tried unwrap_err first.

be-rust/src/commonMain/resources/lang/temper/be/rust/temper-core/src/lib.rs
```diff
@@ -565,17 +565,6 @@ pub fn int64_to_string(i: i64, radix: Option<i32>) -> Arc<String> {
...
-pub fn repack_err_result<T, U, E>(r: &std::result::Result<T, E>) -> std::result::Result<U, E>
-where E: Clone {
-    // TODO: once `!` is out of experimental, the return type could be Result<!, E>.
-    match r {
-        Ok(_)  => panic!(
-            "calls to this should only be inside generated code that includes an !is_ok check"
-        ),
-        Err(x) => Err(x.clone()),
-    }
-}
```
tjpalmer, line 577: No longer needed.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -507,16 +507,14 @@ class RustBackendTest {
...
-            |        return__0 = ();
             |        if x__0 {
             |            return Err(temper_core::Error::new());
             |        }
             |        x = x__0;
-            |        return Ok(return__0);
             |        let selfish = C(std::sync::Arc::new(CStruct {
             |                    x
             |        }));
+            |        return Ok(selfish);
```
tjpalmer, line 517: This actually goes back to the output on main. Updated code to handle this in tmpl helpers below.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -919,7 +920,14 @@ fun List<TmpL.Statement>.splitConstructorBody(): Pair<List<TmpL.Statement>, List
...
     val useStatements = mutableListOf<TmpL.Statement>()
     var reachedUse = false
-    for (statement in this) {
+    var voidReturnName: ResolvedName? = null
+    statements@ for (statement in this) {
+        if (statement is TmpL.LocalDeclaration && statement.descriptor.isVoidLike) {
+            // Track the name but prune out the actual declaraion.
+            voidReturnName = statement.name.name
+            continue@statements
+        }
```
tjpalmer, line 929: Because we're now getting a void var to return but we don't want it and it confuses the constructor splitting here, just ditch everything about it.

PR #484, opened 2026-09-10
Factor void var clean; don't return early in tests

tjpalmer:
- Fix testingAsserts funtest

### DIFF
--- a/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
+++ b/be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
@@ -10,6 +10,7 @@ import lang.temper.be.tmpl.TmpL
 import lang.temper.be.tmpl.TmpLOperator
 import lang.temper.be.tmpl.TypedArg
 import lang.temper.be.tmpl.aType
+import lang.temper.be.tmpl.cleanOutVoidVar
 import lang.temper.be.tmpl.hasSplitSupers
 import lang.temper.be.tmpl.isNullValue
 import lang.temper.be.tmpl.isStdLib
@@ -661,6 +662,7 @@ class RustTranslator(
                 params = listOf(),
                 returnType = "()".toId(pos).wrapResult(),
                 block = translateBlock(
+                    pos = test.body.pos,
                     prefix = listOf(
                         // Init both our own crate and temper_std (in case our crate doesn't) before any tests.
                         // Init is idempotent, so multi-init is ok.
@@ -669,8 +671,10 @@ class RustTranslator(
                         Rust.ExprStatement(pos, Rust.Call(pos, stdInit, listOf("None".toId(pos)))),
                         Rust.LetStatement(pos, pattern = testObjectId, type = null, value = newTest),
                     ),
-                    block = test.body,
+                    statements = test.body.statements.cleanOutVoidVar(),
                     result = result,
+                    // Skip any final return so we can convert soft fail to hard right here instead.
+                    skipLastReturn = true,
                 ),
             ).toItem(
                 attrs = listOf(Rust.AttrOuter(pos, "test".toId(pos))),
@@ -1669,14 +1673,25 @@ class RustTranslator(
         skipLastReturn: Boolean = result != null,
         statementProcessor: StatementProcessor? = null,
     ): Rust.Block {
+        return translateBlock(block.pos, block.statements, prefix, result, skipLastReturn, statementProcessor)
+    }
+
+    private fun translateBlock(
+        pos: Position,
+        statements: List<TmpL.Statement>,
+        prefix: List<Rust.Statement> = listOf(),
+        result: Rust.Expr? = null,
+        skipLastReturn: Boolean = result != null,
+        statementProcessor: StatementProcessor? = null,
+    ): Rust.Block {
         return Rust.Block(
-            block.pos,
+            pos,
             statements = buildList {
                 addAll(prefix)
                 when (statementProcessor) {
-                    null -> processStatements(block.statements, results = this, skipLastReturn = skipLastReturn)
+                    null -> processStatements(statements, results = this, skipLastReturn = skipLastReturn)
                     else -> statementProcessor.processStatements(
-                        block.statements,
+                        statements,
                         results = this,
                         skipLastReturn = skipLastReturn,
                     )
@@ -1993,7 +2008,7 @@ class RustTranslator(
                 results: MutableList<Rust.Statement>,
                 skipLastReturn: Boolean, // ignorable here
             ) {
-                val (initStatements, useStatements) = statements.splitConstructorBody()
+                val (initStatements, useStatements) = statements.cleanOutVoidVar().splitConstructorBody()
                 val context = functionContextStack.last()
                 val inits = buildMap {
                     for (initStatement in initStatements) {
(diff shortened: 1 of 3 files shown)


be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -669,8 +671,10 @@ class RustTranslator(
                         Rust.ExprStatement(pos, Rust.Call(pos, stdInit, listOf("None".toId(pos)))),
                         Rust.LetStatement(pos, pattern = testObjectId, type = null, value = newTest),
                     ),
-                    block = test.body,
+                    statements = test.body.statements.cleanOutVoidVar(),
```
tjpalmer, line 674: I don't think this is *needed* in the end, but I got it in place before double-checking on that, and it does make things cleaner.

Once we have a general rule in tmpl translation for this, we can remove this patch-up function.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1668,15 +1672,26 @@ class RustTranslator(
         result: Rust.Expr? = null,
         skipLastReturn: Boolean = result != null,
         statementProcessor: StatementProcessor? = null,
+    ): Rust.Block {
+        return translateBlock(block.pos, block.statements, prefix, result, skipLastReturn, statementProcessor)
+    }
+
+    private fun translateBlock(
+        pos: Position,
+        statements: List<TmpL.Statement>,
```
tjpalmer, line 1681: Just provide an option for sending statements directly.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
```diff
@@ -1993,7 +2008,7 @@ class RustTranslator(
                 results: MutableList<Rust.Statement>,
                 skipLastReturn: Boolean, // ignorable here
             ) {
-                val (initStatements, useStatements) = statements.splitConstructorBody()
+                val (initStatements, useStatements) = statements.cleanOutVoidVar().splitConstructorBody()
```
tjpalmer, line 2011: Factored out now and maybe not needed at some future point.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -652,7 +650,6 @@ class RustBackendTest {
             |            std::sync::Arc::new(move | | closure_group.fn__1())
             |        };
             |        test___0.assert(Some(actual___0) == Some(3), fn__1.clone());
-            |        return Ok(return__0);
             |        test___0.soft_fail_to_hard()
```
tjpalmer, line 653: The early returns were the big problem, but the other cleanup above is nice, too.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -912,6 +914,40 @@ fun TmpL.BlockStatement?.isPureVirtual(isFnPureVirtual: (TmpL.Callable) -> Boole
     } ?: true
 }
 
+/**
+ * Returns a list with a single Void-typed var removed, which is expected to be
+ * a return var that lingers from the frontend.
+ *
+ * If no such var is found, returns original [this] instance.
+ */
+fun List<TmpL.Statement>.cleanOutVoidVar(): List<TmpL.Statement> {
```
tjpalmer, line 923: This factors out what was mixed into the split constructor loop before.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -912,6 +914,40 @@ fun TmpL.BlockStatement?.isPureVirtual(isFnPureVirtual: (TmpL.Callable) -> Boole
...
+                else -> super.rewriteExpression(x)
+            }
+        }
+    }
+    // For efficiency, dig to immediate kids only, based on expected usage patterns.
+    return mapNotNull { statement ->
+        when (statement) {
+            is TmpL.LocalDeclaration if statement.name.name == voidName -> null
+            is TmpL.Assignment if statement.left.name == voidName -> null
+            is TmpL.ReturnStatement -> returnRewriter.rewriteReturnStatement(statement)
```
tjpalmer, line 945: But it now retains `return Ok(void);` instead of just `return;` because I accommodate for that later now.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -912,6 +914,40 @@ fun TmpL.BlockStatement?.isPureVirtual(isFnPureVirtual: (TmpL.Callable) -> Boole
...
+ *
+ * If no such var is found, returns original [this] instance.
+ */
+fun List<TmpL.Statement>.cleanOutVoidVar(): List<TmpL.Statement> {
+    val voidName = mapFirst { statement ->
+        when (statement) {
+            is TmpL.LocalDeclaration if statement.descriptor.isVoidLike -> statement.name.name
+            else -> null
+        }
+    } ?: return this
```
tjpalmer, line 929: I get out early here for the common case. And also only ever look at top-level kids here like below.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -938,33 +967,25 @@ fun List<TmpL.Statement>.splitConstructorBody(): Pair<List<TmpL.Statement>, List
...
+        ) || statement is TmpL.ReturnStatement && statement.expression.isVoidish()
         // Keep other statements, adjusted if needed, placed in the correct group.
         when {
             !reachedUse -> initStatements
             else -> useStatements
-        }.add(adjustedStatement.deepCopy())
+        }.add(statement.deepCopy())
     }
     return initStatements to useStatements
 }
```
tjpalmer, line 978: Even if the file now has more lines of code at this point, simplifying the above loop is nice.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -938,33 +967,25 @@ fun List<TmpL.Statement>.splitConstructorBody(): Pair<List<TmpL.Statement>, List
...
+    is TmpL.CallExpression -> when (val callee = fn) {
+        is TmpL.SupportCodeWrapper ->
+            callee.supportCode.builtinOperatorId == BuiltinOperatorId.PackOkResult &&
+                parameters.size == 1 &&
+                (parameters.first() as? TmpL.Expression)?.isVoidConstant == true
+        else -> false
+    }
     null -> true
     else -> this.isVoidConstant
 }
```
tjpalmer, line 991: And the proper tmpl thing might be to say `Ok(void)` later, so the above checks for that case.

be-rust/src/commonMain/kotlin/lang/temper/be/rust/RustTranslator.kt
mikesamuel: Ok.  Not obvious from the diff view alone but this is all inside processTest.

mikesamuel (approved):

PR #486, opened 2026-09-11
Use `this` to avoid it disappearing

tjpalmer:
- See self review.

### DIFF
--- a/tooling/src/jvmTest/kotlin/lang/temper/tooling/FindDefTest.kt
+++ b/tooling/src/jvmTest/kotlin/lang/temper/tooling/FindDefTest.kt
@@ -57,7 +57,7 @@ internal open class FindDefTest {
 
     @Test
     fun failToFindThisWithoutClassName() {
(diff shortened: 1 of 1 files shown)


tooling/src/jvmTest/kotlin/lang/temper/tooling/FindDefTest.kt
```diff
@@ -57,7 +57,7 @@ internal open class FindDefTest {
 
     @Test
     fun failToFindThisWithoutClassName() {
-        val goodContext = FileModuleDataTestContext("""class /*1@+0*/Hi { let hi(): Void { this/*0*/; } }""")
+        val goodContext = FileModuleDataTestContext("""class /*1@+0*/Hi { let hi(): Void { this/*0*/.hi(); } }""")
```
tjpalmer, line 60: The `this` reference disappears by generate code stage now, but it didn't used to. It would be good to find the reference even in a useless spot, and I could change processing to work better from earlier stages, but actually using `this` is a workaround that helps test the core capability for now.

mikesamuel (approved):

PR #488, opened 2026-09-14
Stop crashing on unsupported anonymous classes

### LINKED ISSUE #485: ImuChecker crash on anonymous class decl
```ts
$ class {}
interactive#0: Anon__0
$ let a = class {};
1: let a = class {};
       ⇧
[interactive#1:1+4-5]@G: Missing type info for a__4
Exception in thread "main" java.lang.NullPointerException
        at lang.temper.frontend.typestage.ImuChecker.check(ImuChecker.kt:68)
        at lang.temper.frontend.generate.GenerateCodeStage.doAfterInterpretation(GenerateCodeStage.kt:143)
```

tjpalmer:
- Fix #485
- Don't try to make the syntax work, just stop crashing the whole frontend on it
- Examples in the repl after these changes, where lonely `class {}` behavior (entry 0 below) isn't changed this pr:

```ts
$ class {}
interactive#0: Anon__0
$ let a = class {};
1: let a = class {};
       ⇧
[interactive#1:1+4-5]@G: Missing type info for a__4
1: let a = class {};
   ┗━━━━━━━━━━━━━━┛
[interactive#1:1+0-16]@R: Declaration is malformed
interactive#1: fail
```


frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
```diff
@@ -65,7 +65,7 @@ class ImuChecker(
     fun check(tree: Tree) {
         for (child in tree.children) {
             if (child is DeclTree) {
-                child.parts!!.metadataSymbolMap[typeDeclSymbol]?.target
+                child.parts?.metadataSymbolMap[typeDeclSymbol]?.target
```
tjpalmer, line 68: This is the real change, and notice that there were already `?` after it. Maybe I was just being overzealous with the `!!`.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/anonymous-class/expect/generateCode.temper
```diff
@@ -0,0 +1,21 @@
...
+});
+@typeDecl(Anon__2) @stay @reach(\none) let t#3;
+t#3 = type (Anon__2);
+let[inconceivable];
+@fn @visibility(\public) @stay @fromType(Anon__4) let constructor__14;
+constructor__14 = (@stay fn constructor(@impliedThis(Anon__4) this__10: Anon__4) /* return__11 */: Void {
+    return__11 = void
+});
+@typeDecl(Anon__4) @stay @reach(\none) let t#5;
+t#5 = type (Anon__4);
```
tjpalmer, line 19: I think I managed to get 3 different Anon classes here but none assigned to the names given. We don't have many other examples of anonymous classes, so I'm unsure of our intentions with them.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/anonymous-class/expect/generateCode.temper
```diff
@@ -0,0 +1,21 @@
...
+let[inconceivable];
+error (list("@", "`(Leaf`", "export", "`Leaf)`", "`(Eq`", "`(Leaf`", "let", "decoratedInside", "`Leaf)`", "=", "`Eq)`", "`(At`", "@", "`(Leaf`", "imu", "`Leaf)`", "`(Curly`", "`(Leaf`", "class", "`Leaf)`", "{", "}", "`Curly)`", "`At)`"))
```
tjpalmer, line 21: Here's the only place `imu` shows up in the rendered pseudocode, and I think it comes from the `@imu class` case in test.temper, and it's fine if we don't want to support that. Mostly, I was just curious what works and what doesn't.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/anonymous-class/work/test/test.temper
```diff
@@ -0,0 +1,13 @@
+// Anonymous alone was already getting through the front end
+// before starting this test.
+class {}
```
tjpalmer, line 3: So this doesn't seem to give errors, but the cases below do.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/anonymous-class/work/test/test.temper
```diff
@@ -0,0 +1,13 @@
...
+
+// Initial goal here is to keep this from crashing the front end.
+export let plain = class {};
+
+// But check out some variations to see what happens.
+@imu
+export let decorated = class {};
+
+// Including this invalid syntax, just for thinking about.
+export let decoratedInside = @imu class {};
```
tjpalmer, line 13: Again, this last one is where I think the garbage trees are coming from. And that might be fine.

The key point is that after this PR, we don't hard crash the front end for any of these cases, anymore.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/anonymous-class/work/test/test.temper
mikesamuel: Hmm.  I guess this is equivalent to class t#0 {}

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
mikesamuel: If this is happening when it should be there, you can always replaceWithErrorNode(MessageTemplate.MalformedDeclaration) if it doesn't have what it should.

mikesamuel (approved):

frontend/src/commonMain/kotlin/lang/temper/frontend/typestage/ImuChecker.kt
tjpalmer: > If this is happening when it should be there, you can always replaceWithErrorNode(MessageTemplate.MalformedDeclaration) if it doesn't have what it should.

Thanks for the tip! I'll try to keep that in mind for appropriate use cases. In this case, I figure nobody else was dying on it, so this could shouldn't either.

Issue #489, opened 2026-09-14
Multiline string failure


tjpalmer:
This fails:

```ts
interactive#29: void
$ console.log("""
      ~a ${}
      "${1}
      )
  ┏━━━━━━━━━━━━┓
1:┃console.log("""
2:┃    ~a ${}
3:┃    "${1}
4:┃    )
  ┗━━━┛
[interactive#30:1+12 - 4+4]@P: Expected a Expression here
Interpretation ended due to runtime panic
interactive#30: fail
```

But these succeed:

```ts
$ console.log("""
      ~a
      "${1}
      )
a1

interactive#27: void
$ console.log("""
      "${1}
      )
1

interactive#28: void
$ console.log("""
      "a ${}
      "${1}
      )
a
1

interactive#29: void
...
$ console.log("""
      ~a ${}
      "1
      )
a 1

interactive#33: void
$ console.log("""
      ~a ${}
      ~1
      )
a 1
interactive#34: void
```

PR #490, opened 2026-09-15
Support default null args in connected functions

tjpalmer:
- Work toward #456
- Add an example of a function with a default null arg to the connecteds funtest
- Update all backends (well, still not interp) to handle it
- Also add some to each backend test


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppTranslator.kt
```diff
@@ -2978,15 +2978,12 @@ class CppTranslator(
...
-                            else -> cppName
-                        }.also { add(it) }
-                    }
-                },
+                args = defaulting.buildConnectedArgs(
+                    fn = fn,
+                    backendParams = paramNames,
+                    tmplToArg = { pos, name -> cpp.pos(pos) { cpp.name(name) } },
+                    backendToArg = { it },
+                ),
```
tjpalmer, line 2986: I would need to make the small change to all the backends that differed only in subtle ways of building their arg lists for the connected call. So I factored it out to an `inline` function. Each has enough customization things that I wasn't sure it would be a good idea, but I think the factored form does look at least a little simpler than before. Usually saves 3+ lines in each case also.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -715,6 +715,9 @@ class CppBackendTest {
                     |export let inc(i: Int): Int {
                     |    sum(i, 1)
                     |}
+                    |
+                    |@connected
+                    |export let length(s: String? = null): Int;
```
tjpalmer, line 720: And I didn't bother to write the actual connected code for backend tests, but I do have implementations for each in the funtest.

I haven't tested various combinatorics of null default and non-null default and non-optional parameters. I doubt there would be problems, but I haven't tested yet in this PR. Since I did factor out logic, hopefully any needed adjustments in the future would automatically apply to all backends.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -765,27 +768,33 @@ class CppBackendTest {
...
-                |                    int32_t sum(int32_t i_6, int32_t j_7) {
-                |                      return sum(i_6, j_7, nullptr);
+                |                    int32_t length(temper::core::NullableParam<std::string> s) {
+                |                      return _connected::length(s);
                 |                    }
-                |                    int32_t inc(int32_t i_10) {
-                |                      return sum(i_10, 1);
+                |                    int32_t length() {
+                |                      return length(nullptr);
                 |                    }
```
tjpalmer, line 798: Here's the new .cpp code. Note that we can just pass the nullable param on through, unlike for other optional params.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -809,6 +818,8 @@ class CppBackendTest {
                 |                    int32_t sum(int32_t, int32_t, temper::core::NullableParam<int32_t>);
                 |                    int32_t sum(int32_t, int32_t);
                 |                    int32_t inc(int32_t);
+                |                    int32_t length(temper::core::NullableParam<std::string>);
+                |                    int32_t length();
```
tjpalmer, line 822: And the .hpp code that was generated.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpTranslator.kt
```diff
@@ -879,14 +879,12 @@ internal class CSharpTranslator(
...
-                            else -> csharp.name.deepCopy()
-                        }.also { add(it) }
-                    }
-                },
+                args = defaulting.buildConnectedArgs(
+                    fn = decl,
+                    backendParams = parameters,
+                    tmplToArg = { pos, name -> translateName(pos, name) },
+                    backendToArg = { it.name.deepCopy() },
+                ),
```
tjpalmer, line 887: Again, I think the factored form is a bit simpler.

be-lua/src/commonMain/kotlin/lang/temper/be/lua/LuaTranslator.kt
```diff
@@ -1866,20 +1866,17 @@ private class ModuleParts(
...
-                    }
-                },
+                defaulting.buildConnectedArgs(
+                    fn = stmt,
+                    backendParams = params.params,
+                    tmplToArg = { pos, name -> Lua.Name(pos, luaNames.name(name)) },
+                    backendToArg = { arg ->
+                        when (arg) {
+                            is Lua.Param -> arg.name.deepCopy()
+                            is Lua.RestExpr -> arg.deepCopy() // TODO Good rest handling or frontend errors.
```
tjpalmer, line 1876: Note that I tried to retain old comments. And I did put errors on connected functions with rest args in #490, but I was only reviewing that just now. So we could possibly take off this todo, or it might still be a reminder that rest args support could be nice.

be-lua/src/commonTest/kotlin/lang/temper/be/lua/LuaBackendTest.kt
```diff
@@ -520,10 +523,17 @@ class LuaBackendTest {
             |                  inc = function(i__2)
             |                    return sum(i__2, 1);
             |                  end;
+            |                  length = function(s__0)
+            |                    if (s__0 == nil) then
+            |                      s__0 = temper.null;
+            |                    end
+            |                    return _connected.length(s__0);
+            |                  end;
```
tjpalmer, line 531: I don't like using `temper.null`, but Shaw points out that using `nil` for our null has its own downsides. Actual connected code needs to be prepared for `temper.null`, though. And the transform here is nothing new. It was already in LuaTranslator, but I got reminded of it.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -739,22 +745,29 @@ class PyBackendTest {
...
-            |                  return _connected.sum(i_2, j_3, bonus_4)
-            |              def inc(i_6: 'int1', /) -> 'int1':
-            |                  return sum(i_6, 1)
+            |                  return _connected.sum(i_3, j_4, bonus_5)
+            |              def inc(i_7: 'int1', /) -> 'int1':
+            |                  return sum(i_7, 1)
+            |              def length(s_9: 'Union2[str3, None]' = None, /) -> 'int1':
+            |## And here we have a wasted assignment. TODO Clean this out?
+            |                  _s_9: 'Union2[str3, None]' = s_9
+            |                  return _connected.length(s_9)
```
tjpalmer, line 770: Note the comment about my concern on wasted `_s_9`. This seems to be automated (and maybe I wrote it in the past???) for optional params, and I expect non-connected functions would use the new value. But we presumably don't ever need it. Maybe a future PR for just be-py could clean that out generally for all functions with optional nullable params.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -252,6 +255,10 @@ class RustBackendTest {
             |              pub fn inc(i__1: i32) -> i32 {
             |                  return sum(i__1, 1, None);
             |              }
+            |              pub fn length(s__0: Option<impl temper_core::ToArcString>) -> i32 {
+            |## It might be nice to stringify all semi-string before calling connected functions.
+            |                  _connected::length(s__0)
+            |              }
```
tjpalmer, line 261: Another awkwardness, sort of like the `temper.null` awkwardness of be-lua. I like handling defaults before passing along. And I could presumably call `.to_arc_string()` here before passing along also. But I didn't jump hoops for that, and if we make such a change in the future, it would be breaking.

Summary is that I see connected code as utilitarian, whereas the Temper built functions calling into it provide the nice API for general use.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -230,7 +231,24 @@ fun <T> TmpL.CallExpression.mapParameters(
...
     val defaultStatements: List<TmpL.Statement>,
     val parameterMapping: Map<ResolvedName, ResolvedName>,
-)
+) {
+    inline fun <BACKEND_PARAM, ARG> buildConnectedArgs(
+        fn: TmpL.FunctionDeclaration,
+        backendParams: List<BACKEND_PARAM>,
+        tmplToArg: (Position, ResolvedName) -> ARG,
+        backendToArg: (BACKEND_PARAM) -> ARG?,
+    ): List<ARG> {
```
tjpalmer, line 240: The helper I started using in all backends for their args now.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -230,7 +231,24 @@ fun <T> TmpL.CallExpression.mapParameters(
...
+        return buildList {
+            for ((tmpl, backend) in fn.parameters.parameters.zip(backendParams)) {
+                val arg = when {
+                    tmpl.optional -> parameterMapping[tmpl.name.name]?.let { tmplToArg(tmpl.pos, it) }
+                    else -> null
+                } ?: backendToArg(backend)
+                arg?.also { add(it) }
```
tjpalmer, line 247: The subtle new thing in this PR is using `parameterMapping[tmpl.name.name]?` instead of `parameterMapping.getValue(tmpl.name.name)` everywhere, because null default params don't end up in the map, and instead fall through along with the `else` branch to `backendToArg(backend)`.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -253,11 +271,16 @@ fun TmpL.FunctionDeclaration.parameterDefaultStatementsInfo(): DefaultStatements
     val defaultStatements = mutableListOf<TmpL.Statement>()
     val parameterMapping = buildMap parameterMapping@{ // TODO If nullable with default null, what happens here?
         parameters@ for (parameter in parameters.parameters) {
+            parameter.optional || continue@parameters
+            // Defaulting to null doesn't actually assign anything, so skip nulls.
+            val value = parameter.metadata.find { it.key.symbol == optionalSymbol }?.value as? TmpL.ValueData
+            value?.value == TNull.value && continue@parameters
```
tjpalmer, line 277: I was afraid we wouldn't have the default value metadata, but happily we do.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -253,11 +271,16 @@ fun TmpL.FunctionDeclaration.parameterDefaultStatementsInfo(): DefaultStatements
...
             // Start out with original names, but replace them later.
-            if (parameter.optional) {
-                this[parameter.name.name] = parameter.name.name
-            }
+            this[parameter.name.name] = parameter.name.name
         }
+        // If all the optionals were null defaulting, then get out easy anyway.
+        isEmpty() && return@parameterMapping
```
tjpalmer, line 282: There's an earlier exit for no optionals all, but I decided to leave that ignoring whether the default value is null so it can stay a faster check. I didn't benchmark at all, though.

functional-test-suite/src/commonMain/resources/functions/connected/__connected__.py
```diff
@@ -6,3 +6,6 @@ def sum(i: int, j: int, bonus: int) -> int:
 
     def prod(hidden: "_Hidden", j: int) -> int:
         return _connected.Support().prod(hidden.i, j)
+
+    def length(string: str | None) -> int:
+        return -1 if string is None else len(string)
```
tjpalmer, line 11: Here we get to actual connected code for the backends.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.cpp
```diff
@@ -13,5 +13,9 @@ std::int32_t prod(std::shared_ptr<Hidden> const& hidden, std::int32_t j) {
     return support.prod(hidden->get_i(), j);
 }
 
+int32_t length(temper::core::NullableParam<std::string> s) {
+    return s.has_value ? s.value.size() : -1;
+}
```
tjpalmer, line 18: And I didn't worry about consistent length definitions here outside of ascii space. I guess it serves as a warning that connected code introduces risks of inconsistent implementations across backends.

functional-test-suite/src/commonMain/resources/functions/connected/_connected.lua
```diff
@@ -18,4 +18,14 @@ do
     local support = s.Support.new()
     return support:prod(hidden.i, j)
   end
+
+  ---@param s string|nil Or actually temper.null instead of nil ...
+  ---@return integer
+  function _connected.length(s)
+    if s == temper.null then
```
tjpalmer, line 25: So my `string|nil` spec above is actually a lie, which is why I included the trailing comment. I didn't dig up if there's an appropriate way to tag a union with our null value. But my lua vscode plugin avoided whining at me, at least. Maybe it just got confused.

functional-test-suite/src/commonMain/resources/functions/connected/connected.temper.md
```diff
@@ -28,15 +28,28 @@ connected code needs to have access to.
 
     /* unexported */ class Hidden(public i: Int) {}
 
-A simple test will do. And this can't be inlined by Temper, since Temper the
-Temper implementation can only panic.
+Also try a function with a nullable default parameter that defaults to null. Our
+implementations of `length` for this test are sloppy, so only pass in ASCII.
```
tjpalmer, line 32: There's my ascii warning.

functional-test-suite/src/commonMain/resources/functions/connected/connected.temper.md
```diff
@@ -28,15 +28,28 @@ connected code needs to have access to.
...
+implementation can only panic.
+
+TODO How to test interpreter fallback? Different `log` blocks for interp?
+
+    console.log("sum(1, 2): ${sum(1, 2)}");
+    console.log("prod(new Hidden(1), 2): ${prod(new Hidden(1), 2)}");
+    console.log("""
+      ~length "" vs null vs ():
+      ~ ${length("")} vs ${length(null)} vs ${length()}
+    );
```
tjpalmer, line 47: I discovered #489 while writing this log statement.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
mikesamuel: The metadata about whether a param is optional or required?

mikesamuel (approved):

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
tjpalmer: > The metadata about whether a param is optional or required?

It actually has the default value there, or it least it works fine for default value `null`.

PR #491, opened 2026-09-15
frontend: support self-referential type definitions

mikesamuel:
In preparation for user-defined equality and comparison operations, this commit fixes some bugs around self-referential type definitions.

To get rid of the generic equality function, *Mapped*'s subtypes need a way to compare keys.

This commit redefines *MapKey*:

```diff
-interface MapKey extends Equatable {
+interface MapKey<SELF extends MapKey<SELF>> extends Equatable<SELF> {
```

And it similarly redefines *Equatable* to have a type parameter describing the type of arguments it can be equated to.

Now, `String extends MapKey<String>` meaning a string is a kind of map key that can be tested for equality with other strings.

This change caused some problems: *StaticType* has methods that recursively check types and *TypeDefinition* and this change introduced a *TypeFormal* type definition with a nested reference, via *TypeFormal.upperBounds* to a *NominalType* whose type definition is that same *TypeFormal*.

```mermaid
stateDiagram-v2
    MapKey --> SELF : member formal

    SELF --> MapKey#lt;SELF#gt; : upper bound

    MapKey#lt;SELF#gt; --> SELF[Nominal] : actual
    MapKey#lt;SELF#gt; --> MapKey : definition
    SELF[Nominal] --> SELF : definition
```

This commit reworks several methods to use a new *ButNotRecursively* to prevent unnecessary recursion:

- `.equals` assumes two things are equal and looks for counter-evidence
- `.hashCode` pegs the hashCode of a recursive inclusion at 0
- `.renderTo` emits "..." instead of a recursive rendering
- `.destructure` similarly to `renderTo`

This commit also adjusts some TyperTest cases.  Previously, `MapKey` was the least upper bound of `Int32` and `String` for example. Now they extend different types: `String extends MapKey<String>` and `Int32 extends MapKey<Int32>` so their least upper bound is `AnyValue`.

This commit also tweaks *InferBounds*'s checks on whether a type actual can bind to a type formal.
If we have two structurally similar type formals like the below, then one should be able to bind to a reference to the other:

- `<T extends MapKey<T>>`
- `<U extends MapKey<U>>`

Now *TypeContext* supports that by mapping the upper bounds of the target formal using the binding map before soing subtype checks.


be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -935,8 +935,8 @@ class RustBackendTest {
             |            let things__0: temper_core::List<temper_core::AnyValue> = std::sync::Arc::new(vec![1, 2]).unwrap();
             |            let more__0: temper_core::List<i32> = std::sync::Arc::new(vec![1, 2]);
             |            let still__0: temper_core::List<temper_core::AnyValue> = std::sync::Arc::new(vec![1, 2]);
-            |            let yet__0: temper_core::List<temper_core::MapKey> = std::sync::Arc::new(vec![temper_core::MapKey::new(1), temper_core::MapKey::new(std::sync::Arc::new("two".to_string()))]);
-            |            let yetAgain__0: temper_core::List<temper_core::MapKey> = yet__0.clone();
+            |            let yet__0: temper_core::List<std::sync::Arc<String>> = std::sync::Arc::new(vec![std::sync::Arc::new("one".to_string()), std::sync::Arc::new("two".to_string())]);
+            |            let yetAgain__0: temper_core::List<temper_core::MapKey<std::sync::Arc<String>>> = yet__0.clone();
```
mikesamuel, line 939: I don't know if this is right but it doesn't look terribly wrong.

frontend/src/commonMain/kotlin/lang/temper/frontend/disambiguate/TypeDisambiguateMacro.kt
```diff
@@ -926,9 +927,19 @@ internal fun typeDisambiguateMacro(
...
+                    predefinedFormalDecls?.first {
+                        it.word == formalSymbol
+                    }
+                } catch (e: NoSuchElementException) {
+                    // If you're here, you're editing Core.temper and need
+                    // to make corresponding changes to WellKnownTypes.kt,
+                    // or vice versa.
+                    throw NoSuchElementException(
+                        "${doc.context.formatPosition(formalPos)}: No formal $formalSymbol in $typeShape",
+                    ).initCause(e)
```
mikesamuel, line 942: WellKnownTypes defines type formals, and this augments them with information from Core.  If there's a mismatch it's nice to get a nice error message.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/TypeDeclChecker.kt
```diff
@@ -55,25 +55,55 @@ internal class TypeDeclChecker(val module: Module, val logSink: LogSink) {
...
-                strictSuperTypeShape.methods.mapNotNull { m ->
-                    if (isProcessingCore && connectedSymbol in m.metadata) {
-                        null // Some Core methods have no body because they must connect.
-                    } else if (isStd && m.visibility == Visibility.Private && connectedSymbol in m.metadata) {
-                        null // std has some required connections too which are sneakily hidden away
-                    } else if (m.methodKind != MethodKind.Constructor && m.isPureVirtual) {
-                        MethodDescriptor(m.symbol, m.methodKind, m)
-                    } else {
-                        null
+        val allMethodInfos = mutableListOf<MethodInfo>()
```
mikesamuel, line 58: StringIndex has a connected method for equality, but NoStringIndex was complaining it needed an implementation from Equatable.  This is trying to make sure that a masking methods' implementation satisfies a deeper abstract one.

frontend/src/commonTest/kotlin/lang/temper/frontend/typestage/TyperTest.kt
```diff
@@ -249,7 +249,7 @@ class TyperTest {
         |    // which may be required for backwards compatibility.
         |
         |    first(str, int);
-        |/// ┗━━━━━━━━━━━━━┛ : MapKey
+        |/// ┗━━━━━━━━━━━━━┛ : AnyValue
```
mikesamuel, line 252: string and int no longer have a common upper bound of MapKey since MapKey<String> is not MapKey<Int32>

fundamentals/src/commonMain/kotlin/lang/temper/type/ButNotRecursively.kt
```diff
@@ -0,0 +1,31 @@
+package lang.temper.type
+
+import java.util.IdentityHashMap
+
+class ButNotRecursively {
```
mikesamuel, line 21: Will document this

fundamentals/src/commonMain/kotlin/lang/temper/type/MentionedTypes.kt
```diff
@@ -20,7 +20,6 @@ fun addTypeNamesMentionedTo(typeActual: TypeActual, out: MutableSet<ResolvedName
         InvalidType -> Unit
         is OrType -> typeActual.members.forEach { addTypeNamesMentionedTo(it, out) }
         is AndType -> typeActual.members.forEach { addTypeNamesMentionedTo(it, out) }
-        is InfiniBinding -> Unit
```
mikesamuel, line 23: The planned way of avoiding recursivity required constructing types in particular ways.

fundamentals/src/commonMain/kotlin/lang/temper/type/TypeContext.kt
```diff
@@ -340,15 +342,44 @@ class TypeContext {
...
+        // and when we check if `T` is a subtype of each of U's
+        // upper bounds, we end up checking if T is a subtype of MapKey<T>,
+        // which it is by its own `extends` clause.
+
+        return formal.upperBounds.filter { upperBound ->
+            val adjustedUpperBound = MkType.map(upperBound, allBindings)
+
+            !isSubType(actual, adjustedUpperBound)
+        }
+    }
```
mikesamuel, line 371: See commit message around structurally compatible type binding checks in InferBounds.

fundamentals/src/commonMain/kotlin/lang/temper/type/TypeDefinition.kt
mikesamuel, line 1: A lot of detail work to thread in ButNotRecursively.

tjpalmer:
Maybe the ci failures are changes to error messaging for already bad code?

mikesamuel:
Looking into the test failures.

tooling/src/jvmTest/kotlin/lang/temper/tooling/ListLogEntriesTest.kt
```diff
@@ -21,9 +21,7 @@ class ListLogEntriesTest {
         val found = context.moduleData.logEntries.map { it.messageText }
         val expected = listOf(
             "Cannot assign to String from Fish!",
-            // We don't get this anymore when the numeric types share Equatable as a supertype.
-            // TODO Will this be a non-issue after new overloading is finished?
-            // "No callee matches inputs [Invalid, Invalid] among [(Int32, Int32) -> Int32, (Int32) -> Int32, (Int64, Int64) -> Int64, (Int64) -> Int64, (Float64, Float64) -> Float64, (Float64) -> Float64]!",
+            "No callee matches inputs [Invalid, Invalid] among [(Int32, Int32) -> Int32, (Int64, Int64) -> Int64, (Float64, Float64) -> Float64]!",
```
mikesamuel, line 24: This error is back because Equatable is not a supertype of distrinct types anymore.  String, for example, has `Equatable<String>` as a supertype, but not Equatable.

The unary `+`'s are filtered out though.

mikesamuel:
Fixed the test failures.  Minor changes in error messages reported.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -920,8 +920,8 @@ class RustBackendTest {
             |let things = [1, 2] as List<AnyValue>; // making a list of ints but expect list of AnyValue
             |let more = [1 as AnyValue, 2 as AnyValue]; // cast elided in frontend?
             |let still: List<AnyValue> = [1, 2]; // again list of ints treated as list of AnyValue
-            |let yet = [1, "two"];
-            |let yetAgain: List<MapKey> = yet; // MapKey only supported for constraint, not yet explicit value type
+            |let yet = ["one", "two"];
+            |let yetAgain: List<MapKey<String>> = yet; // MapKey only supported for constraint, not yet explicit value type
```
tjpalmer, line 924: This limitation might be ok, although I'd be tempted just to say String instead of `MapKey<String>` in most real use cases.

cli/src/test/kotlin/lang/temper/cli/repl/ReplTest.kt
```diff
@@ -1161,8 +1164,8 @@ class ReplTest {
         assertPending(
             """
                 |1: new Map(5)
-                |           ⇧
-                |[interactive#1:1+8-9]@G: Actual arguments do not match signature: <in K__32 extends AnyValue & MapKey, out V__33 extends AnyValue>(List<Pair<K__32, V__33>>) -> Map<K__32, V__33> expected [List<Pair<MapKey, AnyValue>>], but got [Int32]
+                |   ┗━━━━━━━━┛
+                |[interactive#1:1+0-10]@G: Could not infer type actuals for <in K__33 extends MapKey<K__33>, out V__34 extends AnyValue>(List<Pair<K__33, V__34>>) -> Map<K__33, V__34> given [Int32] in context unknown
```
tjpalmer, line 1168: I like the new error message better.

frontend/src/commonMain/resources/core/core.temper
```diff
@@ -13,7 +13,10 @@ export interface AnyValue {}
  * User-defined types can't currently extend Equatable, but we expect to support
  * this in the future.
  */
-export sealed interface Equatable {}
+export sealed interface Equatable<EQ extends AnyValue> {
+  @operator("==")
+  eq(other: EQ): builtins.Boolean;
+}
```
tjpalmer, line 19: And we also expect operators to work without an associate interface, right?

frontend/src/commonMain/resources/core/core.temper
```diff
@@ -1499,6 +1521,16 @@ export class Int32 extends MapKey {
   /** `this - 1`, the predecessor value used by [snippet/builtin/--]. */
   @connected
   public pred(): builtins.Int32 { content - 1 }
+
+  /** `-1` if `this < 0`, `0` if `this == 0`, or otherwise `1`. */
+  @connected
+  public get signum(): builtins.Int32 {
+    (content >> 31) | (-content >>> 31)
+  }
```
tjpalmer, line 1529: This signum takes some thinking for me.

functional-test-suite/src/commonMain/resources/types/map/map.temper.md
```diff
@@ -2,7 +2,7 @@
 
 Here we test the `Map` type as well as related types `Mapped`, `MapBuilder`, and
 `Pair`. These are similar to `Listed`, `ListBuilder`, and `ListEntry`. We
-also have marker interface `MapKey`, which might allow for custom key types in
+also have interface `MapKey`, which might allow for custom key types in
```
tjpalmer, line 5: Good catch on the word "marker" here, since that's not an ideal description after this pr.

fundamentals/src/commonTest/kotlin/lang/temper/type/TypeContextTest.kt
```diff
@@ -37,7 +37,7 @@ class TypeContextTest {
         """.trimMargin(),
     ) {
         // Temporarily treat all fun interfaces with the same signature as
-        // subtypes of one another at least until we aren't doing FunctionType<->Type2
+        // subtypes of one-another at least until we aren't doing FunctionType<->Type2
```
tjpalmer, line 40: I wouldn't expect a hyphen to be correct here, but I haven't looked it up, and I could be wrong.

fundamentals/src/commonMain/kotlin/lang/temper/type/Types.kt
```diff
@@ -172,13 +206,18 @@ private fun renderParenthesized(outer: TypeOpPrecedence, inner: StaticType, toke
  * A name that may be used to reference the special [snippet/type/Top] type.
  */
 object TopType : SimpleType(), Stayless {
-    override fun destructure(structureSink: StructureSink) {
+    override fun destructure(structureSink: StructureSink, bnr: ButNotRecursively) {
```
tjpalmer, line 209: ButNotRecursively does end up invading a lot.

tjpalmer (approved):

frontend/src/commonMain/resources/core/core.temper
mikesamuel: I should probably cite Hacker's delight.

https://www.informit.com/articles/article.aspx?p=1959565&seqNum=8

Screenshot 2026-09-16 at 2 23 11 pm

The `>>` with `u` and `s` superscripts are unsigned/signed right shift.

tjpalmer: Clever trick for sure.

mikesamuel: I went ahead and hacked together a simple test for that.

```java
    public static short signum(short x) {
        int content = x;
        return (short) ((content >> 31) | (-content >>> 31));
    }

    public static short slowSignum(short content) {
        if (content == 0) { return 0; }
        if (content < 0) { return -1; }
        return 1;
    }

    public static void main(String... argv) {
        short i = Short.MIN_VALUE;
        while (true) {
            short slow = slowSignum(i);
            short branchless = signum(i);
            if (slow != branchless) {
                System.err.println("i=" + i + ", slow=" + slow + ", branchless=" + branchless);
                break;
            }
            if (i == Short.MAX_VALUE) { break; }
            i += 1;
        }
    }
```

fundamentals/src/commonTest/kotlin/lang/temper/type/TypeContextTest.kt
mikesamuel: I'm just making IntelliJ's grammar checker happy.  It's kind of frustrating because it flags an awful lot of things that imo aren't big deals: it wants the hyphen in "super-type" but not in "subtype."

But since comments in Kotlin do end up in docs/for-users, I'm loathe to do without any grammar checking for comments.

fundamentals/src/commonMain/kotlin/lang/temper/type/Types.kt
mikesamuel: Yeah.  I could probably make the code footprint smaller, but it is on the critical path for .equals and .hashCode.

I suppose we could audit where we create TypeShapes and TypeFormals, and maybe hack in some dynamic analysis to check that we don't create multiple instances with the same name.

That might allow us to assume that TypeFormals are equal when their names are equal which might cut that chain specifically for equals and hashCode.

fundamentals/src/commonTest/kotlin/lang/temper/type/TypeContextTest.kt
tjpalmer: > But since comments in Kotlin do end up in docs/for-users, I'm loathe to do without any grammar checking for comments.

I wonder if we should run the generated docs through a grammar checker. Even just through AI, but then still only make the changes we agree with.

frontend/src/commonMain/resources/core/core.temper
mikesamuel: Yes.  Operations are just method or function applications.

PR #492, opened 2026-09-17
Add decorators for keeping code in prod and test

tjpalmer:
- Work toward #456
- Add `@keep` and `@keepTest` decorators to prevent pruning
  - Doesn't include any mechanism to control this per backend
- Add tests that these serve as reachability roots
- Test in BuildTest just for be-py so far that the marked things end up in the right places
- See my discussion in self-review of some alternate names for `@keepTest` especially, and if we want something else, we can change it
- Start placeholders for tests of calling kept functions from backend connected code, but I can't actually do such calling until we make pretty names for non-exported functions, which we likely want to do anyway
  - I'll start pretty naming for non-exported things in a separate PR so these tests can be finished in the future


cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
```diff
@@ -368,13 +368,21 @@ class BuildTest {
...
         topDir.withTextOf("temper.out/py/banana/tests/test_banana.py") { text ->
             assertContains(text, "temper_std.testing")
             assertContains(text, "def test___twiceWorks")
             assertContains(text, "class _Something")
             assertContains(text, "def halve_value_in")
             assertNotContains(text, "nobody_wants_me")
+            // Keep decoration and reachability tests. Now test should be here but not prod.
+            assertNotContains(text, "keep_me_for_prod")
+            assertContains(text, "keep_me_for_test")
+            assertNotContains(text, "prune_me")
```
tjpalmer, line 385: Different "not" handling for the first two in prod vs in test.

cli/src/test/resources/buildDirs/banana/more.temper.md
```diff
@@ -17,6 +30,13 @@
       assert(twice(0.5) == something.value);
     }
 
+The following should be kept in test code by backends, even though unused.
+
+    @keepTest
+    let keepMeForTest(): Void {
+      console.log("I'm yet another that's here for checking in on.");
+    }
```
tjpalmer, line 38: These tests aren't as elaborate as the generate code stage test, but they help verify the end build results better. I guess technically there's some overlap between this and backend test capabilities but with different pros and cons.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/ReachabilityTracer.kt
```diff
@@ -185,32 +187,39 @@ internal class ReachabilityTracer {
             // Various mutually exclusive things we care about.
             when {
                 tree is DeclTree -> {
+                    val metadata = tree.parts?.metadataSymbolMap ?: return@tree VisitCue.SkipOne
                     try {
-                        val metadataSymbolMap = tree.parts?.metadataSymbolMap ?: return@tree VisitCue.SkipOne
```
tjpalmer, line 189: Moved this up a scope for later access. And renamed it to make it shorter.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/ReachabilityTracer.kt
```diff
@@ -185,32 +187,39 @@ internal class ReachabilityTracer {
...
+                                val relatedType = metadata[fromTypeSymbol]?.target
+                                    ?: metadata[typePlaceholderSymbol]?.target ?: return@tree VisitCue.SkipOne
                                 val relatedTypeName = relatedType.typeNameContained() ?: return@tree VisitCue.SkipOne
+                                val (typeKept, typeTestKept) =
+                                    when (val typeDecl = relatedType.typeDefContained()?.stayLeaf?.incoming?.source) {
+                                        is DeclTree -> typeDecl.parts?.metadataSymbolMap?.let { typeMetadata ->
+                                            (keepSymbol in typeMetadata) to (keepTestSymbol in typeMetadata)
+                                        }
+                                        else -> null
+                                    } ?: (false to false)
```
tjpalmer, line 219: This handling is for the local case of associating type members with types. But all the logic around the decorators is mostly to have new ways of tracking prod or test export roots.

frontend/src/commonMain/kotlin/lang/temper/frontend/generate/ReachabilityTracer.kt
```diff
@@ -227,8 +240,19 @@ internal class ReachabilityTracer {
...
+                                    testNames.add(name)
+                                    // This effective root also needs reachability marked.
+                                    tree.replace(tree.size until tree.size) {
+                                        V(vReachSymbol)
+                                        V(vTestSymbol)
+                                    }
+                                }
+                                // Not a root, so we need to evaluate later.
+                                else -> unreachedMap[name] = tree
+                            }
```
tjpalmer, line 255: And here is the top-level def case. For the test reach marking above, we already mark things reached from tests elsewhere, but we don't mark the tests themselves. I guess because they are already tests. And they were the only test roots before. If we don't mark these new test roots but keep them out of `unreachedMap` then they get seen as prod reachable. That's why the marking happens here.

And I have a test elsewhere to make sure these don't get marked twice if also reachable from an actual test or other test root.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/export-some/expect/generateCode.temper
```diff
@@ -1,77 +1,109 @@
-let console#47;
-console#47 = getConsole();
-@fn let `test//`.exportedFunction, @fn conditionallyExportReachable__20, @fn @reach(\test) transitivelyTestReachable__21, @fn exportAndTestReachable__22, @fn initReachable__23, @fn transitivelyInitReachable__24, @fn @reach(\none) unreachableFunction__25;
+let console#70;
+console#70 = getConsole();
+@fn let `test//`.exportedFunction, @fn conditionallyExportReachable__30, @fn @reach(\test) transitivelyTestReachable__31, @fn exportAndTestReachable__32, @fn initReachable__33, @fn transitivelyInitReachable__34, @fn @reach(\none) unreachableFunction__35;
```
tjpalmer, line 3: Renumbering makes this diff more complicated. Some extra things in the way also make the diff a harder to follow.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/export-some/expect/generateCode.temper
```diff
@@ -1,77 +1,109 @@
...
 TestReachableClass__1 = type (TestReachableClass__1);
-@fn @test("testCase") let testCase__26;
+@fn @test("testCase") let testCase__36;
 @typeDecl(UsedOnlyAsPropertyType__3) @stay let UsedOnlyAsPropertyType__3;
 UsedOnlyAsPropertyType__3 = type (UsedOnlyAsPropertyType__3);
-let `test//`.exportedInt;
+@fn let keepReachable__37, @fn @reach(\test) testKeepReachable__38, @fn @keep kept__39;
+@typeDecl(KeptClass__4) @stay @keep let KeptClass__4;
+KeptClass__4 = type (KeptClass__4);
+@fn @keepTest @reach(\test) let testKept__40, @fn @keepTest @reach(\test) testKeptAndTestCalled__41, `test//`.exportedInt;
```
tjpalmer, line 14: Here, note that testKeepReachable, testKept, and testKeptAndTestCalled are all marked `@reach(\test)` once. And `keepReachable` and `kept` are both left at default prod.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/export-some/expect/generateCode.temper
```diff
@@ -1,77 +1,109 @@
...
-testCase__26 = (@stay fn testCase /* return__17 */: Void {
-    return__17 = void;
-    new TestReachableClass__1()
+@fn @method(\constructor) @visibility(\public) @stay @fromType(ExportedClass) let constructor__54;
+constructor__54 = (@stay fn constructor(@impliedThis(ExportedClass) this__13: ExportedClass, propertyOfExportedClass__55 /* aka propertyOfExportedClass */: UsedOnlyAsPropertyType__3) /* return__14 */: Void {
+    return__14 = void;
+    setp(propertyOfExportedClass__51, this__13, propertyOfExportedClass__55)
+});
+@method(\methodOfTestReachable) @visibility(\private) @fn @stay @fromType(TestReachableClass__1) @reach(\test) let methodOfTestReachable__56;
+methodOfTestReachable__56 = (@stay fn methodOfTestReachable(@impliedThis(TestReachableClass__1) this__2: TestReachableClass__1) /* return__18 */: Void {
```
tjpalmer, line 67: Here, `methodOfTestReachable` and `methodOfTestReachable` also are `@reach(\test)`.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/export-some/work/test/test.temper
```diff
@@ -1,20 +1,55 @@
...
 ) {
   private let methodOfExportedClass(): Void { exportAndTestReachable() }
 }
+
 class TestReachableClass {
-  private let methodOfTestReachable(): Void { transitivelyTestReachable(); exportAndTestReachable() }
+  private let methodOfTestReachable(): Void {
+    transitivelyTestReachable();
+    exportAndTestReachable();
+  }
```
tjpalmer, line 25: Reformatted to make both calls clearer.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/export-some/work/test/test.temper
```diff
@@ -1,20 +1,55 @@
...
 }
-@test("testCase") let testCase(): Void { new TestReachableClass(); }
+
+@test("testCase") let testCase(): Void {
+  new TestReachableClass();
+  testKeptAndTestCalled();
+}
```
tjpalmer, line 31: Added `testKeptAndTestCalled();` here.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/export-some/work/test/test.temper
```diff
@@ -1,20 +1,55 @@
...
+
 class UsedOnlyAsPropertyType {}
+
 initReachable();
+
+// Now also try out decorations for keeping.
+
+let keepReachable(): Void { console.log(""); }
+let testKeepReachable(): Void { console.log(""); }
```
tjpalmer, line 40: Just to make sure these are kept by keepers.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/generate-code/export-some/work/test/test.temper
```diff
@@ -1,20 +1,55 @@
...
+class KeptClass {
+  public keptClassMethod(): Void {}
+}
+
+@keepTest
+let testKept(): Void { testKeepReachable(); }
+
+// Double-checking this one doesn't get marked test-reachable twice.
+@keepTest
+let testKeptAndTestCalled(): Void { console.log(""); }
```
tjpalmer, line 55: And the various explicit keepers.

I'm unsure if `@keepTest` or `@keepForTest` or `@testKeep` or even "keep" vs "kept" is best for these decorations. I also considered trying an optional `@keep("test")` vs `@keep` for an optional decorator arg. Not sure if we have infrastructure for that, and then it also encourages other string values, but I don't know what they'd be.

functional-test-suite/src/commonMain/resources/functions/connected/connected.temper.md
```diff
@@ -78,3 +65,18 @@ And reuse the same hider instance just for fun.
 new Hider(3).plus(4): 7
 new Hider(3).times(4): 12
 ```
+
+## Kept definitions
+
+TODO Use this for backends as we provide stable naming for unexported things.
```
tjpalmer, line 71: Note the comment that I don't actually use the below yet from connected code in any backends because its name still gets mangled (well, suffixed) everywhere. So this needs tested in future PRs. I considered leaving this funtest change out of this PR after starting this, but I figured I might as well leave the reminder for later use.

fundamentals/src/commonMain/kotlin/lang/temper/value/NamesAndSymbols.kt
```diff
@@ -527,6 +527,15 @@ val vImportedSymbol = Value(importedSymbol)
...
 val vReachSymbol = Value(reachSymbol)
 
+/**
+ * Marker to treat unreached code as reachable in order to keep it in translations.
+ * The primary use case is helper code for use by backend connected code.
+ */
+val keepSymbol = Symbol("keep")
+
+/** Same as [keepSymbol] but for test reachability. */
+val keepTestSymbol = Symbol("keepTest")
```
tjpalmer, line 537: See my discussion of this naming in the generate code stage test, in context of usage.

interp/src/commonMain/kotlin/lang/temper/interp/BuiltinEnvironment.kt
```diff
@@ -180,6 +182,23 @@ private object Builtins {
...
+             * See also [snippet/builtin/@keepTest].
+             */
+            keyPair(MetadataDecorator(keepSymbol, "@keep") { void }),
+
+            /**
+             * <!-- snippet: builtin/@keepTest -->
+             * # `@keepTest` decorator
+             * Acts like [snippet/builtin/@keep] but for test reachability.
+             */
+            keyPair(MetadataDecorator(keepTestSymbol, "@keepTest") { void }),
```
tjpalmer, line 200: I presume `@keep("test")` would be notably more code than this, also.

cli/src/test/resources/testing/connecteds/bitset.temper.md
```diff
@@ -71,3 +71,5 @@ Just about any basic test of functionality will do here.
       bitset.add(5050);
       assert(bitset.contains(5050));
     }
+
+TODO Test connected used by the test and a `@keepTest` helper for that.
```
tjpalmer, line 75: This is one of two test cases I commented or got underway but can't finish until I start stabilizing names for unexported things.

cli/src/test/resources/buildDirs/banana/more.temper.md
```diff
@@ -9,6 +9,19 @@
...
+The following should be kept in translated code, even though unused.
+
+    @keep
+    let keepMeForProd(): Void {
+      console.log("I'm just here for checking in on.");
+    }
+
+And for contrast, the following should be pruned out in backends that prune.
+
+    let pruneMe(): Void {
```
mikesamuel, line 21: When I am old and wrinkly like a prune, I plan to bother my kids with "why don't you call me more often."

cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
```diff
@@ -368,13 +368,21 @@ class BuildTest {
             assertNotContains(text, "temper_std.testing")
             assertContains(text, "def twice")
             assertNotContains(text, "nobody_wants_me")
+            // Keep decoration and reachability tests. Prod should be here but not test.
+            assertContains(text, "keep_me_for_prod")
+            assertNotContains(text, "keep_me_for_test")
+            assertNotContains(text, "prune_me")
```
mikesamuel: I would feel better about these negative tests if it was more like

```kt
    assertNotContains(text, Regex("""(?i)keep.?me.?for.?test"""))
```

Seems like that'd be more robust across things like underscore -> camelcase changes to backend conventions.  Unlikely to cause problems, but I checked and `kotlin.test` already has an overload for (String, Regex) at https://github.com/JetBrains/kotlin/blob/2.4.20/libraries/kotlin.test/common/src/main/kotlin/kotlin/test/Assertions.kt#L776
Maybe our handwritten assertNotContains(String, String) could be done on top of a regex test.

mikesamuel (approved):
I think the testonly aspect of it is really nice.

That kind of thing is good for software supply chain stuff where the fewer prod dependencies the better.

cli/src/test/resources/buildDirs/banana/more.temper.md
tjpalmer: > When I am old and wrinkly like a prune, I plan to bother my kids with "why don't you call me more often."

👴

cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
tjpalmer: > I would feel better about these negative tests if it was more like
> 
> ```kotlin
>     assertNotContains(text, Regex("""(?i)keep.?me.?for.?test"""))
> ```

Perhaps, although this is for be-py, specifically. But maybe we could change our formatting sometime and forget about this.

For the "keep_me_for_prod" and "keep_me_for_test" cases, we're safe already because we flip-flop checking presence and absence in different files. But "prune_me" (👴) is potentially risky.

tjpalmer: Added regex testing against prune me. Excerpt from failure when I changed the regex to "keep.*me":

```py
Expected the char sequence from <temper.out/py/banana/banana/banana.py> to exclude the regex, case insensitively
char sequence
...
  def keep_me_for_prod_12() -> 'None':
      console_40.log("I'm just here for checking in on.")
regex
  keep.*me
```

Since factored substring checking, I also checked that failure manually. Excerpt:

```py
Expected the char sequence from <temper.out/py/banana/banana/banana.py> to exclude the substring
char sequence
...
  def keep_me_for_prod_12() -> 'None':
      console_40.log("I'm just here for checking in on.")
substring
  keep_me_for_prod
```

PR #493, opened 2026-09-18
Improve python non-exported naming

tjpalmer:
- Work toward #456
- Avoid suffixing non-exported top levels so they can have stable names for calling from connected code
- Track naming better to avoid duplicate name definitions
- Add a "private" name reach that's separate from "internal"
- After this pr and #492, we can test such calls (and other next steps would include working more backends to stabilize naming)
- Also prefix non-local, non-exported things with `_` for presumably improved python tooling
- Some limitations with this pr:
  - Still no attempt at avoiding suffixes on locals (and doing that well would involve tracking scopes)
  - Some things still sneak through without `_` prefix or now have suffix when they didn't before, but we can possibly improve still in future PRs
- I did test this also on this branch:

```
prismora> temper test -b py
Tests passed: 12 of 12
```


be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -32,6 +32,7 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
     private var nameCounter = 0
     private val supportCodeMap = mutableMapOf<SupportCode, OutName>()
     private val nameInfo = mutableMapOf<Pair<ModuleName, ResolvedName>, PyInfo>()
+    private val chosenNames = mutableMapOf<Pair<ModuleName, OutName>, ResolvedName>()
```
tjpalmer, line 40: For avoiding duplicates.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -66,7 +67,16 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
             val (localName, descriptor) = findLocal(module, name)
             val kind = descriptor?.idKind() ?: TmpL.IdKind.Value
             val reach = descriptor?.idReach(ignoreImport = true) ?: TmpL.IdReach.Internal
-            val outName = pythonizeName(localName, kind, reach)
+            // TODO Also treat params as non-local?
+            val local = context?.node?.let { it is TmpL.TopLevel || it is TmpL.DotAccessible } == false
+            val outName = pythonizeName(localName, kind, reach, local = local).let { outName ->
+                when (chosenNames.putIfAbsent(module to outName, name)) {
```
tjpalmer, line 78: Here's where we store while avoiding duplicates. And my `local` tracking could possibly become a 4th "reach" option, but I wasn't convinced to do that here yet.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -66,7 +67,16 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
...
+                    null, name -> outName
+                    else -> pythonizeName(localName, kind, TmpL.IdReach.Private, local = local).also { unique ->
+                        chosenNames[module to unique] = name
+                    }
```
tjpalmer, line 82: Basically just treat any collision as a "private" name because we need it unique, and private here keeps the suffix.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -89,7 +99,7 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
                 importedName = null,
             )
             lookup.allModules.forEach { module ->
-                nameInfo[module to name] = info
+                nameInfo.putIfAbsent(module to name, info)
```
tjpalmer, line 107: This was overwriting previous choices. I wasn't sure if someone was expecting common names across all modules like the loop is doing here, but funtests seem to be fine.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -160,16 +170,23 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
...
                 TmpL.IdKind.Type ->
                     "_$styledName" // won't be keyword if not starting with `_`, which has other issues
                 // Type formals still need suffices for uniqueness right now.
-                TmpL.IdKind.TypeFormal, TmpL.IdKind.Value ->
-                    concatIfVerbose(styledName, "_$uid") // numeric suffix, won't be a keyword
+                TmpL.IdKind.TypeFormal, TmpL.IdKind.Value -> {
+                    val adjusted = when {
+                        local -> styledName
+                        else -> "_$styledName" // hint hidden if not local
+                    }
```
tjpalmer, line 190: Here's where private non-locals get prefixed.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -160,16 +170,23 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
...
+                    concatIfVerbose(adjusted, "_$uid") // numeric suffix, won't be a keyword
+                }
             }
+            TmpL.IdReach.Internal -> "_$styledName"
```
tjpalmer: And internal (within a module but not private to a class or other local scope) just gets prefixed and not suffixed.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyTranslatorExt.kt
```diff
@@ -48,14 +48,15 @@ fun PyTranslator.dotName(name: TmpL.DotName, source: ResolvedName? = null): OutN
...
+        TmpL.IdReach.Internal, TmpL.IdReach.Private -> pyNames.name(member.name)
     }
 }
 
 fun PyTranslator.methodReferenceNameText(method: TmpL.MethodReference): String {
     return method.method?.let { methodShape ->
         when (methodShape.visibility.idReach()) {
-            TmpL.IdReach.Internal -> (methodShape.name as? ResolvedName)?.let { pyNames.name(it).outputNameText }
+            TmpL.IdReach.Internal, TmpL.IdReach.Private ->
+                (methodShape.name as? ResolvedName)?.let { pyNames.name(it).outputNameText }
```
tjpalmer, line 59: Some of this was just me trying to make good guesses for behavior and calling it good when things seemed ok.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -22,21 +22,65 @@ class PyBackendTest {
...
-            |from typing import Any as Any0, TypeVar as TypeVar1, Generic as Generic2, Callable as Callable3
-            |T_1 = TypeVar1('T_1', bound = Any0)
-            |U_3 = TypeVar1('U_3', bound = Any0)
-            |class Thing(Generic2[T_1]):
+            |from typing import Any as Any1, TypeVar as TypeVar2, Generic as Generic3, Callable as Callable4, Sequence as Sequence5
+            |from builtins import int as int6, bool as bool7
+            |from temper_core import list_filter as list_filter0
+            |_list_filter_22 = list_filter0
+            |T_1 = TypeVar2('T_1', bound = Any1)
+            |U_3 = TypeVar2('U_3', bound = Any1)
```
tjpalmer, line 35: Prefixing these type vars might be good in the future.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -22,21 +22,65 @@ class PyBackendTest {
...
             |    "They're the same thing!\n\nt__0: T__0\n  The thing to return\n"
-            |    return t_8
+            |    return t_9
+            |_ls: 'Sequence5[int6]' = (0, 1, 2, 3)
+            |def _fn(x_16: 'int6', /) -> 'bool7':
+            |    return x_16 & 1 == 0
+            |evens: 'Sequence5[int6]' = _list_filter_22(_ls, _fn)
+            |def _fn_20(x_17: 'int6', /) -> 'bool7':
+            |    return x_17 & 1 == 1
+            |odds: 'Sequence5[int6]' = _list_filter_22(_ls, _fn_20)
```
tjpalmer, line 52: I still haven't made the `fn` things temporary names, but my uniqueness tracking does force a suffix on the second one.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -22,21 +22,65 @@ class PyBackendTest {
...
+            |export let tame(name: String): String { new Fame().blame(name) }
+        """.trimMargin(),
+        want = """
+            |from builtins import str as str1
+            |from temper_core import str_cat as str_cat0
+            |_str_cat_21 = str_cat0
+            |class _Fame:
+            |    __slots__ = ()
+            |    def _shame_12(this_1, name_13: 'str1', /) -> 'str1':
+            |        return _str_cat(name_13)
```
tjpalmer, line 75: Prefixed and also suffixed private method.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -22,21 +22,65 @@ class PyBackendTest {
...
+            |    def blame(this_2, name_16: 'str1', /) -> 'str1':
+            |        return _str_cat_21(this_2._shame_12(name_16), '!')
+            |    def __init__(this, /) -> None:
+            |        pass
+            |def _str_cat(name_10: 'str1', /) -> 'str1':
+            |    return _str_cat_21('Hello, ', name_10)
```
tjpalmer, line 81: Prefixed but not suffixed non-exported `str_cat` -> `_str_cat`. Just realized I probably should have called it `strCat` in temper, but the right thing gets tested anyway. And the imported names *are* temporaries, so they get suffixed automatically in my current code.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -142,29 +186,29 @@ class PyBackendTest {
...
             |from typing import Dict as Dict4, Union as Union7, Sequence as Sequence8
             |from builtins import str as str5, int as int6, bool as bool9, isinstance as isinstance10, len as len0
-            |len_16 = len0
-            |list_get_17 = list_get1
-            |int_add_18 = int_add2
-            |console_4: 'LoggingConsole3' = LoggingConsole3(__name__)
+            |_len_16 = len0
+            |_list_get_17 = list_get1
+            |_int_add_18 = int_add2
+            |_console_4: 'LoggingConsole3' = LoggingConsole3(__name__)
```
tjpalmer, line 192: These get ugly, but they really should be prefixed/hidden.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -268,7 +312,7 @@ class PyBackendTest {
             |        pass
             |class C:
             |    __slots__ = ()
-            |    def __init__(this, /) -> None:
+            |    def __init__(this_0, /) -> None:
```
tjpalmer, line 315: My uniqueness apparently prevents all `__init__` self vars from avoiding suffix, which apparently we'd managed before. But I think we'll be ok. Would be better to call them all `self` at some future point, anyway. And non-init self vars were already getting suffixed. So I didn't try to work around this.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -654,11 +698,11 @@ class PyBackendTest {
             |            from temper_core import LoggingConsole as LoggingConsole2, adapt_generator_factory as adapt_generator_factory3, str_cat as str_cat0, async_launch as async_launch1
```
tjpalmer, line 698: I still haven't gotten these original imports prefixed yet here.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -654,11 +698,11 @@ class PyBackendTest {
...
             |            from builtins import str as str5, Exception as Exception7
             |            from typing import Union as Union6, Generator as Generator8
-            |            str_cat_21 = str_cat0
-            |            async_launch_22 = async_launch1
-            |            console_5: 'LoggingConsole2' = LoggingConsole2(__name__)
+            |            _str_cat_21 = str_cat0
+            |            _async_launch_22 = async_launch1
+            |            _console_5: 'LoggingConsole2' = LoggingConsole2(__name__)
             |            @adapt_generator_factory3
```
tjpalmer, line 704: Example of a direct import name being used later instead of one of our bonus renames that we seem to do (such as in the immediately preceding lines).

be/src/commonMain/kotlin/lang/temper/be/names/DescriptorChain.kt
```diff
@@ -88,12 +88,25 @@ fun DescriptorChain.idKind(): TmpL.IdKind = when (node) {
...
+    when (reach) {
+        TmpL.IdReach.Internal -> when {
+            // Private for imports fails at least for be-py handling. Not sure yet about others.
+            // node is TmpL.Import -> TmpL.IdReach.Private
+            node is TmpL.TopLevelDeclaration -> TmpL.IdReach.Internal
+            else -> TmpL.IdReach.Private
+        }
+        else -> reach
+    }
 }
```
tjpalmer, line 110: Needed some extra logic to treat non-top-levels as private.

be/src/commonMain/kotlin/lang/temper/be/tmpl/tmpL.out-grammar
```diff
@@ -993,7 +993,13 @@ LocalFunctionDeclaration;
...
 
 enum IdKind = Type | TypeFormal | Value;
-enum IdReach = External | Internal;
+
+/**
+ * This meshes export or not with public or private.
+ * Export and public are visible externally.
+ * Internal is visible within a module, but private only inside an instance.
+ */
+enum IdReach = External | Internal | Private;
```
tjpalmer, line 1002: Again, started wondering at the end if I should add an `| Local` here, but I didn't do that yet.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLHelpers.kt
```diff
@@ -358,13 +358,13 @@ fun TmpL.FunctionDeclaration.idReach() = when (name.name) {
...
-    TmpL.IdReach.Internal
+    TmpL.IdReach.Private
 }
 
 fun TmpL.Visibility.idReach() = if (this.ordinal >= TmpL.Visibility.Protected.ordinal) {
     TmpL.IdReach.External
 } else {
-    TmpL.IdReach.Internal
+    TmpL.IdReach.Private
 }
```
tjpalmer, line 368: Visibility is on members, and they go private, not internal. I don't think most backends care about this enum, though.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyBackend.kt
```diff
@@ -287,7 +287,11 @@ class PyBackend private constructor(
     private var pyNames: PyNames? = null
 
     override fun translate(finished: TmpL.ModuleSet): List<OutputFileSpecification> {
-        val pyNames = PyNames(LookupNameVisitor().visit(finished), abbreviated = config.abbreviated)
+        val pyNames = PyNames(
+            visit = LookupNameVisitor().visit(finished),
+            abbreviated = config.abbreviated,
+            genre = finished.genre,
+        )
```
tjpalmer, line 294: Surprised it didn't already have it, but give genre to naming now.

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
```diff
@@ -160,15 +175,25 @@ class PyNames(visit: LookupNameVisitor?, private val abbreviated: Boolean = fals
...
+                        local -> styledName
+                        else -> "_$styledName" // hint hidden if not local
+                    }
+                    concatIfVerbose(adjusted, "_$uid") // numeric suffix, won't be a keyword
+                }
+            }
+            TmpL.IdReach.Internal -> when (genre) {
+                Genre.Documentation -> styledName
+                Genre.Library -> "_$styledName"
             }
```
tjpalmer, line 197: The doc gen test expected pretty names even when not exported. And internal before didn't get prefixed. Did get suffixed or not depending on `abbreviated`. But in any case, this keeps the doc test unchanged, which I presume is wanted.

cli/src/test/kotlin/lang/temper/cli/BuildTest.kt
```diff
@@ -373,7 +373,7 @@ class BuildTest {
             assertContains(text, "temper_std.testing")
             assertContains(text, "def test___twiceWorks")
             assertContains(text, "class _Something")
-            assertContains(text, "def halve_value_in")
+            assertContains(text, "def _halve_value_in")
```
tjpalmer, line 376: We do want this change for this function:

```ts
let halveValueIn(test: Test, something: Something): Void { ... }
```

be-py/src/commonMain/kotlin/lang/temper/be/py/PyNames.kt
mikesamuel: Smells like a bug we got mostly lucky with.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
mikesamuel: Yeah. I don't know what the semantics for type vars are.  Like does typed Python allow specifying type arguments by parameter name, and if so does that require a non-underscored name?

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
```diff
@@ -22,21 +22,65 @@ class PyBackendTest {
...
+            |    return x_17 & 1 == 1
+            |odds: 'Sequence5[int6]' = _list_filter_22(_ls, _fn_20)
+            |
+        """.trimMargin(),
+    )
+
+    @Test
+    fun unexported() = assertGeneratedCode(
+        input = $$"""
+            |// `str_cat` was `game` for rhyming, but `str_cat` lets us conflict with import.
```
mikesamuel, line 60: When dad jokes and code quality collide. 😿

be/src/commonMain/kotlin/lang/temper/be/tmpl/tmpL.out-grammar
mikesamuel: And Local, if it existed, would mean: effectively private because there doesn't exist any string that you could name it that external code could then use to refer to it?

mikesamuel (approved):
Great that you tracked down that assigned-name clobbering bug.

I spent more time looking at the TmpL changes and the backend test changes than the rest.  looks good.

be-py/src/commonTest/kotlin/lang/temper/be/py/PyBackendTest.kt
tjpalmer: > Yeah. I don't know what the semantics for type vars are. Like does typed Python allow specifying type arguments by parameter name, and if so does that require a non-underscored name?

Python uses `[T, U]` syntax for type params and type args, so they wouldn't have syntax for being named. Maybe some reflection access exists, though. Or some other reason to access them that I'm not fresh on.

be/src/commonMain/kotlin/lang/temper/be/tmpl/tmpL.out-grammar
tjpalmer: > And Local, if it existed, would mean: effectively private because there doesn't exist any string that you could name it that external code could then use to refer to it?

Currently, I call them Private, so sort of, but I already needed different rules for them here from class members that are private.

PR #494, opened 2026-09-19
Support relational operators as operators

mikesamuel:
=========================================

The subsections each cover a weakness in the prior relational operator representation and semantics, and how this commit fixes them or takes a step back that will enable a later step forward.

## Frontend Representation

Previously, relational operators (`==`, `!=`, `<`, `<=`, `>=`, `>`, `<=>`) were each supported via a suite of builtin functions.  There were seperate `<` implementations for *Int32*, *String*, *Float64*, etc.

Now, only *Int32* has the full suite, and other builtin types support `==` and `<=>`.  The following desugarings are how the frontend represents these families of related operators internally.

| Operator  | Desugaring                                                    |
|-----------|---------------------------------------------------------------|
| `a == b`  | *EqMacro* desugars to a null-safe use of the `_==_` operator  |
| `a != b`  | `!(a == b)`                                                   |
| `a <=> b` | Represented directly as a *DotHelper* of the `_<=>_` operator |
| `a < b`   | `(a <=> b) < 0`                                               |
| `a <= b`  | `(a <=> b) <= 0`                                              |
| `a >= b`  | `(a <=> b) >= 0`                                              |
| `a > b`   | `(a <=> b) > 0`                                               |

Since `<` is directly supported for *Int32* and `<=>` returns an *Int32*, the desugaring for `<` above always uses a known basic operation needing no recursive desugaring.

Many *BuitinOperatorId* cases have gone away and existing backends no longer need to support them in *SupportNetwork*, and some gaps have been filled. Equality and comparison *BuiltinOperatorId*s now exist for *Int64* and *Boolean*.

## Null safety

Previously, many of the `==` and `!=` accepted `null`s.

Now, the builtin `==` implementations operated on non-null types making it easier to translate them to primitive infix operators.

The *EqMacro* waits until type information is available, then desugars `==` to possibly branching code like the below:

```
let f(a: String?, b: String?): Boolean { a == b }

// ->

let f(a: String?, b: String?): Boolean {
  if (isNull(a) {
    isNull(b)
  } else if (isNull(b)) {
    false
  } else {
    notNull(a) == notNull(b)
    //         ^- string variant of equality
  }
}
```

*EqMacro* carries around a DotHelper with extensions so this solves user equality: now *StringIndexOption* equality is a method with `@operator("==")` metadata and is not special cased.

With more work to collect candidate resolutions, extension methods for equality and comparison can be supported.

## Generic comparison

And there was a generic comparison algorithm that was not translatable. It "worked" based on runtime type information, so you could say something like `let f<T>(a: T, b: T): Boolean { a < b }` but in languages like PHP and Perl where *Float64* and *String* are represented using the same underlying type there is no way to know whether to perform lexicographic or numeric comparison.  On languages like JavaScript where the same numeric type represents *Int32* and *Float64* there is no way to know whether the sign bit on zeroes needs to be taken into account.

Special hacks to generic comparison supported *StringIndexOption*, but for user classes in general, there was no way to specify the behaviour of `<=>` or `==` for user classes.

As stated above, those hacks are gone.  Also, the *BuiltinOperatorId* for generic comparison and equality are gone, and hacky RTTI functions have been removed from existing backends' support libraries.

## Backend translation burden

This is a case of two steps forward, one step back.

Backends no longer need to support the generic comparison algo, and no longer need to support many variants of `<`, `<=`, etc.

But `a <=> b < 0` is not idiomatic and many backend languages do not support ternary comparison natively.  (Python2 did via `__cmp__` but Python3 moved away from ternary comparison and no longer assigns special semantics to `__cmp__` methods.)

Now *SupportNetwork* allows overriding a method to simplify a use of ternary comparison nested inside a binary infix integer comparison.

Existing backends all use this to resugar `a <=> b < 0` to `a < b` where possible.  It's not always possible.
For example, the Java backend uses both representations:

For *Int64*s, be-java outputs `a < b`.
For *String*s, be-java outputs something like `a.compareTo(b) < 0` except that it uses a core library method because Temper's string ordering is lexicographic by code-point, not UTF-16.

## Cover functions

This commit gets rid of the last remaining uses of cover functions which have stalled replacing *StaticType* with *Type2*.

Now, no cover functions are actually created and a future PR will remove *CoverFunction*s as a class of things the Typer needs to be aware of.


be-cpp/src/commonMain/kotlin/lang/temper/be/cpp/CppSupportNetwork.kt
```diff
@@ -433,6 +417,39 @@ internal object CppSupportNetwork : SupportNetwork {
             }
         }
     }
+
+    override fun simplifyPossibleComparison(
```
mikesamuel, line 421: This is the new SupportNetwork method for simplifying ternary comparisons that are later compared to 0.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
```diff
@@ -89,7 +89,7 @@ class CppBackendTest {
                 |        return temper::core::Int::div_wrap(x, y);
                 |      }
                 |    }catch(const temper::core::TemperBubble & ) {
-                |      if(x != 0) {
+                |      if( ! (x == 0)) {
```
mikesamuel, line 92: This isn't ideal.  We could fix this and not introduce a special != operator by either hacking rendering of `!` in cpp.out-grammar or changing the inline support code for `!`.

Not sure why the be-cpp* token spacer is so bad.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/std/Regex/RegexSupport.cs
```diff
@@ -84,16 +84,16 @@ static List<IntRange> CodeSetToIntRangeSet(CodeSet codeSet)
                     case CodeRange range:
                         ranges.Add(new IntRange(range.Min, range.Max + 1));
                         break;
-                    case Digit:
+                    case DigitSpecial:
```
mikesamuel, line 87: I changed some names in std/regex since `when` clauses that have `Dot -> ...` instead of `is DotSomething -> ...` relied on generic comparison.

be-csharp/src/commonTest/kotlin/lang/temper/be/csharp/CSharpBackendTest.kt
```diff
@@ -1467,7 +1473,7 @@ class CSharpBackendTest {
             |                          {
             |                              string fn__0()
             |                              {
-            |                                  return "expected 1 == (" + C::Core.ConvertToString(1) + ") not (" + C::Core.ConvertToString(1) + ")";
+            |                                  return "expected true";
```
mikesamuel, line 1476: I might try to see if EqMacro can avoid collapsing until after TestMacro has had a chance to capture the string forms of expressions.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaSupportNetwork.kt
```diff
@@ -211,11 +245,25 @@ sealed class JavaSupportCode(
         tokenSink.name(baseName, inOperatorPosition = false)
 }
 
-typealias ExprFactory = JavaLang.(pos: Position, args: List<J.Expression>) -> J.Expression
+typealias ExprFactory = JavaLang.(
+    pos: Position,
+    args: List<J.Expression>,
+    translator: JavaTranslator.ModuleScope,
```
mikesamuel, line 251: Passing in the translator allows access to JavaNames which is useful for seeing whether a type is a reference type or a primitive type.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaSupportNetwork.kt
```diff
@@ -272,11 +326,24 @@ class JavaSeparateStatic(
...
     qualifiedName: QualifiedName,
     opId: BuiltinOperatorId? = null,
-) : JavaSeparate(lang, qualifiedName, opId) {
+    connectedKey: String? = null,
+) : JavaSeparate(lang, qualifiedName, opId, connectedKey) {
     override fun toString(): String = "JavaSeparateStatic($baseName)"
 }
 
-typealias StaticArgBuilder = (ModuleInfo, Position) -> List<J.Argument>
+internal fun JavaSupportCtx.inlineSupport(
```
mikesamuel, line 334: JavaSupportCtx allows the same kind of (Java <= 8 vs Java 17) stuff but also carries enough information to auto-attach builtinOperatorId/connectedKey stuff to support codes.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaSupportNetwork.kt
```diff
@@ -456,24 +491,44 @@ val JavaLang.operatorLe by receiver {
...
+
+    // That means that the first two `==`s below does numeric comparison,
+    // but the last does not.
+    //
+    //      Integer a = new Integer(123);
+    //      Integer b = new Integer(123);
+    //
+    //      a == 123;  // integer equality
+    //      123 == b;  // integer equality
+    //      a == b;    // reference equality
```
mikesamuel, line 511: This is why we need access to names, but JavaTranslator also includes fixes to the translation of notNull(...) to clarify.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaSupportNetwork.kt
```diff
@@ -493,89 +548,35 @@ private fun doubleRelational(
...
 // Miscellany
 
 /** Just be yourself. */
-val JavaLang.identity by receiver { inlineSupport("identity", 1) { _, args -> args[0] } }
+internal val JavaLang.identity by receiver { inlineSupport("identity", 1) { _, args, _ -> args[0] } }
+internal val JavaSupportCtx.identity by receiver { lang.identity }
 
-val JavaLang.isNull by receiver {
-    inlineSupport("isNull", 1, BuiltinOperatorId.IsNull) { pos, args ->
+internal val JavaLang.isNull by receiver {
```
mikesamuel, line 578: Formulaic changes from here on out in this file.

be-java/src/commonMain/kotlin/lang/temper/be/java/JavaSupportNetwork.kt
```diff
@@ -863,95 +872,80 @@ val JavaLang.strCatExpr by receiver {
...
-}
-val JavaLang.stringIndexOptionCompareToLe by receiver {
-    comparison("core.type StringIndexOption.compareTo()::le", JavaOperator.LessEquals)
-}
-val JavaLang.stringIndexOptionCompareToLt by receiver {
-    comparison("core.type StringIndexOption.compareTo()::lt", JavaOperator.LessThan)
-}
-val JavaLang.stringIndexOptionCompareToNe by receiver {
-    comparison("core.type StringIndexOption.compareTo()::ne", JavaOperator.NotEquals)
+internal val JavaSupportCtx.stringIndexOptionCompareToEq by receiver {
```
mikesamuel, line 947: Collapsing all the compareTo variants to one and one equality variant.

be-js/src/commonMain/kotlin/lang/temper/be/js/JsSupportNetwork.kt
```diff
@@ -397,12 +440,20 @@ internal object JsSupportNetwork : SupportNetwork {
 }
 
 private val supportedAutoConnecteds = setOf(
+    "core.type DenseBitVector.constructor()",
+    "core.type DenseBitVector.get()",
+    "core.type DenseBitVector.set()",
+    "core.type Deque.constructor()",
+    "core.type Deque.add()",
+    "core.type Deque.get isEmpty()",
+    "core.type Deque.removeFirst()",
```
mikesamuel, line 449: Sorted some things.

be-js/src/commonMain/resources/lang/temper/be/js/temper-core/core.js
```diff
@@ -204,24 +206,38 @@ export const cmpFloat = (a, b) => {
...
 
 /**
- * @template {string | number | boolean} T
- * @param {T} a
- * @param {T} b
- * @returns {number}
+ * Equality check consistent with cmpFloat64.
+ * @param {number} a
+ * @param {number} b
+ * @return {boolean}
```
mikesamuel, line 212: Got rid of generic comparison.  Added helpers for boolean and int64 comparison.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/regex/runtime.lua
mikesamuel, line 1: Changing generic comparison callouts to int comparison.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
mikesamuel, line 1: ⚠️ Mike is writing Lua ⚠️

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
```diff
@@ -1015,122 +1015,78 @@ function temper.float_eq(a, b)
...
-        return b == b
-    end
-    if a == 0 and b == 0 then
-        return zero_cmp(a, b) > 0
-    end
-    return a > b
-end
-
-function temper.float_ge(a, b)
+function temper.float_cmp(a, b)
```
mikesamuel, line 1018: Unchanged from red L1126 below, but diff isn't doing a great job here.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -140,12 +140,12 @@ def make_bubble_exception() -> Exception:
 
 
 def bubble() -> Any:
-    "Raises an exception for Temper bubbling."
+    """Raises an exception for Temper bubbling."""
```
mikesamuel, line 143: IDE wants triple-quoted strings for pydoc. 🤷

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -156,92 +156,18 @@ def float_cmp(left: float, right: float) -> int:
...
 
 
-def float_eq(left: float, right: float) -> bool:
-    "Checks if two floats are exactly equal, caring about nan and sign of zeros."
-    return (left == right and copysign(1.0, left) == copysign(1.0, right)) or (
-        isnan(left) and isnan(right)
-    )
+def bool_cmp(left: bool, right: bool) -> int:
+    """Three-way compares booleans."""
+    return left - right
```
mikesamuel, line 161: bools are ints.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -264,6 +190,12 @@ def arith_int_mod(dividend: int, divisor: int) -> int:
         q += 1
     return dividend - divisor * q
 
+
+def int64_cmp(left: bool, right: bool) -> int:
+    """Three-way compares booleans."""
+    return left - right
```
mikesamuel, line 196: No under/over flow on bigints.

be-rust/src/commonTest/kotlin/lang/temper/be/rust/RustBackendTest.kt
```diff
@@ -900,7 +900,7 @@ class RustBackendTest {
             |            struct ClosureGroup___0 {}
             |            impl ClosureGroup___0 {
             |                fn fn__0(& self) -> std::sync::Arc<String> {
-            |                    return std::sync::Arc::new("expected i < nums.length".to_string());
+            |                    return std::sync::Arc::new("expected i <=> nums.length < 0".to_string());
```
mikesamuel, line 903: Should change PseudoCode to do a nicer job with this when all the resugaring bells and whistles are on.

be/src/commonMain/kotlin/lang/temper/be/tmpl/TmpLOperator.kt
```diff
@@ -45,13 +45,13 @@ sealed class TmpLOperator(
     /** `<=` applied to Ints */
     object LeInt : Infix(WellKnownTypes.booleanType2, TmpLOperatorDefinition.Le)
 
-    /** `<=` applied to Ints */
+    /** `<` applied to Ints */
```
mikesamuel, line 48: copy/paste typos

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -78,31 +77,11 @@ private fun fTypeTypeToBoolean(sides: TypeShape, nullity: Nullity): Signature2 {
     )
 }
 
-private val fIntIntNullableToBoolean = fTypeTypeToBoolean(WKT.intTypeDefinition, Nullity.OrNull)
```
mikesamuel, line 81: Ordered comparison ops are no longer null accepting.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -235,6 +246,22 @@ private class StringCompareFun(
     }
 }
 
+private class BoolCompareFun(
```
mikesamuel, line 249: Was missing this one.  I figured why not have a way to order everything in the JSON value domain.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -1094,471 +1042,92 @@ object BuiltinFuns {
...
     }
+    val vLtIntFn = Value(ltIntFn)
 
-    val ltGenericFn: CallableValue = ComparisonFunction(
-        "<",
-        BuiltinOperatorId.LtGeneric,
-        WKT.booleanType2,
-    ) { cmp -> TBoolean.value(cmp < 0) }
-
-    /**
```
mikesamuel, line 1131: These doc comments mostly moved into BuiltinOperatorSpecs unchanged though I added some notes on how the desugaring works.

docs/for-users is the best place to review those doc changes.

builtin/src/commonMain/kotlin/lang/temper/builtin/BuiltinFuns.kt
```diff
@@ -1657,10 +1226,32 @@ object BuiltinFuns {
     val vDoPure = Value(doPure)
 }
 
+fun makeTypeFormal2(
+    fnName: String,
+    nameSuffix: String,
+    upperBounds: List<Type2>,
+): Pair<TypeFormal, Type2> =
```
mikesamuel, line 1233: Working around detekt complaining about uses of vararg spread operator.

Hoping to collapse these when StaticType goes away.

builtin/src/commonMain/kotlin/lang/temper/builtin/DesugarOperation.kt
```diff
@@ -127,10 +130,19 @@ object DesugarOperation : SpecialFunction, StaylessMacroValue, NamedBuiltinFun {
...
                     }
+                    when (nameText) {
+                        "==" -> return@classify "==" to OpClassification.Eq
+                        "!=" -> return@classify "==" to OpClassification.Ne
+                        "<" -> return@classify "<=>" to OpClassification.Lt
+                        "<=" -> return@classify "<=>" to OpClassification.Le
+                        ">=" -> return@classify "<=>" to OpClassification.Ge
+                        ">" -> return@classify "<=>" to OpClassification.Gt
+                        "<=>" -> return@classify "<=>" to OpClassification.Cmp
+                    }
```
mikesamuel, line 141: Here's where the desugaring happens.

frontend/src/commonMain/resources/core/core.temper
```diff
@@ -646,7 +646,7 @@ export class String extends MapKey<String> {
 
   @connected
   @operator("==")
-  public eq(other: builtins.String): builtins.Boolean { content == other }
+  public eq(other: builtins.String): builtins.Boolean;
```
mikesamuel, line 649: I don't want these to potentially lead to recursive invocation now that `==` is routed here.

frontend/src/commonMain/resources/std/regex/regex.temper.md
mikesamuel, line 1: Once we have auto-structural equality, there'll be a way to say `Dot -> ...` but rather than hack in equality for RegexNodes I thought it easier to use `is` clauses consistently in the matches here which had been relying on generic equality semantics.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/syntax-macro/desugar-equivalence-and-comparison-ops/expect/generateCode.temper
mikesamuel, line 1: This file and the next two are a new test of the comparison desugarings.

functional-test-suite/src/commonMain/resources/algos/myers-diff/myers-diff-lib.temper.md
```diff
@@ -40,11 +40,11 @@ A patch is a list of [changes][Change].
 The diff function compares two inputs, and computes a [Patch]: the changes
 from [left] required to derive [right].
 
-    export let diff<T extends Equatable<T> /*-| Change<T> | List<T> | Patch<T> */>(
+    export let diff<T /*-| Change<T> | List<T> | Patch<T> */>(
       left: List<T>,
       right: List<T>,
       /** Are two items the same? */
-      eq: fn (T, T): Boolean = fn (a: T, b: T): Boolean { a == b }
```
mikesamuel, line 47: Relied on generic comparison.  meta interfaces will enable this kind of default.

fundamentals/src/commonMain/kotlin/lang/temper/value/UnpackValue.kt
```diff
@@ -48,8 +48,13 @@ fun unpackValue(
          * but you can use an underscore to separate digit groups.
          *
          * ```temper
-         * [123,456,789] == [123 , 456 , 789] &&  // Commas separate elements
-         * [123_456_789] == [123456789]
```
mikesamuel, line 52: We haven't actually defined equality on Lists.

I think I need to thread through extension function resolutions to handle that.

interp/src/commonMain/kotlin/lang/temper/interp/BuiltinEnvironment.kt
```diff
@@ -53,14 +53,6 @@ private object Builtins {
     val nameKeyToValue: Map<String, Value<*>>
     init {
         val m = mutableMapOf(
-            "<" to Value(BuiltinFuns.lessThanFn),
-            ">" to Value(BuiltinFuns.greaterThanFn),
-            "<=" to Value(BuiltinFuns.lessEqualsFn),
-            ">=" to Value(BuiltinFuns.greaterEqualsFn),
-            "==" to Value(BuiltinFuns.equalsFn),
-            "!=" to Value(BuiltinFuns.notEqualsFn),
-            "<=>" to Value(BuiltinFuns.cmpFn),
```
mikesamuel, line 62: Getting rid of the last uses of cover functions.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
tjpalmer: I wouldn't change rendering for `!=` prettiness but maybe could customize the support code as you say. And I think it's ok to defer that past this pr.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpSupportNetwork.kt
```diff
@@ -235,12 +237,61 @@ object CSharpSupportNetwork : SupportNetwork {
...
+                        ComparisonKind.LessThanOrEqual -> CSharpOperator.LessEquals
+                        ComparisonKind.GreaterThanOrEqual -> CSharpOperator.GreaterEquals
+                        ComparisonKind.GreaterThan -> CSharpOperator.GreaterThan
+                    },
+                ),
+            ),
+            typeActuals = tmpl.typeActuals.deepCopy(),
+            parameters = freeParameters,
+        )
+    }
```
tjpalmer, line 294: I wonder there's a way to share more `simplifyPossibleComparison` logic across backends, but ok not to worry it for this pr.

frontend/src/commonMain/resources/std/regex/regex.temper.md
```diff
@@ -143,47 +143,46 @@ export class CodePoints(
...
 export interface Special extends RegexNode {}
-export let Begin: Special = doPure { (): Special => class Begin extends Special {}; new Begin() };
-export let Dot: Special = doPure { (): Special => class Dot extends Special {}; new Dot() };
-export let End: Special = doPure { (): Special => class End extends Special {}; new End() };
+class BeginSpecial extends Special {}
+export let Begin: Special = doPure { (): Special => new BeginSpecial() };
+class DotSpecial extends Special {}
+export let Dot: Special = doPure { (): Special => new DotSpecial() };
+class EndSpecial extends Special {}
+export let End: Special = doPure { (): Special => new EndSpecial() };
```
tjpalmer, line 171: Maybe we do just need some kind of explicit singleton support at some point. But I guess I can survive this for now.

And I guess we didn't break lua regex with these changes? We might need to update the regex parser and engine projects.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/define/classes-with-disclosures/expect/define.temper
```diff
@@ -1,77 +1,77 @@
-@fn let f__6;
+@fn let f__11;
```
tjpalmer, line 1: I suspect the new auto-update was nice for all these number changes.

be-csharp/src/commonMain/resources/lang/temper/be/csharp/temper-core/Core.cs
```diff
@@ -75,21 +75,6 @@ internal static int Clamp(this int i, int min, int max)
             return Math.Max(min, Math.Min(i, max));
         }
 
-        // TODO Make this generic instead of object?
-        public static int Compare(object a, object b)
-        {
```
tjpalmer, line 80: Glad to see these generic compares gone.

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/init.lua
tjpalmer: > ⚠️ Mike is writing Lua ⚠️

💪

be-lua/src/commonMain/resources/lang/temper/be/lua/temper-core/regex/runtime.lua
tjpalmer: I made the comment on regex.temper.md before seeing this change. So looks like you updated this file at least for something. Did you edit in place or change the source repo and rebuild?

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
tjpalmer: > IDE wants triple-quoted strings for pydoc. 🤷

I think I've run black on this before. So maybe they care about different things. Or maybe my memory is off.

be-py/src/commonMain/resources/lang/temper/be/py/temper-core/temper_core/__init__.py
```diff
@@ -264,6 +190,12 @@ def arith_int_mod(dividend: int, divisor: int) -> int:
...
+
+
 def arith_bit_shl32(n: int, shift: int) -> int:
     """
     Performs an Int32 left shift.
```
tjpalmer, line 201: Reducing lines of code on some of these files is nice.

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
tjpalmer: Do we have any backend tests of implementing `==` or `<=>` for user classes and seeing what we generate? If we don't yet, then in the future, we should plan to generate operator overloads in appropriate languages.

tjpalmer (approved):

be-cpp/src/commonTest/kotlin/lang/temper/be/cpp/CppBackendTest.kt
mikesamuel: > Do we have any backend tests of implementing == or <=> for user classes and seeing what we generate? If we don't yet, then in the future, we should plan to generate operator overloads in appropriate languages.

No, and we should.  And I think I probably need to add backend tests to make sure Java/C#/Py bridge `.equals(Object)`/`.Equals(object)`/`.__eq__(x)` to any `@operator("==")` method, and probably `operator==` for C++ and `eq` for Rust iirc.

frontend/src/commonTest/resources/lang/temper/frontend/stage-tests/define/classes-with-disclosures/expect/define.temper
mikesamuel: Yeah.  Thinking of maybe tweaking assertModuleAtStage so it comes out with the more canonical `__0`s so there's less diff spam going forward.

be-csharp/src/commonMain/kotlin/lang/temper/be/csharp/CSharpSupportNetwork.kt
mikesamuel: Yeah. I was thinking that, but the set of actual builtin operators / connecteds is specific to the actual backed.  Like CmpStrStr is very much ok on some, not on others.

Now that I think about it, it might be nice if Java used `java.lang.Double.compare` which iirc matches our order.

mikesamuel:
Responded to some of your questions, and not seeing the responses here.  Will try to figure out why not.

